PR #128, opened 2024-02-26
refactor(desktop): use 2 separate country lists for entry exit gateways

doums:
- use 2 different country lists for entry and exit gateways
- exit gateways filtered out gateway with **no** IPR

### DIFF
--- a/nym-vpn-desktop/src-tauri/src/commands/node_location.rs
+++ b/nym-vpn-desktop/src-tauri/src/commands/node_location.rs
@@ -1,18 +1,13 @@
-use std::env;
-
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use tauri::State;
-use tracing::{debug, error, instrument, trace};
+use tracing::{debug, instrument, trace};
 use ts_rs::TS;
 
 use crate::{
     country::{Country, FASTEST_NODE_LOCATION},
     error::{CmdError, CmdErrorSource},
-    http::{
-        client::HTTP_CLIENT,
-        explorer_api::{JsonGateway, EXPLORER_API_URL, GATEWAYS_ENDPOINT},
-    },
+    http::{explorer_api, nym_api},
     states::{app::NodeLocation, SharedAppData, SharedAppState},
 };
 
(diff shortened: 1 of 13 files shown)


nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -85,38 +84,152 @@ pub async fn get_node_location(
...
+        // remove any duplicate identity key
+        .unique()
+        // find the corresponding country code in the explorer response
+        // for each identity key
+        // mapping to a list of locations
+        .filter_map(|identity_key| {
+            let location = explorer_json
+                .iter()
+                .find(|gateway| gateway.gateway.identity_key == identity_key)
+                .map(|gateway| gateway.location.clone());
```
neacsu, line 224: You could do `find_map`

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -85,38 +84,152 @@ pub async fn get_node_location(
...
+            match location {
+                Some(Some(location)) => Some(location),
+                _ => None,
```
neacsu, line 227: You could just add `flatten` to the iterator above

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -85,38 +84,152 @@ pub async fn get_node_location(
...
+    debug!("parsing responses");
+    let list = nym_api_json
+        .into_iter()
+        // only retain gateways with IP packet router
+        // mapping to a list of identity keys
+        .filter_map(|gateway| {
+            gateway
+                .self_described
+                .and_then(|desc| desc.ip_packet_router)
+                .map(|_| gateway.bond.gateway().identity_key.clone())
```
neacsu, line 213: I think this works, but it's a bit hard to read. Maybe just do another `filter_map` in the main iterator, and append the identity key to a tuple, so that it gets carried forward where you need it.

nym-vpn-desktop/src/pages/location/NodeLocation.tsx
```diff
@@ -47,11 +48,16 @@ function NodeLocation({ node }: { node: NodeHop }) {
 
   // request backend to refresh cache
   useEffect(() => {
-    invoke<Country[]>('get_node_countries')
+    invoke<Country[]>('get_countries', {
+      nodeType: node === 'entry' ? 'Entry' : 'Exit',
```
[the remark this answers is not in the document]
doums: So it's already typed (inferred here by typescript, concrete type is `NodeHop`)

For instance that would raise TS errors:
```
node === 'BadValue' ? 'Entry' : 'Exit',
```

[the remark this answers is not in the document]
doums: > it would be better to have the Exit type explicit as well? 

I'd say no, there is no real added value to type tauri "commands" arguments, since it's cross language, you would have to maintain corresponding TS types to tauri args anyway and so I prefer not to do that

Lets keepit simple, no need to add a custom layer on top of tauri `invoke` API

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
doums: sorry for this one I dont see what you mean, the code is valid (yes it works), so unless you share some code I'm not able to change it

nym-vpn-desktop/src/pages/location/NodeLocation.tsx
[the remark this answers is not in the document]
neacsu: I was referring that you would fail before calling `invoke`, by checking that node is either `entry` or `exit`, so that it doesn't go on the `Exit` path in case it's different then `entry`. But since I'm not very familiar with TS, if it already does that by some typing magic, then it's all good.

neacsu (approved):

PR #130, opened 2024-02-27
Feature/ipv6

neacsu:


nym-vpn-cli/src/commands.rs
```diff
@@ -79,9 +85,13 @@ pub(crate) struct CliArgs {
...
-    /// The IP address of the nym TUN device that wraps IP packets in sphinx packets.
-    #[arg(long, alias = "ip", value_parser = validate_ip)]
-    pub(crate) nym_ip: Option<Ipv4Addr>,
+    /// The IPv4 address of the nym TUN device that wraps IP packets in sphinx packets.
+    #[arg(long, alias = "ipv4", value_parser = validate_ipv4, requires = "nym_ipv6")]
+    pub(crate) nym_ipv4: Option<Ipv4Addr>,
+
+    /// The IPv6 address of the nym TUN device that wraps IP packets in sphinx packets.
+    #[arg(long, alias = "ipv6", value_parser = validate_ipv6, requires = "nym_ipv4")]
+    pub(crate) nym_ipv6: Option<Ipv6Addr>,
```
octol, line 94: I wonder, maybe we should allow not having ipv6? Although I don't know how common that is in practice to only have a ipv4 address.. :shrug:

nym-vpn-cli/src/main.rs
```diff
@@ -73,14 +73,19 @@ async fn run() -> Result<()> {
 
     let entry_point = parse_entry_point(&args)?;
     let exit_point = parse_exit_point(&args)?;
+    let nym_ips = if let (Some(ipv4), Some(ipv6)) = (args.nym_ipv4, args.nym_ipv6) {
```
octol, line 76: optional: create a constructor for IPPair

nym-vpn-cli/src/commands.rs
neacsu: After pondering on it over the weekend, I think it's safe to assume that desktop or mobile machines with network stack that don't have support for IPv6 are pretty rare. Not sure how to test this, but I think the only place there might be an error would be either when we setup the IPv6 routes or the IPv6 address of the interface. In rest, we can safely assign v6 addresses that will never be used, and it would be more robust for our part to do so.
In case such error happens, I think it would be better to deal with it at that point, probably making that specific error a warning instead, since proactively trying to figure out the v6 support of a machine doesn't seem trivial.

nym-vpn-cli/src/main.rs
neacsu: Done: https://github.com/nymtech/nym/pull/4440

nym-vpn-lib/src/routing.rs
```diff
@@ -233,21 +236,15 @@ pub async fn setup_routing(
         device_mtu = dev.as_ref().mtu(),
     );
 
+    let _ipv6_addr = config.tun_ips.ipv6.to_string();
```
octol, line 239: nitpick: remove leading _

octol (approved):

nym-vpn-lib/src/routing.rs
neacsu: This is to not have warnings on building for platforms different than linux and macos

PR #131, opened 2024-02-28
refactor(desktop): switch to embedded k/v store

doums:
App data store (used to save user settings values and any other persistent values the app need to) has been designed as a raw `toml` file written on disk. Stored at `.local/share/nym-vpn/app-data.toml` (Linux)
In a typed serializable way. This was functional and worked OK, but at the cost of lot of required code duplicates for each crud operations (and bad perfs as it's a matter of read/write on file).
Plus to expose one crud operation to JS side, it required to write a dedicated tauri command.
Another big issue was that any stale toml data written on the file (eg. after related rust structures update) would hard break any read (and so deserializing) attempt of the data, causing hard stop of the app. The only fix to that were to rm the file, so that a fresh one can be written.

Let's switch to something more optimized, reliable, and migrate/recoverable friendly.

Say hello to the embedded k/v database -> https://sled.rs/

- drop app data store serialized `toml` file
- switch to an embedded k/v database
- refactor to clean all those dirty code dups
- expose a simple tauri command API to give JS side direct crud access to the db
- strong typed keys
- db values are serialized to Json bytes, so directly comestible by both Rust and JS without the need to create an alternative layer
- by default discard any deserialization error**

** When a key's value become staled and does not match Rust target structure anymore, eg. after some Rust type updates, the rust call will fail with a deserialization error. For now, any deserialization error to concrete Rust structure/serde JsonValue, are discarded by removing the corresponding key. Making the store system more resilient.

Note: For now, removing those key from store are not an issue since it's not critical data in any form. The db is currently used to only store custom user settings.
In the future, once we start to store more sensitive data, we would evolve the store system to be granular on discarding desieralize errors.


nym-vpn-desktop/src-tauri/src/main.rs
```diff
@@ -116,7 +105,10 @@ async fn main() -> Result<()> {
     // Read the env variables in the provided file and export them all to the local environment.
     nym_config::defaults::setup_env(app_config.env_config_file.clone());
 
-    let app_state = AppState::try_from((&app_data, &app_config)).map_err(|e| {
+    info!("Creating k/v embedded db");
+    let db = Db::new()?;
+
+    let app_state = AppState::try_from((&db, &app_config)).map_err(|e| {
```
octol, line 111: Imho from the users perspective it would be pretty neat to still continue to the UI if there as a failure here, and display it. So that people who don't run from the terminal also get to see what went wrong

nym-vpn-desktop/src-tauri/src/commands/db.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub ui_theme: Option<UiTheme>,
+    pub ui_root_font_size: Option<u32>,
+    pub vpn_mode: Option<VpnMode>,
+    pub entry_node_location: Option<NodeLocation>,
+    pub exit_node_location: Option<NodeLocation>,
+}
+
+#[instrument(skip_all)]
+#[tauri::command]
+pub async fn db_get_batch(db: State<'_, Db>) -> Result<AppData, CmdError> {
```
jstuczyn, line 36: I feel like this function would have been way simpler if instead we just stored `AppData` as a single serialized struct in say a `.toml` file. or even a stupid `.json`. I said it before, I feel like `sled` is an overkill. we don't need its performance here.

plus I reckon it will be way more user-readable

nym-vpn-desktop/src-tauri/src/commands/db.rs
```diff
@@ -0,0 +1,138 @@
...
+        _ => CmdError::new(
+            CmdErrorSource::InternalError,
+            format!("Failed to insert key: {e}"),
+        ),
+    })
+}
+
+#[instrument(skip(db))]
+#[tauri::command]
+pub async fn db_flush(db: State<'_, Db>) -> Result<usize, CmdError> {
```
jstuczyn, line 130: I don't think flushing operation is something we should ever expose to the frontend. if anything we should just flush on every write and call it a day

nym-vpn-desktop/src-tauri/src/commands/db.rs
```diff
@@ -0,0 +1,138 @@
...
+    db.insert(key, &value).map_err(|e| match e {
+        DbError::Serialize(e) => CmdError::new(
+            CmdErrorSource::CallerError,
+            format!("Failed to insert key, bad JSON input: {e}"),
+        ),
+        _ => CmdError::new(
+            CmdErrorSource::InternalError,
+            format!("Failed to insert key: {e}"),
+        ),
+    })
```
jstuczyn, line 125: does it make sense to expose raw database operations to the frontend? shouldn't we just have helpers like `set_autoconnect` instead?

nym-vpn-desktop/src-tauri/Cargo.toml
```diff
@@ -32,6 +32,9 @@ reqwest = { version = "0.11", features = ["json"] }
 itertools = "0.12"
 clap = { version = "4.5", features = ["derive"] }
 build-info = "0.0.34"
+sled = "0.34.7"
+strum = { version = "0.26", features = ["derive"] }
+tap = "1.0.1"
```
jstuczyn, line 37: this can be now replaced with `inspect_err` from standard library : )

jstuczyn (commented):
I haven't looked into details of the implementation since I don't have sufficient knowledge of the underlying vpn implementation. I also haven't looked at all at your TS code but I trust all is good there : )

nym-vpn-desktop/src-tauri/src/commands/db.rs
doums: > I don't think flushing operation is something we should ever expose to the frontend

Why that? Flush is exposed because user can quit from within the app via a button "Quit", and we want to flush the db before exit basically

nym-vpn-desktop/src-tauri/src/commands/db.rs
doums: > does it make sense to expose raw database operations to the frontend

In short: Absolutely **YES** and this is the main reason of that new implem

Long: this is already explained in PR description, but I'll try to explain it using a different wording
With storing data within a simple file (as it is currently)
you have to provide a dedicated tauri API for each value frontend want to get/set
That means you want to set/get the uiTheme from the frontend and saveit, you have to code 2tauri dedicated "commands"
one to set, one to get
you have to repeat that for any other saved value, so for selected entry location, exit location and so on.
It's a bad design, it creates code duplication, exact same logic is repeated you get it.

nym-vpn-desktop/src-tauri/Cargo.toml
doums: nice one, didnt know about it, thx ;)

nym-vpn-desktop/src-tauri/src/commands/db.rs
doums: The whole purpose of this refactoring is to allow crud ops on key/value pairs. To have the flexibility needed by UI interactions.
This is why we don't want to store the entire app data as a single block of data. As we want to be able to make single value update/read and so on.

> we don't need its (sled) performance here

So I agree, and I didn't say it was the main reason to switch on sled. But free performance isn't something always positive and good to take?

> plus I reckon it will be way more user-readable

Yes but we don't need that, only code access is needed. Not "human" access. It's app internal data; not like a config file, where dev/user would have direct interaction with.

nym-vpn-desktop/src-tauri/src/main.rs
doums: @octol if app state can't be recreated from the config file and/or db store, that means it will be 100% nonfunctional and the issue causing this is really an important and severe one (disk access refused, hard IO error etc).

Anyway, discussing that specific point is out of the scope of this PR:
please open a ticket if you really think it's necessary.

nym-vpn-desktop/src-tauri/src/commands/db.rs
doums: > we should just flush on every write and call it a day

pushed b8f8f04 to flush after db mutations

octol (approved):
PR has been discussed extensively elsewhere, marking as approved. But let's wait for @jstuczyn in case he has additional feedback

jstuczyn (approved):

PR #133, opened 2024-02-28
feat(desktop): add licenses page

doums:
add licenses page (under settings -> Legal)
the license content is served using an `iframe` rendering licenses.html file generated by `cargo-about` command


octol (changes requested):
Make sure to generate the licenses as part of the build

octol:
Oh and another thing, we need to remember to add the frontend libraries as well. (Can be done in a later PR)

doums:
> Make sure to generate the licenses as part of the build

@octol I think we **only** want to generate the licenses file while making **release** build. So as an additional step in realease workflow.

We don't want it to run each time for non release builds/dev run:
- as license generation need an extra tool to be installed on the system (`cargo-about`)
-  the generation takes a considerable amount of time (~30s on my machine)
- in future PR we will have to add JS licenses generation which will add more duration

octol:
> > Make sure to generate the licenses as part of the build
> 
> @octol I think we **only** want to generate the licenses file on **release** build. So as an additional step in realease workflow.
> 
> We don't want it to run each time for regular builds/dev runs:
> 
>     * as license generation requires a 3rd party tool to be installed on the system (`cargo-about`)
> 
>     * the generation takes a considerable amount of time (~20s on my machine)
> 
>     * in future PR we will have to add JS licenses generation which will add more duration

Yep that's fine to only generate it as part of the release builds

Makefile
```diff
@@ -50,7 +50,7 @@ generate-licenses-cli-json:
 	cargo about generate --all-features --format json -o all_licenses_cli.json
 
 generate-licenses-desktop:
-	cargo about generate --all-features -m $(DESKTOP_DIR)/Cargo.toml $(DESKTOP_DIR)/about.hbs -o all_licenses_desktop.html
+	cargo about generate --all-features -m $(DESKTOP_DIR)/Cargo.toml $(DESKTOP_DIR)/about.hbs -o nym-vpn-desktop/public/licenses-rust.html
```
octol, line 53: Imho let's make the same path change to the json build rule below.

Also, since I see that you added public/licenses-rust.html, we might as well remove the license file that is currently in the root dir

.github/workflows/release-nym-vpn-desktop.yml
```diff
@@ -148,6 +143,14 @@ jobs:
       - name: Prepare desktop build
         run: mkdir nym-vpn-desktop/dist
 
+      - name: Generate licenses file (Rust)
+        env:
+          DESKTOP_DIR: nym-vpn-desktop/src-tauri
+          PUBLIC_DIR: nym-vpn-desktop/public
+        run: |
+          rm $PUBLIC_DIR/licenses-rust.html || true
+          cargo about generate --all-features -m $DESKTOP_DIR/Cargo.toml $DESKTOP_DIR/about.hbs -o $PUBLIC_DIR/licenses-rust.html
```
octol, line 152: Imho let's use the make target

octol (approved):

PR #146, opened 2024-03-08
Handle new IpPacketResponseData::Info

octol:
Client changes for https://github.com/nymtech/nym/pull/4447


neacsu (approved):
You probably want to update the nym commit too :D

PR #147, opened 2024-03-08
Windows build support of library

neacsu:


octol:
I think we can skip cargo build and cargo test on windows for now, only do clippy since that should pass with the current changes

PR #152, opened 2024-03-08
Report the gateway connected to

octol:
Report back the connected gateway to the frontend.

NOTE: the final bit is commented out until the frontend can handle the receiving the type.


nym-vpn-desktop/src-tauri/src/commands/connection.rs
```diff
@@ -248,7 +248,7 @@ pub async fn disconnect(
     })?;
     debug!("Stop message sent");
 
-    Ok(app_state.state)
+    Ok(app_state.state.clone())
```
doums, line 251: why these `clone()` here?

octol: It's because the connection state now contains a String with the connected gateway, so it does not implement Copy anymore

nym-vpn-desktop/src-tauri/src/events.rs
```diff
@@ -0,0 +1,68 @@
...
+            error,
+            start_time,
+        }
+    }
+}
+
+pub trait AppHandleEventEmitter {
+    fn emit_disconnected(&self, error: Option<String>);
+    fn emit_connected(&self, now: OffsetDateTime, gateway: String);
+}
```
doums, line 41: better if you also implement `emit_{disconnecting/connecting}`

octol: yep good point!

doums (approved):

PR #155, opened 2024-03-09
ci(desktop): enable cache

doums:
enable rust and node cache to speed up jobs


.github/workflows/release-nym-vpn-desktop.yml
```diff
@@ -137,9 +137,11 @@ jobs:
           path: ${{ env.WG_GO_LIB_PATH }}
 
       - name: Install Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
```
octol, line 140: What's the argument for keeping this? Vs the rust code?

doums: since there is no reason to not keeping it, node-setup cache only cache node installation and node project dependencies installation
caching does not play with JS build steps

octol (approved):

PR #157, opened 2024-03-10
ci(desktop): remove no_strip env var

doums:


octol:
Let's keep this here for a bit while we wait for people to test the current binaries, to see if we need to disable this or not.

PR #174, opened 2024-03-14
Feature/ios support

neacsu:


.github/workflows/ci-nym-vpn-cli.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           workspaces: |
             nym-vpn-cli
 
+      - name: Add iOS target
```
octol, line 57: Suggestion: put this step directly after the rust-toolchain step

nym-vpn-lib/src/platform/macos.rs
```diff
@@ -38,17 +27,32 @@ fn init_logs() {
...
 
     if get_vpn_state().await != ClientState::Uninitialised {
         warn!("VPN was already inited. Try starting it");
         return;
     }
 
-    let mut vpn = NymVpn::new(entry_gateway, exit_router);
-    vpn.gateway_config.api_url = api_url;
+    let mut vpn = NymVpn::new(config.entry_gateway, config.exit_router);
+    vpn.gateway_config.api_url = config.api_url;
```
octol, line 55: I think you'll probably want to set explorer_url too?

nym-vpn-lib/src/platform/ios.rs
```diff
@@ -0,0 +1,81 @@
...
+        warn!("VPN was already inited. Try starting it");
+        return;
+    }
+
+    let mut vpn = NymVpn::new(
+        config.entry_gateway,
+        config.exit_router,
+        config.tun_provider,
+    );
+    vpn.gateway_config.api_url = config.api_url;
```
octol, line 79: I think you'll probably want to set explorer_url too?

octol (approved):

nym-vpn-lib/src/platform/ios.rs
neacsu: For now yep, we do want that. But outside this PR, I think it would be better to actually replace the two URLs with the .env provided ones, since we rely on that configuration anyway.

PR #177, opened 2024-03-16
Add android country iso api

zaneschepke:


nym-vpn-lib/src/platform/android.rs
```diff
@@ -173,11 +172,57 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_defaultTunConfig<'env>
...
 
+impl<'env> IntoJava<'env> for String {
+    type JavaType = JString<'env>;
+
+    fn into_java(self, env: &JNIEnv<'env>) -> Self::JavaType {
+        env.new_string(&self).expect("Failed to create Java String")
+    }
+}
+
+pub trait IntoJava<'env> {
```
octol, line 188: Let's make a comment here about that ideally we'd use the IntoJava trait in jnix and it's impl for String, but that we didn't get it to work so this is a quick fix for now. In the future maybe we can remove this trait and use the jnix directly

nym-vpn-lib/src/platform/android.rs
```diff
@@ -173,11 +180,94 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_defaultTunConfig<'env>
...
     _this: JObject<'_>,
 ) -> JObject<'env> {
+    use jnix::IntoJava;
     let env = JnixEnv::from(env);
-
     TunConfig::default().into_java(&env).forget()
 }
 
+// ideally we use IntoJava trait from jnix, quick fix for now
+impl<'env> IntoJava<'env> for String {
```
octol, line 189: suggestion: usually the interface (trait) comes before the concrete implementation

nym-vpn-lib/src/platform/android.rs
```diff
@@ -105,23 +104,31 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_initVPN(
...
     };
-    let api_url = Url::from_str(&String::from_java(&env, api_url)).expect("Invalid api url");
-    let explorer_url =
-        Url::from_str(&String::from_java(&env, explorer_url)).expect("Invalid explorer url");
     let entry_gateway: EntryPoint = serde_json::from_str(&String::from_java(&env, entry_gateway))
-        .expect("Invalid entry gateway");
-    let exit_router: ExitPoint =
-        serde_json::from_str(&String::from_java(&env, exit_router)).expect("Invalid exit router");
+        .map_err(|e| error!("Could not parse entry point {:?}", e))
+        .unwrap();
```
octol, line 119: Do we want to panic on incorrect input? Earlier in the function I see a naked `return;`, consider doing the same unless it's possible to return an error to the caller

nym-vpn-lib/src/platform/android.rs
```diff
@@ -105,23 +104,31 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_initVPN(
...
+    let exit_router: ExitPoint = serde_json::from_str(&String::from_java(&env, exit_router))
+        .map_err(|e| error!("Could not parse exit point {:?}", e))
+        .unwrap();
```
octol, line 122: Same as above

nym-vpn-lib/src/platform/android.rs
```diff
@@ -215,6 +305,9 @@ enum Error {
     #[error(display = "Failed to select() on tunnel device")]
     Select(#[error(source)] nix::Error),
 
+    #[error(display = "Gateway does not contain a two character country ISO")]
```
octol, line 308: Just a side note for something to consider in the future: convention in is to start Display impl strings for error types with lowercase, since usually they are embedded as part of another string at the top-level when they are display. E.g
```
if let Err(err) = foo() {
    error!("Failed to call foo: {err}");
}
```

nym-vpn-lib/src/platform/android.rs
```diff
@@ -300,6 +393,29 @@ fn try_sending_random_udp(is_ipv6_enabled: bool) -> Result<(), SendRandomDataErr
     Ok(())
 }
 
+fn parse_api_url_from_java(env: &JnixEnv, api_url: JString<'_>) -> Url {
+    match Url::from_str(&String::from_java(env, api_url)) {
```
octol, line 397: This looks fine to me, here is another style for the same thing that you might come across. IMHO both styles are fine

```rust
    Url::from_str(&String::from_java(env, api_url)).unwrap_or_else(|err| {
        error!("Could not parse api_url: {:?}. Using the default.", err);
        gateway_client::Config::default().api_url
    })
```

octol: There is however a functional impact of this, the default value is for mainnet, and if all other settings if for a different network e.g sandbox then you'll end up with an inconsistent mix, and also probably silently "doing the wrong thing". I suggest returning an error here all the way back up to the top and not attempting to proceed

nym-vpn-lib/src/platform/android.rs
```diff
@@ -300,6 +393,29 @@ fn try_sending_random_udp(is_ipv6_enabled: bool) -> Result<(), SendRandomDataErr
...
+}
+
+fn parse_explorer_url_from_java(env: &JnixEnv, explorer_url: JString<'_>) -> Option<Url> {
+    match Url::from_str(&String::from_java(env, explorer_url)) {
+        Err(err) => {
+            error!(
+                "Could not parse explorer_url: {:?}. Using the default.",
+                err
+            );
+            gateway_client::Config::default().explorer_url
```
octol, line 413: See comment above, imho we should probably return an error here instead of using the (likely wrong) default value

.github/workflows/ci-nym-vpn-cli.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04-arc, custom-runner-mac-m1, windows-2022 ]
+        os: [ ubuntu-22.04, custom-runner-mac-m1, windows-2022 ]
```
octol, line 22: Doesn't it work with the arc runner?

zaneschepke: Oh, I'm not sure when or how that got removed 😄

zaneschepke: I guess something is wrong with arc so I am switching it back.

octol: It failed because of the change further down, the arc runner should be fine

nym-vpn-lib/src/platform/android.rs
```diff
@@ -133,12 +149,9 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_runVPN(_env: JNIEnv, _
     let state = RUNTIME.block_on(get_vpn_state());
     if state != ClientState::Disconnected {
         warn!("Invalid vpn state: {:?}", state);
-        return;
```
neacsu, line 136: Why remove this `return`, when there's no reason to continue in a corrupted state?

nym-vpn-lib/src/platform/android.rs
```diff
@@ -133,12 +149,9 @@ pub extern "system" fn Java_net_nymtech_vpn_NymVpnService_runVPN(_env: JNIEnv, _
...
     }
 
-    let vpn = RUNTIME
-        .block_on(take_vpn())
-        .expect("VPN configuration was cleared before it could be used");
```
neacsu, line 141: I think the previous log was a bit better as it was more explicit in what is happening, as in this case, the VPN was initialized, but then got disconnected between the check from above and the call from here.

nym-vpn-lib/src/platform/android.rs
zaneschepke: Clippy complained.

nym-vpn-lib/src/platform/android.rs
zaneschepke: Okay, we can revert.

nym-vpn-lib/src/platform/android.rs
neacsu: Hmm, can you reproduce that clippy run? It doesn't complain if I run with it.

zaneschepke: I don't see that run anymore. Lets go with it if clippy is happy 😆

neacsu (approved):

PR #179, opened 2024-03-18
Complete interface for nym device configuration

neacsu:


nym-vpn-lib/src/routing.rs
```diff
@@ -68,14 +71,16 @@ impl RoutingConfig {
     ) -> Self {
         debug!("TUN device IPs: {}", tun_ips);
         let mut mixnet_tun_config = tun2::Configuration::default();
+        let mtu = vpn.nym_mtu.unwrap_or(DEFAULT_TUN_MTU);
         // only IPv4 is supported by tun2 for now
         mixnet_tun_config.address(tun_ips.ipv4);
-        mixnet_tun_config.mtu(vpn.nym_mtu.unwrap_or(DEFAULT_TUN_MTU));
+        mixnet_tun_config.mtu(mtu);
```
octol, line 77: I wonder if we can have the same MTU for wireguard and mixnet, since they are nested? Probably the inner one needs to be slightly smaller?

octol (approved):

nym-vpn-lib/src/routing.rs
neacsu: After offline discussions, we concluded that the outer-most layer is the one that matters for MTU, so the two devices shouldn't be related

PR #191, opened 2024-03-19
build(desktop): add appimage installer

doums:
- update release workflow to upload `AppImage` artifact (including sha256sum) in release's assets root
- provide a bash script to install the AppImage easily

Once merged the script could be served from a nice URL and this would be the only installation instruction
```
curl -fsSL https://install.nymvpn.com/appimage | bash
```


nym-vpn-desktop/installer/appimage.sh
```diff
@@ -0,0 +1,131 @@
...
+  fi
+}
+
+need_cmd mktemp
+temp_dir=$(mktemp -d)
+install_dir="$HOME/.local/bin"
+icons_dir="$HOME/.local/share/icons"
+appimage="nym-vpn_${version}.AppImage"
+target_appimage="nym-vpn.appimage"
+desktop_dir="$HOME/.local/share/applications"
```
octol, line 65: What about looking these directories up by referring to XDG? Not sure we can assume that they will exist

doums: `~/.local/share` is very standard and `XDG_DATA_HOME` default. Even if the sub dirs don't exist, `install -D...` will create them as needed.

doums: but yes we can check if `$XDG_DATA_HOME` is set first and default to `$HOME/.local/share` if not (according to XDG base dir spec)

doums: Regarding `$HOME/.local/bin` there is no XDG variable for it, but it is specified in the spec

> User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.

see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

octol (approved):

PR #192, opened 2024-03-20
Mixnet connection beacon

octol:
Create a mixnet connection beacon and keep track of if the entry gateway is not routing our mixnet traffic

### DIFF
--- a/nym-vpn-lib/src/connection_monitor.rs
+++ b/nym-vpn-lib/src/connection_monitor.rs
@@ -64,8 +64,8 @@ impl ConnectionMonitor {
                 _ = report_interval.tick() => {
                     debug!("Connection statistics: {:?}", self.stats);
 
-                    // Send hello messages, so listerners can hear that we are still alive
-                    task_client.send_status_msg(Box::new(ConnectionMonitorStatus::Hello));
+                    // Send I'm alive messages, so listerners can hear that we are still there
+                    task_client.send_status_msg(Box::new(ConnectionMonitorStatus::ImAlive));
 
                     if let Some(latest_self_ping) = self.stats.latest_self_ping {
                         if latest_self_ping.elapsed() > SELF_PING_EXPIRY {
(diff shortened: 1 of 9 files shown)


nym-vpn-lib/src/connection_monitor.rs
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                }
+                _ = report_interval.tick() => {
+                    debug!("Connection statistics: {:?}", self.stats);
+
+                    // Send hello messages, so listerners can hear that we are still alive
+                    task_client.send_status_msg(Box::new(ConnectionMonitorStatus::Hello));
+
+                    if let Some(latest_self_ping) = self.stats.latest_self_ping {
+                        if latest_self_ping.elapsed() > SELF_PING_EXPIRY {
```
neacsu, line 71: Suggestion: Maybe use the opposite of this affirmation and the `continue` action, and the print and error message outside of the outer `if` statement, so that you don't duplicate it

nym-vpn-lib/src/connection_monitor.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+// Just like in nym_task::TaskManager and TaskStatus, strictly speaking this is not an error, but a
+// status message. We're just piggybacking on the error trait for now. In the future, we might want
+// to create a separate trait in nym_task::TaskManager
+#[derive(thiserror::Error, Debug)]
+pub enum ConnectionMonitorStatus {
+    #[error("hello")]
+    Hello,
```
neacsu, line 93: Maybe a proper name would be `Alive` or something that shows it's still working?

neacsu (approved):

nym-vpn-lib/src/connection_monitor.rs
octol: Yep it's a imho a good idea to dedup, although in this case I expect these two branches to diverge very soon with additional stats put in place

nym-vpn-lib/src/connection_monitor.rs
octol: Good point, I'll change

PR #193, opened 2024-03-20
Increase subnet range for clients

neacsu:


octol (commented):
There is some validation done in command.rs as well

octol:
> There is some validation done in command.rs as well

My bad, the current changes should cover that

octol (approved):

PR #203, opened 2024-03-22
feat: vpn proto service definition

doums:
Initial proposition for protobuf service definition of the future IPC layer (between daemon and client apps)

based on the following resources:
- https://grpc.io/docs/what-is-grpc/core-concepts/
- https://protobuf.dev/programming-guides/proto3/
- https://protobuf.dev/programming-guides/style/
- https://protobuf.dev/programming-guides/dos-donts/
- https://protobuf.dev/programming-guides/api/

### DIFF
--- a/proto/nymvpn.proto
+++ b/proto/nymvpn.proto
@@ -1,0 +1,98 @@
+syntax = "proto3";
+
+package nym.vpn;
+
+message Gateway {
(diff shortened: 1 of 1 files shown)


octol (approved):
I think this is a good initial definition, let's merge it so we can iterate on it

PR #212, opened 2024-03-26
Fix/uniffi android

zaneschepke:
Remove vpn status check from gateway calls because we need to call this API before we use the vpn

Fix gateways call bug where lib was returning country name enum instead of country code


zaneschepke:
Also, Country as an enum does not make much sense to me. It is error prone and I think this should really be a struct/dictionary in udl/uniffi, but I could not figure out how to get that working.

neacsu (approved):

PR #220, opened 2024-03-26
Temporarily add GH hosted windows runner

octol:


Bykow:
Not relevant anymore right ? @octol ?

PR #222, opened 2024-03-27
Extract GatewayClient to separate crate

octol:
Extract out `GatewayClient` to a separate crate `nym-gateway-directory`


nym-vpn-lib/src/platform/mod.rs
```diff
@@ -209,3 +209,30 @@ async fn get_low_latency_entry_country(
 
     Ok(country)
 }
+
+uniffi::custom_type!(EntryPoint, String);
+uniffi::custom_type!(ExitPoint, String);
```
neacsu, line 214: Suggestion: Maybe put this up with the rest?

Cargo.toml
```diff
@@ -1,6 +1,11 @@
 [workspace]
 resolver = "2"
-members = ["nym-vpn-lib", "nym-vpn-cli", "nym-vpnd"]
+members = [
+    "crates/nym-gateway-directory",
```
neacsu, line 4: Is the `crates` naming usual for rust projects? `common`, like in the mixnet repo, sounds more familiar, but I might just be melancholic here :D

neacsu (approved):

nym-vpn-lib/src/platform/mod.rs
octol: Done!

PR #246, opened 2024-04-05
Import credentials command

octol:
Add import-credentials command to cli client


nym-vpn-cli/src/commands.rs
```diff
@@ -24,31 +24,32 @@ fn pretty_build_info_static() -> &'static str {
...
 #[derive(Parser)]
 #[clap(author = "Nymtech", version, about, long_version = pretty_build_info_static())]
 pub(crate) struct CliArgs {
     /// Path pointing to an env file describing the network.
     #[arg(short, long, value_parser = check_path)]
     pub(crate) config_env_file: Option<PathBuf>,
 
+    /// Path to the data directory of the mixnet client.
+    #[arg(long)]
+    pub(crate) config_path: Option<PathBuf>,
```
neacsu, line 36: Shouldn't this be part of the `ImportCredentialArgs`, and non-optional, since we unwrap it later?

nym-vpn-cli/src/main.rs
```diff
@@ -124,6 +128,22 @@ async fn run_vpn(args: commands::RunArgs) -> Result<()> {
     Ok(())
 }
 
+async fn import_credential(
+    args: commands::ImportCredentialArgs,
+    config_path: Option<PathBuf>,
+) -> Result<()> {
+    let config_path = config_path.expect("config path not set");
```
neacsu, line 135: Seems like no reason to use it as `Option` here

nym-vpn-cli/src/main.rs
```diff
@@ -124,6 +128,22 @@ async fn run_vpn(args: commands::RunArgs) -> Result<()> {
...
+    let raw_credential = match args.credential_type.credential_data {
+        Some(data) => data,
+        None => fs::read(
+            args.credential_type
+                .credential_path
+                .expect("unreachable according to clap"),
```
neacsu, line 141: In case we add a third type there, this would be reachable (although caught early). Not sure if there's a better way to handle clap restrictions (maybe something like a named match on `ImportCredentialType` where you can check all possible combinations of the arguments, so that we get compile time checks in case of modifications?)

nym-vpn-cli/src/commands.rs
octol: it's also used in the run command, for lookup up keys

nym-vpn-cli/src/main.rs
octol: Yes you're right, I'll add proper error handling here

nym-vpn-cli/src/main.rs
octol: Yes good point, I'll map the clap struct to an enum to catch this

neacsu (approved):

PR #257, opened 2024-04-09
Send structured android messages in log

octol:
Send structured (JSON) messages in log for the Android frontend to consume. In particular, send a message if the VPN stops

### DIFF
--- a/nym-vpn-lib/src/platform/android.rs
+++ b/nym-vpn-lib/src/platform/android.rs
@@ -335,7 +335,7 @@ impl From<&crate::Error> for AndroidControlMessage {
     }
 }
 
-fn send_android_control_message(exit_status_msg: &NymVpnExitStatusMessage) {
+fn print_android_control_message(exit_status_msg: &NymVpnExitStatusMessage) {
     let android_control_message: AndroidControlMessage = exit_status_msg.into();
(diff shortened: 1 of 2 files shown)


nym-vpn-lib/src/platform/android.rs
```diff
@@ -288,3 +288,65 @@ fn is_public_ip(addr: IpAddr) -> bool {
...
+            },
+            _ => AndroidControlMessage {
+                kind: AndroidControlMessageType::Error,
+                message: Some(error.to_string()),
+            },
+        }
+    }
+}
+
+fn send_android_control_message(exit_status_msg: &NymVpnExitStatusMessage) {
```
neacsu, line 338: `send` naming put my mind into looking for a different code here, until I realised it's actually indirectly "sent" because kotlin reads and interprets those changes. I think a better name would be `print` though

neacsu (commented):
It looks good to me, but let's wait for Zane's input on this as well, as it would need some implementation on the kotlin side as well

nym-vpn-lib/src/platform/android.rs
octol: ah yes good point

octol:
I'm going to close this in favour of https://github.com/nymtech/nym-vpn-client/pull/260, for now at least

PR #274, opened 2024-04-10
Fix paths-ignore in desktop ci workflows

octol:


doums (approved):
Looks overkill and crazy to me triggering JS/Rust desktop CI on everything which is not "cli" changes, like not a good idea

but as discussed with  @octol this is what he wants

octol:
> Looks overkill and crazy to me triggering JS/Rust desktop CI on everything which is not "cli" changes, like not a good idea
> 
> but as discussed with @octol this is what he wants

Correct, the previous rules led to broken builds and manual intervention. Better to let the machines work for you instead

PR #284, opened 2024-04-11
feat: add nym-vpn-proto crate lib

doums:


crates/nym-vpn-proto/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub mod pb {
```
octol, line 1: I don't think we need this?

octol (approved):
Approved with the recommendation to skip the pb module

nym-vpnc/Cargo.toml
```diff
@@ -11,13 +11,14 @@ license.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap.workspace = true
+clap = { workspace = true, features = ["derive"] }
 prost.workspace = true
-tokio.workspace = true
+tokio = { workspace = true, features = ["full"]}
```
octol, line 16: did it not work without `full`? Note that adding more features can increase compile time even if the components are not used

PR #286, opened 2024-04-11
filter exit gateways by build time and version

zaneschepke:
Add filter to exit gateways to filter by both version and build time

### DIFF
--- a/crates/nym-gateway-directory/src/entries/described_gateway.rs
+++ b/crates/nym-gateway-directory/src/entries/described_gateway.rs
@@ -43,7 +43,6 @@ impl DescribedGatewayWithLocation {
                 .build_timestamp
                 .parse::<DateTime<Utc>>()
                 .ok()
-                .map_or(None, |d| Some(d))
         })?
     }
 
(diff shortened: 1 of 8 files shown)


crates/nym-gateway-directory/src/entries/described_gateway.rs
```diff
@@ -23,6 +27,36 @@ impl DescribedGatewayWithLocation {
...
 
+    pub fn has_current_api_version(&self) -> bool {
+        self.has_v6_exit_build_timestamp() && self.has_current_build_version()
+    }
+
+    fn has_v6_exit_build_timestamp(&self) -> bool {
+        self.gateway
+            .self_described
+            .as_ref()
+            .map(|d| {
```
octol, line 38: Suggestion: 
.filter_map for parsing step
Then
.filter for date selection 
To keep it flat and non-nested 

Also you can split those two step into two functions

fn build_timestamp -> timestamp 

fn is_current_build -> bool

octol (approved):
Looks good! Left some suggestions

PR #297, opened 2024-04-16
refactor(desktop): grpc client

doums:
switch to grpc daemon split 😈  _wip_


doums:
merge handled by #368

PR #302, opened 2024-04-17
Credential check

octol:


nym-vpn-lib/src/credentials.rs
```diff
@@ -39,3 +44,128 @@ pub async fn import_credential_file(credential_file: PathBuf, data_path: PathBuf
     let raw_credential = fs::read(credential_file)?;
     import_credential(raw_credential, data_path).await
 }
+
+pub async fn check_imported_credential(data_path: PathBuf, gateway_id: &str) -> anyhow::Result<()> {
```
jstuczyn, line 48: but this won't necessarily check the latest imported credential. because `fetch_valid_credential` will call `get_next_unspent_credential` which, from what i recall, will get a credential with lowest id

nym-vpn-lib/src/mixnet_connect.rs
```diff
@@ -289,6 +290,14 @@ pub(crate) async fn setup_mixnet_client(
     debug!("mixnet client has wireguard_mode={enable_wireguard}");
     let mixnet_client = if let Some(path) = mixnet_client_key_storage_path {
         debug!("Using custom key storage path: {:?}", path);
+        let gateway_id = mixnet_entry_gateway.to_base58_string();
+        if let Err(err) =
+            credentials::check_imported_credential(path.to_path_buf(), &gateway_id).await
+        {
+            // UGLY: flow needs to restructured to sort this out, but I don't want to refactor all that just before release.
+            task_client.mark_as_success();
```
octol, line 298: `.disarm()` would be more descriptive ...

nym-vpn-lib/src/credentials/check.rs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+
+    // TODO: verify?
+
+    Ok(())
+}
+
+pub async fn check_credential_base58(credential: &str) -> Result<(), CredentialError> {
```
octol, line 42: @zaneschepke I'm thinking this is what you'll want to call

nym-vpn-lib/src/mixnet_connect.rs
jstuczyn: then update `nym` dependency version and call it : )

nym-vpn-lib/src/credentials/check.rs
```diff
@@ -0,0 +1,107 @@
...
+            debug!("credential is a free pass");
+            if freepass_info.expired() {
+                return Err(CredentialError::FreepassExpired {
+                    expiry_date: freepass_info.expiry_date().to_string(),
+                });
+            }
+        }
+    }
+
+    // TODO: verify?
```
jstuczyn, line 37: we kinda need to remember to include this later on

nym-vpn-lib/src/credentials/check.rs
```diff
@@ -0,0 +1,107 @@
...
+        .map_err(|err| CredentialError::FailedToDecodeBase58Credential { source: err })?;
+    check_raw_credential(raw_credential).await
+}
+
+pub async fn check_credential_file(credential_file: PathBuf) -> Result<(), CredentialError> {
+    let raw_credential = fs::read(credential_file)?;
+    check_raw_credential(raw_credential).await
+}
+
+pub async fn check_imported_credential(
```
jstuczyn, line 54: this doesn't seem to be used? 
plus it suffers from the same issue we had before -> it verifies the "oldest" available credential rather than the one you just imported

nym-vpn-lib/src/credentials/helpers.rs
```diff
@@ -0,0 +1,59 @@
...
+    let credential_db_path = storage_path.credential_database_path;
+    debug!("Credential store: {}", credential_db_path.display());
+    Ok((
+        nym_credential_storage::initialise_persistent_storage(credential_db_path.clone()).await,
+        credential_db_path,
+    ))
+}
+
+pub(super) fn get_nyxd_client() -> Result<QueryHttpRpcNyxdClient, CredentialError> {
+    let network = NymNetworkDetails::new_from_env();
```
jstuczyn, line 28: just to make sure: in mobile you're setting the env to have this info available, right?

nym-vpn-lib/src/credentials/check.rs
octol: It's called before connect

nym-vpn-lib/src/credentials/check.rs
zaneschepke: Correct

PR #333, opened 2024-04-25
feat(grpc): add health client gen

doums:
- add generation of service health client
- add nice tracing of grpc requests

minimum grpc tracing level can be set using `grpc` key in `RUST_LOG` env var

```
RUST_LOG=nym_vpnd=debug,grpc=trace
```

### DIFF
--- a/crates/nym-vpn-proto/build.rs
+++ b/crates/nym-vpn-proto/build.rs
@@ -5,10 +5,14 @@ use std::{env, path::PathBuf};
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // needed for reflection
-    let descriptor_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("vpn_descriptor.bin");
+    let vpn_fd = PathBuf::from(env::var("OUT_DIR").unwrap()).join("vpn_descriptor.bin");
+    tonic_build::configure()
+        .file_descriptor_set_path(vpn_fd)
+        .compile(&["../../proto/nym/vpn.proto"], &["../../proto/nym/"])?;
 
     tonic_build::configure()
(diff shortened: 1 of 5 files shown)


crates/nym-vpn-proto/build.rs
```diff
@@ -5,10 +5,14 @@ use std::{env, path::PathBuf};
...
+    let vpn_fd = PathBuf::from(env::var("OUT_DIR").unwrap()).join("vpn_descriptor.bin");
     tonic_build::configure()
-        .file_descriptor_set_path(descriptor_path)
+        .file_descriptor_set_path(vpn_fd)
         .compile(&["../../proto/nym/vpn.proto"], &["../../proto/nym/"])?;
+
+    tonic_build::configure()
+        // server implementation is handled by tonic-health crate
+        .build_server(false)
+        .compile(&["../../proto/grpc/health.proto"], &["../../proto/grpc/"])?;
```
octol, line 16: Suggestion: extract out the proto dir into a separate constant

nym-vpnd/Cargo.toml
```diff
@@ -28,6 +28,9 @@ tracing.workspace = true
 tonic-health.workspace = true
 tonic-reflection.workspace = true
 
+# match the version required by tonic
+http = "0.2.12"
```
octol, line 32: Suggestion: if we need to match tonic, have a look to see if maybe they re-export these types there

octol (approved):

nym-vpnd/Cargo.toml
doums: I had checked, but tonic seems not to reexport these types (unless I missed it)

PR #334, opened 2024-04-25
Add connect options for two-hop, background traffic etc

octol:
Add support for connection options

### DIFF
--- a/proto/nym/vpn.proto
+++ b/proto/nym/vpn.proto
@@ -39,6 +39,11 @@ message ExitNode {
 message ConnectRequest {
   EntryNode entry = 1;
   ExitNode exit = 2;
+  bool disable_routing = 3;
+  bool enable_two_hop = 4;
+  bool enable_poisson_rate = 5;
+  bool disable_background_cover_traffic = 6;
+  bool enable_credentials_mode = 7;
 }
 
 message ConnectResponse {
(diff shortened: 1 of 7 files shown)


proto/nym/vpn.proto
```diff
@@ -39,6 +39,11 @@ message ExitNode {
 message ConnectRequest {
   EntryNode entry = 1;
   ExitNode exit = 2;
+  bool disable_routing = 3;
+  bool enable_two_hop = 4;
+  bool enable_poisson_rate = 5;
+  bool disable_background_cover_traffic = 6;
+  bool enable_credentials_mode = 7;
```
doums, line 46: imo it would be better to expose all these options as `optional` (keeping the same behavior as with the client lib)
so clients would just have to set what they need

octol: For these booleans, the default is false, so making them optional wouldn't change anything as far as client interaction goes

doums: Declaring them optional will produce the defaults to be controlled by the server implementation, and so defaults are unified and non-ambiguous for clients
plus if client don't need to deal with these options, they just dont

based on the too long discussions we had and that you are very motivated to not using `optional` attribute, let's do as you want

tommyv1987 (approved):

PR #337, opened 2024-04-25
feat(desktop): getting countries from lib

zaneschepke:
### backend
- Add getting exit and entry countries from lib to desktop
- Add get low latency country from lib

### frontend
refactored to handle the loading state of the countries (while fetching them from the lib)
- it's fetching both entry/exit countries, in parallel and in a UI non-blocking way, that means it does not delay the initial startup sequence of the app (splash screen duration before main UI is shown and effective)
- in the UI there is a message for user "Country list is loading" on the initial load (when there is no country data cached yet and so no available data)


zaneschepke:
Is there a possibility to switch desktop to uniffi? 

A lot of code is copy pasta

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -1,21 +1,21 @@
...
     error::{CmdError, CmdErrorSource},
-    http::{
-        client::{HttpError, HTTP_CLIENT},
-        explorer_api, nym_api,
-    },
     states::{app::NodeLocation, SharedAppState},
 };
 
+const NYM_API_SANDBOX_URL: &str = "https://sandbox-nym-api1.nymtech.net/api";
+const NYM_EXPLORER_SANDBOX_URL: &str = "https://sandbox-explorer.nymtech.net/api";
```
octol, line 17: These shouldn't be needed

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -89,164 +89,74 @@ pub async fn get_node_location(
...
-            )
+#[instrument(skip_all)]
+async fn get_low_latency_entry_country() -> Result<Country, CmdError> {
+    let api_url = get_api_url()?;
+    let explorer_url = get_explorer_url()?;
+    let config = Config {
+        api_url,
+        explorer_url: Some(explorer_url),
+        harbour_master_url: None,
+    };
```
octol, line 114: I'd suggest
```rust
let config = Config::new_from_env();
```
that will set nym-api, explorer-api, and harbourmaster-api urls

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -89,164 +89,74 @@ pub async fn get_node_location(
 pub async fn get_countries(node_type: NodeType) -> Result<Vec<Country>, CmdError> {
     debug!("get_countries");
     match node_type {
-        NodeType::Entry => get_entry_countries().await,
-        NodeType::Exit => get_exit_countries().await,
+        NodeType::Entry => get_gateway_countries(false).await,
+        NodeType::Exit => get_gateway_countries(true).await,
     }
 }
```
octol, line 95: I would suggest we try to fetch both of these in the same call to avoid reaching out to harbourmaster twice (it's unfortunately slow). I think you should be able to use `GatewayClient::lookup_described_entry_and_exit_gateways_with_location()`.

doums:
> Is there a possibility to switch desktop to uniffi?
> 
> A lot of code is copy pasta

@zaneschepke desktop backend is already Rust, so using an FFI layer for rust to rust does not make much sense. If we're forced to duplicate code from uniffi it looks like more a code architecture issue.
My suggestion would be to extract (to any rust crate lib) the duplicate logic from the uniffi implementation, so both uniffi and desktop could use it.

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
zaneschepke: If this works, I'll change for uniffi as well.

nym-vpn-desktop/src-tauri/src/error.rs
```diff
@@ -32,6 +32,24 @@ impl CmdError {
...
+}
+
+impl From<nym_vpn_lib::gateway_directory::Error> for CmdError {
+    fn from(value: nym_vpn_lib::gateway_directory::Error) -> Self {
+        CmdError {
+            source: CmdErrorSource::InternalError,
+            message: value.to_string(),
+        }
+    }
+}
```
doums, line 51: @octol is all errors from `nym_vpn_lib::gateway_directory` and `nym_vpn_lib::error::Error` can be considered as "internal error" and not "user responsibility" (aka "caller") errors?

octol: The calls I've seen used here can only really fail for "internal" reasons, like the endpoint not returning the expected data. The exception I think would be if the API urls are not set correctly, that would probably qualify as a caller error

doums: Ok thx
Also, the trait implem for `nym_vpn_lib::error::Error` is not needed and not wanted, as there is plenty of errors from lib calls that can be qualified as "user" related errors
so @zaneschepke lets remove that one

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
zaneschepke: I think this is an optimization we will need to make later. It requires too much change on the clients to make this change right now.

doums:
I'll cleanup a bit

doums:
okay @zaneschepke I refactored a bit the frontend side to handle the loading state of the countries (while fetching them from the lib) so
- it's fetching both entry/exit countries, in parallel and in a UI non-blocking way, that means it does not delay the initial startup sequence of the app (splash screen duration before main UI is shown and effective)
- in the UI there is a message for user "Country list is loading" on the initial load (when there is no country data cached yet and so available data)

doums (approved):

octol (approved):

nym-vpn-desktop/src-tauri/src/commands/node_location.rs
```diff
@@ -68,7 +63,13 @@ pub async fn set_node_location(
 #[tauri::command]
 pub async fn get_fastest_node_location() -> Result<Country, CmdError> {
     debug!("get_fastest_node_location");
-    Ok(FASTEST_NODE_LOCATION.clone())
+    get_low_latency_entry_country().await.map_err(|e| {
```
octol, line 66: So I wonder about this, I hope this is not used in the UI to select the fastest country, but only for display purposes

doums: Fastest feature is still disabled in frontend, so it's not used yet

PR #363, opened 2024-04-29
Expose data path for credential db

rokas-ambrazevicius:
Expose path for credential db


nym-vpn-lib/src/lib.rs
```diff
@@ -232,6 +232,7 @@ impl NymVpn<MixnetVpn> {
         exit_point: ExitPoint,
         #[cfg(target_os = "android")] android_context: talpid_types::android::AndroidContext,
         #[cfg(target_os = "ios")] ios_tun_provider: Arc<dyn OSTunProvider>,
+        credential_data_path: Option<PathBuf>
```
neacsu, line 235: No need to add it as an argument here, you can just set it in the `sync_run_vpn` function later

neacsu (commented):
The .swift file should be in `nym-vpn-lib/uniffi/nym_vpn_lib.swift`, where there's already the file. Just overwrite that one and commit the diff

neacsu (approved):

PR #368, opened 2024-05-02
nymvpn-x

doums:
Introduces the next nymvpn desktop client, code name `nymvpn-x`

It's a rewrite of the previous client implementation. Switching to the new vpn**d** (aka daemon) split architecture, using `gRPC` for communication between vpn daemon and client


nymvpn-x/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+tracing-subscriber =  { version = "0.3.1", features = ["tracing-log", "env-filter"] }
+anyhow = "1.0"
+dotenvy = "0.15.7"
+thiserror = "1.0"
+ts-rs = { version = "7.0.0", features = ["chrono-impl"] }
+once_cell = "1.18.0"
+toml = "0.8.5"
+time = "0.3.9"
+# TODO use the crate once it's published, for now use local path
+nym-vpn-lib = { path = "../../nym-vpn-lib" }
```
octol, line 29: Don't think you need this now? I think you can use nym-gateway-directory directly and not have to worry about building wireguard and all that stuff

.github/workflows/ci-nymvpn-x-rust.yml
```diff
@@ -0,0 +1,116 @@
...
+            echo " ✗ unknown platform/arch [${{ matrix.os }}]"
+            exit 1
+          fi
+          echo " ✓ PLATFORM_ARCH: $platform_arch"
+          echo " ✓ TRIPLET: $triplet"
+          echo "PLATFORM_ARCH=$platform_arch" >> $GITHUB_ENV
+          echo "TRIPLET=$triplet" >> $GITHUB_ENV
+          echo "RUSTFLAGS=-L ${GITHUB_WORKSPACE}/build/lib/${triplet}" >> $GITHUB_ENV
+
+      - name: Install Go
```
octol, line 91: Can skip this

.github/workflows/ci-nymvpn-x-rust.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Install protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup MSBuild.exe
+        if: contains(matrix.os, 'windows')
+        uses: microsoft/setup-msbuild@v2
+
+      - name: Build wireguard
```
octol, line 105: Can skip

nymvpn-x/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+clap = { version = "4.5", features = ["derive"] }
+build-info = "0.0.34"
+sled = "0.34.7"
+strum = { version = "0.26", features = ["derive"] }
+tonic =  { version = "0.11", features = ["channel"] }
+bs58 = "0.5.1"
+
+# TODO Ugly workaround to force a working setup for nym-vpn-lib
+# We should get rid of this ASAP
+shadowsocks = { version = "~1.14.2" }
```
octol, line 43: Don't need these anymore

nymvpn-x/src-tauri/vcruntime140.dll
octol, line 1: Let's add binaries until we're certain we don't need them. It's an unfortunate limitation of git where binaries will bloat the repo

nymvpn-x/src-tauri/Cargo.toml
doums: to the only usage I see is there https://github.com/nymtech/nym-vpn-client/pull/368/files#diff-176db7941721b48c4835c8d375fb267849129e94b3056cc25dfdf5b2c4d8e37f

image

so it depends if we still need these calls or not?

.github/workflows/ci-nymvpn-x-rust.yml
```diff
@@ -0,0 +1,106 @@
+name: ci-nymvpn-x-rust
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - ".github/workflows/ci-nymvpn-x-rust.yml"
+      - "nymvpn-x/src-tauri/**"
+      - "crates/nym-vpn-proto/**"
```
octol, line 9: Make sure to include all crates you're using here, or the whole crates directory to make it more robust

nymvpn-x/src-tauri/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+mod events;
+mod fs;
+mod gateway;
+mod grpc;
+mod network;
+mod states;
+mod vpn_status;
+mod window;
+
+pub const APP_DIR: &str = "nym-vpn";
```
octol, line 42: nym-vpn-x, since this doesn't run as root it wont be able to read the filed in that directory if you have used the other client first

nymvpn-x/src-tauri/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+const APP_CONFIG_FILE: &str = "config.toml";
+const ENV_APP_NOSPLASH: &str = "APP_NOSPLASH";
+const VPND_DEFAULT_ENDPOINT: &str = "http://[::1]:53181";
```
octol, line 45: Let's use the socket by default, then we can easily restrict usage to privileged users only

nymvpn-x/src-tauri/src/country.rs
```diff
@@ -0,0 +1,39 @@
...
+});
+
+// TODO use countries requested from the backend instead of hardcoded ones
+pub static DEFAULT_ENTRY_COUNTRY: Lazy<Country> = Lazy::new(|| Country {
+    code: String::from("FR"),
+    name: String::from("France"),
+});
+pub static DEFAULT_EXIT_COUNTRY: Lazy<Country> = Lazy::new(|| Country {
+    code: String::from("CH"),
+    name: String::from("Switzerland"),
```
octol, line 19: This caused an error for me, with no Swiss exit gateways available

nymvpn-x/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "nymvpn-x"
```
octol, line 2: I think we should stick to `nym-vpn-x`, since we use "NymVPN" and also that's what everything else is using

nymvpn-x/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+tracing = "0.1"
+tracing-subscriber =  { version = "0.3.1", features = ["tracing-log", "env-filter"] }
+anyhow = "1.0"
+dotenvy = "0.15.7"
+thiserror = "1.0"
+ts-rs = { version = "7.0.0", features = ["chrono-impl"] }
+once_cell = "1.18.0"
+toml = "0.8.5"
+time = "0.3.9"
+futures = "0.3.15"
```
octol, line 28: I don't think this is used?

nymvpn-x/src-tauri/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+reqwest = { version = "0.11", features = ["json"] }
```
octol, line 29: This too, is it used?

nymvpn-x/src-tauri/src/country.rs
doums: yea good one, but as you can see given the `TODO` comment above, it will be covered in a later PR

nymvpn-x/src-tauri/Cargo.toml
doums: This app is a standalone client, and as a standalone I don't see any reason to follow naming of other nym apps. As you said the official name of the app is `NymVPN`, so it translates well to -> `nymvpn`-x for the underlying rust crate.

nymvpn-x/src-tauri/src/main.rs
doums: Agree, but lets cover that in a following PR as it requires a bit more work

nymvpn-x/src-tauri/Cargo.toml
octol: I disagree, I think `nym-vpn-x` is better and matches all other crate names have. Also imho the directory should match.  Actually, thinking about it now even better would be `nym-vpn-ui` for the directory, and fits better with the rest of the repo, and puts it in context with the daemon and stuff like that. We can keep the package name and the name that the resulting appimage is generated from to `nymvpn-x` though

octol (changes requested):

octol (approved):
Approved. 
Naming has been discussed extensively elsewhere, but I still maintain that it would be better if the resulting executable was called `nym-vpn-x` for consistency with everything else. Even small stuff like tab-complete in the terminal will be impacted by this when you have `nym-vpnd` and `nym-vpnc` also installed in the user path

PR #374, opened 2024-05-03
fix(vpnx): connect request

doums:
- fix `disconnect` state transitions
- add grpc ipc transport support

### DIFF
--- a/nym-vpn-x/src-tauri/src/commands/connection.rs
+++ b/nym-vpn-x/src-tauri/src/commands/connection.rs
@@ -149,7 +149,7 @@ pub async fn connect(
         enable_two_hop: two_hop_mod,
         enable_poisson_rate: false,
         disable_background_cover_traffic: false,
-        enable_credentials_mode: true,
+        enable_credentials_mode: false,
         dns,
     });
 
@@ -178,13 +178,14 @@ pub async fn disconnect(
     grpc: State<'_, Arc<GrpcClient>>,
 ) -> Result<ConnectionState, CmdError> {
     debug!("disconnect");
-    let mut app_state = state.lock().await;
+    let app_state = state.lock().await;
     if !matches!(app_state.state, ConnectionState::Connected) {
         return Err(CmdError::new(
             CmdErrorSource::CallerError,
             &format!("cannot disconnect from state {:?}", app_state.state),
         ));
     };
+    drop(app_state);
 
     let mut vpnd = grpc.vpnd().await.map_err(|_| {
         warn!("not connected to the daemon");
@@ -193,26 +194,20 @@ pub async fn disconnect(
 
     // switch to "Disconnecting" state
     trace!("update connection state [Disconnecting]");
+    let mut app_state = state.lock().await;
     app_state.state = ConnectionState::Disconnecting;
     app.emit_disconnecting();
+    drop(app_state);
 
     let request = Request::new(DisconnectRequest {});
     let response = vpnd.vpn_disconnect(request).await.map_err(|e| {
         let error_msg = format!("failed to disconnect: {e}");
         error!("grpc vpn_disconnect: {}", e);
-        // TODO handle error properly
-        // just switch back to "Connected" state for now
-        app_state.state = ConnectionState::Connected;
         CmdError::new(CmdErrorSource::DaemonError, &error_msg)
     })?;
     debug!("grpc response: {:?}", response);
 
-    trace!("update connection state [Disconnected]");
-    app_state.state = ConnectionState::Disconnected;
-    app_state.connection_start_time = None;
-    app.emit_disconnected(None);
-
-    Ok(app_state.state.clone())
+    Ok(ConnectionState::Disconnecting)
 }
 
 #[instrument(skip_all)]
--- a/nym-vpn-x/src-tauri/src/commands/connection.rs
+++ b/nym-vpn-x/src-tauri/src/commands/connection.rs
@@ -196,8 +196,8 @@ pub async fn disconnect(
     trace!("update connection state [Disconnecting]");
     let mut app_state = state.lock().await;
     app_state.state = ConnectionState::Disconnecting;
-    app.emit_disconnecting();
     drop(app_state);
+    app.emit_disconnecting();
 
     let request = Request::new(DisconnectRequest {});
     let response = vpnd.vpn_disconnect(request).await.map_err(|e| {
--- a/nym-vpn-x/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-x/src-tauri/src/grpc/client.rs
@@ -1,18 +1,32 @@
+use std::path::PathBuf;
+
 use anyhow::{anyhow, Result};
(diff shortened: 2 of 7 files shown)


nym-vpn-x/src-tauri/src/commands/connection.rs
```diff
@@ -178,13 +178,14 @@ pub async fn disconnect(
...
     debug!("disconnect");
-    let mut app_state = state.lock().await;
+    let app_state = state.lock().await;
     if !matches!(app_state.state, ConnectionState::Connected) {
         return Err(CmdError::new(
             CmdErrorSource::CallerError,
             &format!("cannot disconnect from state {:?}", app_state.state),
         ));
     };
+    drop(app_state);
```
octol, line 188: I think having the scope track the ownership instead of explicit drop would be preferable

nym-vpn-x/src-tauri/src/commands/connection.rs
```diff
@@ -193,26 +194,20 @@ pub async fn disconnect(
 
     // switch to "Disconnecting" state
     trace!("update connection state [Disconnecting]");
+    let mut app_state = state.lock().await;
     app_state.state = ConnectionState::Disconnecting;
     app.emit_disconnecting();
+    drop(app_state);
```
octol, line 200: See other comment

octol (changes requested):

doums:
@octol regarding IPC transport, it seems like the crate `parity-tokio-ipc ` could have some issue to compile and work on Windows 11, I'm a bit concerned about it

see https://github.com/paritytech/parity-tokio-ipc/issues/39

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -22,37 +36,61 @@ pub enum VpndStatus {
...
+            Transport::Http(endpoint) => {
+                NymVpndClient::connect(endpoint.clone()).await.map_err(|e| {
+                    warn!("failed to connect to the daemon: {}", e);
+                    anyhow!("failed to connect to the daemon: {}", e)
+                })
+            }
+            Transport::Ipc(socket) => {
+                let channel = get_channel(socket.clone()).await.map_err(|e| {
+                    warn!("failed to connect to the daemon: {}", e);
+                    anyhow!("failed to connect to the daemon: {}", e)
```
octol, line 69: This pattern of both having a log warning and return an error for the same thing seems like a code smell. Log these things at the top-level when you need to

This comment applies to all of them

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -132,3 +170,42 @@ impl From<ServingStatus> for VpndStatus {
...
+
+impl Default for Transport {
+    fn default() -> Self {
+        Transport::Ipc(DEFAULT_SOCKET_PATH.into())
+    }
+}
+
+impl From<(&AppConfig, &Cli)> for Transport {
+    fn from((config, cli): (&AppConfig, &Cli)) -> Self {
+        let http_mode = config.grpc_http_mode.unwrap_or(cli.grpc_http_mode);
```
octol, line 191: Convention is usually that CLI should override config

nym-vpn-x/src-tauri/src/commands/connection.rs
octol: Imho this is more than a style thing, resource ownership tracking scope is a relatively core Rust idiom with many advantages over explicit drop

Using scope it's more robust and less risk of mistakes, especially as the code evolves and with early returns

octol (changes requested):

octol:
> @octol regarding IPC transport, it seems the crate `parity-tokio-ipc` could have some issue to compile and work on Windows 11, I'm a bit concerned about it
> 
> see [paritytech/parity-tokio-ipc#39](https://github.com/paritytech/parity-tokio-ipc/issues/39)

I woudn't worry about it, for windows it's possible we'll use something platform specific anyway

nym-vpn-x/src-tauri/src/grpc/client.rs
doums: yes it is, but here `cli.grpc_http_mode` is a boolean flag and so it is always set (default to false if user dont set the flag)
So we cant check if the user explicilty set the flag to false or didnt set it at all, and so it can't take precedence over the config option

nym-vpn-x/src-tauri/src/grpc/client.rs
octol: The above comment is primarily for future PRs, and using anyhow here seems like a quick hack for now until we setup error enums for these components. So yeah something to keep that in mind for follow up PRs

nym-vpn-x/src-tauri/src/commands/connection.rs
doums: Like you said it is your opinion, and I dont share it.
For example take a look at the official Mutex doc -> https://doc.rust-lang.org/std/sync/struct.Mutex.html

> To unlock a mutex guard sooner than the end of the enclosing scope, either create an inner scope or drop the guard manually.

```rust
// We drop the `data` explicitly because it's not necessary anymore and the
// thread still has work to do. This allows other threads to start working on
// the data immediately, without waiting for the rest of the unrelated work
// to be done here.
//
// It's even more important here than in the threads because we `.join` the
// threads after that. If we had not dropped the mutex guard, a thread could
// be waiting forever for it, causing a deadlock.
// As in the threads, a block could have been used instead of calling the
// `drop` function.
drop(data);
```

Where is it stated that one should "avoid" using drop because it's bad?  Or even is discouraged in favor to use inner scopes?

So when you said (tldr) "using `drop` is bad, it's not rust idiom", it's your own take.

nym-vpn-x/src-tauri/src/grpc/client.rs
doums: TLDR:

`config.grpc_http_mode.unwrap_or(cli.grpc_http_mode)` is the same as

if cli.grpc_http_mode is true -> use it
if not fallback to the config value

octol: grpc_http_mode is Option<bool> right? Doesn't that mean that if the user has it set in the config then the cli will be ignored?

doums: No it is `bool` as any boolean flag when using clap, mean the flag is either set or not

doums: > Doesn't that mean that if the user has it set in the config then the cli will be ignored?

yes, but what's the deal? The default config is empty by default, use have to manually customize it

octol: > No it is bool as any boolean flag when using clap, mean the flag is either set or not

I mean the value from the config file is `Option<bool>`, leading to the surprising behaviour where the cli does nothing if it's set in the config file

PR #375, opened 2024-05-03
Block in place when resetting dns

octol:


doums (approved):
yup fix the runtime panic

PR #380, opened 2024-05-06
Zip macOS CLI to support SPM

rokas-ambrazevicius:
Need a `.zip` for macOS lib. SPM does not support `.tar.gz`.


neacsu (approved):

octol (changes requested):
Was talking to Rokas about this and let's hold off merging for a bit

PR #384, opened 2024-05-08
Status streaming in nym-vpnd

octol:
- Start on streaming connection status
- wip
- Initial working status update streaming

### DIFF
--- a/nym-vpnc/src/main.rs
+++ b/nym-vpnc/src/main.rs
@@ -4,7 +4,7 @@
 use anyhow::Result;
 use clap::Parser;
 use nym_vpn_proto::{
-    ConnectRequest, DisconnectRequest, ImportUserCredentialRequest, StatusRequest,
+    ConnectRequest, DisconnectRequest, Empty, ImportUserCredentialRequest, StatusRequest,
 };
 use vpnd_client::ClientType;
 
@@ -33,6 +33,7 @@ async fn main() -> Result<()> {
         Command::ImportCredential(ref import_args) => {
             import_credential(client_type, import_args).await?
         }
+        Command::ListenToStatus => listen_to_status(client_type).await?,
     }
     Ok(())
 }
@@ -94,4 +95,17 @@ async fn import_credential(
 
 fn parse_encoded_credential_data(raw: &str) -> bs58::decode::Result<Vec<u8>> {
     bs58::decode(raw).into_vec()
+}
+
+async fn listen_to_status(client_type: ClientType) -> Result<()> {
+    let mut client = vpnd_client::get_client(client_type).await?;
+    let request = tonic::Request::new(Empty {});
+    let mut stream = client
+        .listen_to_connection_status(request)
+        .await?
+        .into_inner();
+    while let Some(response) = stream.message().await? {
+        println!("{:?}", response);
+    }
+    Ok(())
 }
--- a/nym-vpnd/src/command_interface/listener.rs
+++ b/nym-vpnd/src/command_interface/listener.rs
@@ -30,6 +30,7 @@ enum ListenerType {
 
 pub(super) struct CommandInterface {
     vpn_command_tx: UnboundedSender<VpnServiceCommand>,
+    // vpn_status_rx: tokio::sync::mpsc::Receiver<VpnServiceStatusResult>,
     listener: ListenerType,
 }
 
--- a/nym-vpnd/src/command_interface/listener.rs
+++ b/nym-vpnd/src/command_interface/listener.rs
@@ -5,8 +5,10 @@ use std::{
     fs,
     net::SocketAddr,
     path::{Path, PathBuf},
+    pin::Pin,
 };
 
+use futures::StreamExt;
 use nym_vpn_lib::{
     gateway_directory::{EntryPoint, ExitPoint},
     NodeIdentity, Recipient,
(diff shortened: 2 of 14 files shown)


proto/nym/vpn.proto
```diff
@@ -129,15 +102,7 @@ service NymVpnd {
...
-  // // and return to disconnected state
-  // rpc Cancel (Empty) returns (Empty) {}
-  // rpc GetEntryNodeLocations (Empty) returns (LocationListResponse) {}
-  // rpc GetExitNodeLocations (Empty) returns (LocationListResponse) {}
-  // rpc GetFastestEntryGateway (Empty) returns (GatewayResponse) {}
-  // rpc GetFastestExitGateway (Empty) returns (GatewayResponse) {}
-  // rpc SetVpnMode (SetVpnModeRequest) returns (Empty) {}
-  // rpc GetVpnMode (Empty) returns (GetVpnModeResponse) {}
+  rpc ListenToConnectionStateChanges (Empty) returns (stream ConnectionStateChange) {}
+  rpc ListenToConnectionStatus (Empty) returns (stream ConnectionStatusUpdate) {}
```
doums, line 106: I'd suggest to use shorter names for these new rpc methods, like using `ConnectionState` and `ConnectionStatus` as it's obvious reading the method's signatures and return types it's rpc streaming (and so avoiding redundancy)

as ref this is the official proto def of grpc tutorial:
https://github.com/grpc/grpc-go/blob/master/examples/route_guide/routeguide/route_guide.proto

nym-vpnc/src/main.rs
```diff
@@ -95,3 +96,16 @@ async fn import_credential(
...
+    let mut client = vpnd_client::get_client(client_type).await?;
+    let request = tonic::Request::new(Empty {});
+    let mut stream = client
+        .listen_to_connection_status(request)
+        .await?
+        .into_inner();
+    while let Some(response) = stream.message().await? {
+        println!("{:?}", response);
+    }
+    Ok(())
```
jstuczyn, line 110: should we perhaps log an additional message (maybe to stderr, or as an `error!` log) when we exit the loop, i.e. the channel closes? My understanding is that we never expect it to shutdown, unless the app itself is going off

nym-vpnd/src/command_interface/listener.rs
```diff
@@ -39,6 +49,8 @@ impl CommandInterface {
     ) -> Self {
         Self {
             vpn_command_tx,
+            status_tx: tokio::sync::broadcast::channel(10).0,
+            _connection_state_tx: tokio::sync::broadcast::channel(10).0,
```
jstuczyn, line 53: that looks rather iffy to put channels there for the sake of putting them (since we're dropping half of them immediately); perhaps the `CommandInterface` should hold an `Option`?

nym-vpnd/src/command_interface/listener.rs
```diff
@@ -176,6 +200,49 @@ impl NymVpnd for CommandInterface {
             success: status.is_success(),
         }))
     }
+
+    type ListenToConnectionStatusStream = Pin<
```
jstuczyn, line 204: perhaps use `BoxStream` alias: https://docs.rs/futures/latest/futures/stream/type.BoxStream.html for slightly better readibility

nym-vpnd/src/service/exit_listener.rs
```diff
@@ -18,22 +19,25 @@ impl VpnServiceExitListener {
...
-                    nym_vpn_lib::NymVpnExitStatusMessage::Failed(err) => {
+                    nym_vpn_lib::NymVpnExitStatusMessage::Failed(ref err) => {
                         error!("VPN exit: fail: {err}");
                         self.set_shared_state(VpnState::ConnectionFailed(err.to_string()));
+                        listener_vpn_exit_tx.send(exit_res).ok();
                     }
                 },
                 Err(err) => {
-                    error!("exit listener fail: {err}");
+                    error!("Exit listener: {err}");
```
jstuczyn, line 40: I thought you were the one wanting our errors to start with lowercase : )

nym-vpnd/src/service/status_listener.rs
```diff
@@ -20,33 +21,39 @@ impl VpnServiceStatusListener {
...
+        mut listener_vpn_status_tx: StatusSender,
+    ) {
         tokio::spawn(async move {
             while let Some(msg) = vpn_status_rx.next().await {
-                self.handle_status_message(msg).await;
+                let listener_msg = self.handle_status_message(msg).await;
+
+                // Forward the status message to the command listener so that it can provide these
+                // on its streaming endpoints
+                listener_vpn_status_tx.send(listener_msg).await.unwrap();
```
jstuczyn, line 56: perhaps `.ok()` that sucker so it wouldnt cause a panic during shutdown?

jstuczyn (approved):

nym-vpnd/src/service/exit_listener.rs
octol: Yep, for error enums, so that they can be included as substrings in e.g log entries like these :)

nym-vpnc/src/main.rs
octol: it's not handling the `Ok(None)` case but doesn't the `?` early return cover the case you are mentioning?

proto/nym/vpn.proto
octol: A distinction that is worth taking into account here is that unlike e.g `ListFeatures` in that example these streams are indefinite, while `ListFeatures` as far as I can tell will stream a set of entries from an existing container. This would not be apparent from the method's signature and I think is significant enough to include in the name

PR #387, opened 2024-05-10
Stream connection state changes in nym-vpnd

octol:
Create a RPC function for nym-vpnd that streams connection state changes

### DIFF
--- a/nym-vpnd/src/command_interface/start.rs
+++ b/nym-vpnd/src/command_interface/start.rs
@@ -19,7 +19,7 @@ use super::{
 };
 use crate::{
     cli::CliArgs,
-    service::{VpnServiceCommand, VpnServiceStatusResult, VpnState},
+    service::{VpnServiceCommand, VpnServiceStateChange},
 };
 
 fn grpc_span(req: &http::Request<()>) -> Span {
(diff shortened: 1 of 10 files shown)


nym-vpnd/src/command_interface/start.rs
```diff
@@ -100,17 +119,25 @@ pub(crate) fn start_command_interface(
     let uri_addr = default_uri_addr();
 
     let handle = std::thread::spawn(move || {
+        // Explicitly create a multu-threaded runtime for the command interface.
+        // We should consider also explictly setting the number of threads here to something like
+        // max(num_vcpu, 4) or something like that.
```
jstuczyn, line 124: that's a good point, should we perhaps deterine the vcpu split in `main.rs` and assign them accordingly to internal runtimes?

nym-vpnd/src/service/vpn_service.rs
```diff
@@ -92,6 +92,8 @@ impl VpnServiceDisconnectResult {
     }
 }
 
+// Respond with the current state of the VPN service. This is currently almos the same as VpnState,
```
jstuczyn, line 95:
```suggestion
// Respond with the current state of the VPN service. This is currently almost the same as VpnState,
```

nym-vpnd/src/command_interface/start.rs
octol: It's an interesting challenge I've been pondering about, I'm leaning towards we should have more threads than vcpu. But it will depend, probabaly

PR #390, opened 2024-05-13
ci: add wf for vpnx release

doums:
- add a wf for vpnx release (new desktop client)
- only support linux and macos
- based on the current production release wf

### DIFF
--- a/.github/workflows/release-nym-vpn-x.yml
+++ b/.github/workflows/release-nym-vpn-x.yml
@@ -1,0 +1,452 @@
+name: release-nym-vpn-x
+on:
+  schedule:
+    - cron: "4 4 * * *"
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Tag name for release"
+        required: false
+        default: nightly-vpn-x
+      publish:
+        type: boolean
+        description: Publish Release
+        required: true
+        default: false
+  push:
+    tags:
+      - nym-vpn-x-v[0-9]+.[0-9]+.[0-9]+*
+
+env:
+  CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


.github/workflows/release-nym-vpn-x.yml
```diff
@@ -0,0 +1,452 @@
...
+        run: |
+          ls -la ./
+          rm -rf ./* || true
+          rm -rf ./.??* || true
+          ls -la ./
+
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
```
octol, line 55: In the past this has bitten us were jobs reused old files, so imho let's skip sparse checkout

octol (changes requested):

.github/workflows/release-nym-vpn-x.yml
doums: these are the only nym deps
image

so it's fine (should be)

octol: Well, should be, but there is also no performance advantage to doing a sparse checkout at least for our repo. So imho we should take the safe(r) route here

doums:
Do we need different secrets for tauri signing and apple singing stuff?

.github/workflows/release-nym-vpn-x.yml
doums: ok forgot "should be", it's just fine

octol: imho let's remove the sparse please

zaneschepke (approved):

PR #395, opened 2024-05-14
add admin dialog

zaneschepke:


doums:
I don't understand why "Admin privileges" would be required on the new client since the VPN connection is delegated to the daemon (running as Windows service, which will need admin privileges indeed).

Unless I'm missing something, the client app does not require any admin privileges to work.

zaneschepke:
> I don't understand why "Admin privileges" would be required on the new client since the VPN connection is delegated to the daemon (running as Windows service, which will need admin level indeed).
> 
> Unless I'm missing something, the client app does not require any admin level to work.

I think this admin privileges is supposed to be for the installer to install the service?

doums:
> I think this admin privileges is supposed to be for the installer to install the service?

ah yea make sense

PR #397, opened 2024-05-14
Handle errors for importing credentials in nym-vpnd

octol:
Return errors occuring when importing credentials in nym-vpnd

This unfortunately involves a lot of mapping between error enums into something suitable for IPC


proto/nym/vpn.proto
```diff
@@ -95,6 +95,29 @@ message ImportUserCredentialRequest {
...
+    FAILED_TO_QUERY_CONTRACT = 8;
+  }
+
+  ImportErrorType kind = 1;
+
+  // Detailed error message for logging and debugging
+  string message = 2;
+
+  // Optional additional details
+  map<string, string> details = 3;
```
doums, line 120: use the `optional` attribute for optional field

nym-vpnd/src/command_interface/credential_import.rs
```diff
@@ -0,0 +1,70 @@
...
+                message: err.to_string(),
+                details: Default::default(),
+            },
+            ImportCredentialError::DeserializationFailure {
+                reason: _,
+                ref location,
+            } => ProtoImportError {
+                kind: ImportErrorType::DeserializationFailure as i32,
+                message: err.to_string(),
+                details: hashmap! { "location".to_string() => location.to_string_lossy().to_string() },
```
jstuczyn, line 38: Do we actually need a separate crate and dep for this macro? maybe we could have a helper function like `fn into_map(k: impl<Into<String>>, v: impl<Into<String>) -> HashMap<String, String>`, then you could remove those `to_string()` and instead call it via `details: into_map("location", location)`; alternatively, but that's just thinking aloud because it would be an overkill: some trait + macro to automatically pull the field name


Also I'm not entirely convinced by storing structured data in an unstructured hashmap but I guess that's the limitation of protobuf for us...

nym-vpnd/src/service/error.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum ImportCredentialError {
+    #[error("vpn is connected")]
+    VpnRunning,
+
+    #[error("credential already imported")]
+    CredentialAlreadyImported,
+
+    #[error("generic error: {0}")]
+    Generic(String),
```
jstuczyn, line 18: that's a bit obscure error enum, but I guess we gotta do what we gotta do.
I just have a personal bias against error enum variants that just wrap around a string, but I recognise sometimes we have no better options and I think this might be one of those instances

nym-vpnd/src/service/error.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn from(err: CredentialError) -> Self {
+        let mut error = ImportCredentialError::Generic(err.to_string());
+        match err {
+            CredentialError::FailedToSetupStoragePaths { .. } => {}
+            CredentialError::FailedToCreateCredentialStoreDirectory { .. } => {}
+            CredentialError::FailedToReadCredentialStoreMetadata { .. } => {}
+            CredentialError::FailedToSetCredentialStorePermissions { .. } => {}
+            CredentialError::FailedToInitializePersistentStorage { .. } => {}
+            CredentialError::FreepassExpired { expiry_date } => {
+                error = ImportCredentialError::FreepassExpired {
```
jstuczyn, line 53: why do we mutate the `error` defined at the beginning rather than just returning here directly? all `mut` are spooky : )

jstuczyn: (then at the end, you could have a catch-all branch like `CredentialError::FailedToReadCredentialFile { .. } | CredentialError::FailedToGetNextUsableCredential { .. } | ... => ImportCredentialError::Generic(err.to_string())`

jstuczyn (approved):

nym-vpnd/src/command_interface/credential_import.rs
octol: So yeah, this was my initial approach, but the complexity started to pile up once I wanted to have multiple entries, `into_map([(k1,v1),(k2,v2)])`, where v1 and v2 are different types (but both implement to_string)

nym-vpnd/src/service/error.rs
octol: having explicit `return` everywhere ended up being verbose, and I kinda wanted to list them all explicitly with the intention to incrementally start handling all of them.
However, I'm instead reworking the upstream error now to try to sort this out without this big nasty match

proto/nym/vpn.proto
octol: `optional` is not supported on `map` fields, and it's already effectively optional since an empty map should be serialized to zero bytes

nym-vpnd/src/service/error.rs
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                    NymIdError::StorageError { source } => {
+                        // There was a recent change for the upstream crate that adds a new variant
+                        // to StorageError to capture duplicate entries. Until that change makes
+                        // its way to the vpn-lib, we just match on the string as a temporary
+                        // solution.
+                        if let Some(StorageError::InternalDatabaseError(db_error)) =
+                            source.downcast_ref::<StorageError>()
+                        {
+                            if db_error.to_string().contains("code: 2067") {
```
jstuczyn, line 62: checking for string existence is a bit iffy but I understand the context and the fact this code is stable and well defined in sql itself

PR #398, opened 2024-05-14
Add extra macOS linker flags

rokas-ambrazevicius:
Add support to load daemon via native macOS API, through native macOS app.

Versioning is done:
```
	<key>CFBundleShortVersionString</key>
	<string>0.1</string>
	<key>CFBundleVersion</key>
	<string>1</string>
```
https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion
https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,9 @@ build-wireguard:
 build-vpn-cli:
 	RUSTFLAGS="-L $(CURDIR)/build/lib/$(ARCH)" cargo build --release
 
+build-vpn-cli:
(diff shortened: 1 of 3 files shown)


Daemon/Info.plist
octol, line 1: Imho let's place these files in `nym-vpnd/mac`

octol (approved):

Daemon/Info.plist
rokas-ambrazevicius: Updated 👍

PR #408, opened 2024-05-16
Strongly typed connection failure in nym-vpnd

octol:
Send strongly typed connection fail errors from nym-vpnd. There are so many possible error cases so currently most of them are mapped to Unhandled. This is something we'll have to gradually iterate on, to at least map the common ones to specific enum cases

### DIFF
--- a/proto/nym/vpn.proto
+++ b/proto/nym/vpn.proto
@@ -86,7 +86,24 @@ message ConnectionStatusUpdate {
 }
 
 message Error {
-  string message = 1;
+  enum ErrorType {
+    ERROR_TYPE_UNSPECIFIED = 0;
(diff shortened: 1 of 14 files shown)


neacsu (approved):

proto/nym/vpn.proto
```diff
@@ -86,7 +86,39 @@ message ConnectionStatusUpdate {
 }
 
 message Error {
-  string message = 1;
+  enum ErrorType {
+    ERROR_TYPE_UNSPECIFIED = 0;
+
+    // An error that was not explicitly handled by the vpn service. This should
+    // not happend but it will while we interate on mapping out all possible
```
doums, line 93: typo

PR #437, opened 2024-05-23
refactor(vpnx): backend specialized error handling

doums:
Refactors the overall backend to frontend error handling. The purpose is adding localization and specialization to error messages in the UI.
Any call to the App backend will potentially return the following error type including an error `key` that the frontend can safely use to map to localized error messages:

```rust
/// Generic error type made to be passed to the frontend and
/// displayed in the UI as localized error message
/// `Bkd` stands for 'Backend', source side from where the error is emitted
pub struct BkdError {
    /// Human readable error message for debugging/logs purposes
    pub message: String,
    /// Error key to be used in the UI to display localized error message
    pub key: ErrorKey,
    /// Extra data to be passed along to help specialize the problem
    pub data: Option<HashMap<String, String>>,
}
```

**NOTE** `key` is a required field, that is any backend error is now "specialized"

For now, this is the keys supported and exposed by the app backend
```rust
pub enum ErrorKey {
    /// Generic unhandled error
    UnknownError,
    /// Any error that is not explicitly handled, and not related
    /// to the application layer
    /// Extra data should be passed along to help specialize the problem
    InternalError,
    /// gRPC bare layer error, when a RPC call fails (aka `Tonic::Status`)
    /// That is, the error does not come from the application layer
    GrpcError,
    /// Happens when the app is not connected to a running daemon
    /// and attempts to make a gRPC call
    NotConnectedToDaemon,
    /// Forwarded from proto
    ConnectionTimeout,
    /// Forwarded from proto
    ConnectionGatewayLookup,
    /// Forwarded from proto
    ConnectionNoValidCredential,
    /// Forwarded from proto
    CredentialInvalid,
    /// Forwarded from proto
    CredentialVpnRunning,
    /// Forwarded from proto
    CredentialAlreadyImported,
    /// Forwarded from proto
    CredentialStorageError,
    /// Forwarded from proto
    CredentialDeserializationFailure,
    /// Forwarded from proto
    CredentialExpired,
}
```

This new error system allows the frontend to improve its error reporting in the UI, eg.

```typescript
    await invoke('add_credential', { credential: credential.trim() })
      .then(() => {
        navigate(routes.root);
      })
      .catch((e: BkdError) => {
        console.log('backend error:', e);
        if (e.key === 'CredentialExpired' && e.data?.expiration) {
          // TODO the expiration date format passed from the backend is not ISO_8601
          // So we have to parse it manually
          setError(
            `${eT(e.key)}: ${dayjs(e.data.expiration, 'YYYY-MM-DD').format('YYYY-MM-DD')}`,
          );
        } else {
          setError(eT(e.key));
        }
      });
```

```typescript
  useEffect(() => {
    if (error?.key === 'ConnectionNoValidCredential') {
      navigate(routes.credential);
      dispatch({ type: 'reset-error' });
    }
  }, [error, dispatch, navigate]);
```

### DIFF
--- a/nym-vpn-x/src-tauri/src/commands/node_location.rs
+++ b/nym-vpn-x/src-tauri/src/commands/node_location.rs
@@ -6,7 +6,7 @@ use ts_rs::TS;
 use crate::{
     country::Country,
     db::{Db, Key},
-    error::BkdError,
+    error::BackendError,
     gateway::{get_gateway_countries, get_low_latency_entry_country},
     states::{app::NodeLocation, SharedAppState},
 };
@@ -24,7 +24,7 @@ pub async fn set_node_location(
     db: State<'_, Db>,
     node_type: NodeType,
     location: NodeLocation,
-) -> Result<(), BkdError> {
+) -> Result<(), BackendError> {
     debug!("set_node_location");
     let mut state = app_state.lock().await;
     match node_type {
@@ -41,11 +41,11 @@ pub async fn set_node_location(
     match node_type {
         NodeType::Entry => {
             db.insert(Key::EntryNodeLocation, &location)
-                .map_err(|_| BkdError::new_internal("Failed to save location in db", None))?;
+                .map_err(|_| BackendError::new_internal("Failed to save location in db", None))?;
         }
         NodeType::Exit => {
             db.insert(Key::ExitNodeLocation, &location)
-                .map_err(|_| BkdError::new_internal("Failed to save location in db", None))?;
+                .map_err(|_| BackendError::new_internal("Failed to save location in db", None))?;
         }
     }
 
@@ -54,11 +54,11 @@ pub async fn set_node_location(
 
 #[instrument]
 #[tauri::command]
-pub async fn get_fastest_node_location() -> Result<Country, BkdError> {
+pub async fn get_fastest_node_location() -> Result<Country, BackendError> {
     debug!("get_fastest_node_location");
     get_low_latency_entry_country().await.map_err(|e| {
         error!("failed to get fastest node location: {}", e);
-        BkdError::new_internal("failed to get fastest node location", None)
+        BackendError::new_internal("failed to get fastest node location", None)
     })
 }
 
@@ -67,7 +67,7 @@ pub async fn get_fastest_node_location() -> Result<Country, BkdError> {
 pub async fn get_node_location(
     app_state: State<'_, SharedAppState>,
     node_type: NodeType,
-) -> Result<NodeLocation, BkdError> {
+) -> Result<NodeLocation, BackendError> {
     debug!("get_node_location");
     Ok(match node_type {
         NodeType::Entry => app_state.lock().await.entry_node_location.clone(),
@@ -77,16 +77,16 @@ pub async fn get_node_location(
 
 #[instrument]
 #[tauri::command]
-pub async fn get_countries(node_type: NodeType) -> Result<Vec<Country>, BkdError> {
+pub async fn get_countries(node_type: NodeType) -> Result<Vec<Country>, BackendError> {
     debug!("get_countries");
     match node_type {
         NodeType::Entry => get_gateway_countries(false).await.map_err(|e| {
             error!("failed to get node locations: {}", e);
-            BkdError::new_internal("failed to get node locations", None)
+            BackendError::new_internal("failed to get node locations", None)
         }),
         NodeType::Exit => get_gateway_countries(true).await.map_err(|e| {
             error!("failed to get node locations: {}", e);
-            BkdError::new_internal("failed to get node locations", None)
+            BackendError::new_internal("failed to get node locations", None)
         }),
     }
 }
--- a/nym-vpn-x/src-tauri/src/commands/node_location.rs
+++ b/nym-vpn-x/src-tauri/src/commands/node_location.rs
@@ -6,7 +6,7 @@ use ts_rs::TS;
 use crate::{
     country::Country,
     db::{Db, Key},
-    error::{CmdError, CmdErrorSource},
+    error::BkdError,
     gateway::{get_gateway_countries, get_low_latency_entry_country},
     states::{app::NodeLocation, SharedAppState},
 };
@@ -24,7 +24,7 @@ pub async fn set_node_location(
     db: State<'_, Db>,
     node_type: NodeType,
     location: NodeLocation,
-) -> Result<(), CmdError> {
+) -> Result<(), BkdError> {
     debug!("set_node_location");
     let mut state = app_state.lock().await;
     match node_type {
@@ -40,20 +40,12 @@ pub async fn set_node_location(
     debug!("saving new location in db");
     match node_type {
         NodeType::Entry => {
-            db.insert(Key::EntryNodeLocation, &location).map_err(|_| {
-                CmdError::new(
-                    CmdErrorSource::InternalError,
-                    "Failed to save location in db",
-                )
-            })?;
+            db.insert(Key::EntryNodeLocation, &location)
+                .map_err(|_| BkdError::new_internal("Failed to save location in db", None))?;
         }
         NodeType::Exit => {
-            db.insert(Key::ExitNodeLocation, &location).map_err(|_| {
-                CmdError::new(
-                    CmdErrorSource::InternalError,
-                    "Failed to save location in db",
-                )
-            })?;
+            db.insert(Key::ExitNodeLocation, &location)
+                .map_err(|_| BkdError::new_internal("Failed to save location in db", None))?;
         }
     }
 
@@ -62,14 +54,11 @@ pub async fn set_node_location(
 
 #[instrument]
 #[tauri::command]
-pub async fn get_fastest_node_location() -> Result<Country, CmdError> {
+pub async fn get_fastest_node_location() -> Result<Country, BkdError> {
     debug!("get_fastest_node_location");
     get_low_latency_entry_country().await.map_err(|e| {
         error!("failed to get fastest node location: {}", e);
-        CmdError::new(
-            CmdErrorSource::InternalError,
-            "failed to get fastest node location",
-        )
+        BkdError::new_internal("failed to get fastest node location", None)
     })
 }
 
@@ -78,7 +67,7 @@ pub async fn get_fastest_node_location() -> Result<Country, CmdError> {
 pub async fn get_node_location(
     app_state: State<'_, SharedAppState>,
     node_type: NodeType,
-) -> Result<NodeLocation, CmdError> {
+) -> Result<NodeLocation, BkdError> {
     debug!("get_node_location");
     Ok(match node_type {
         NodeType::Entry => app_state.lock().await.entry_node_location.clone(),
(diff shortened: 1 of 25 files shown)


nym-vpn-x/src-tauri/src/error.rs
```diff
@@ -20,100 +24,160 @@ pub enum CmdErrorSource {
...
-#[derive(Error, Debug, Serialize, Deserialize, TS)]
+#[derive(Error, Debug, Serialize, Deserialize, TS, Clone)]
 #[ts(export)]
-pub struct CmdError {
-    #[source]
-    pub source: CmdErrorSource,
+/// Generic error type made to be passed to the frontend and
+/// displayed in the UI as localized error message
+/// `Bkd` stands for 'Backend', source side from where the error is emitted
+pub struct BkdError {
```
rokas-ambrazevicius, line 32: Wouldn't it be better to name it `BackendError`? You would save a comment + it would be easier to understand the code. Unless, I am unaware of some Rust codestyle...

nym-vpn-x/src/hooks/useI18nError.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Hook to get the translation function for backend errors
+ *
+ * @returns The translation function
+ */
+function useI18nError() {
+  const { t } = useTranslation('errors');
+
+  const translateError = useCallback(
```
rokas-ambrazevicius, line 13: 👍 Love the more user friendly errors with localizations

nym-vpn-x/src-tauri/src/commands/node_location.rs
```diff
@@ -88,22 +77,16 @@ pub async fn get_node_location(
...
-            )
+            BkdError::new_internal("failed to get node locations", None)
         }),
         NodeType::Exit => get_gateway_countries(true).await.map_err(|e| {
             error!("failed to get node locations: {}", e);
-            CmdError::new(
-                CmdErrorSource::InternalError,
-                "failed to get node locations",
-            )
+            BkdError::new_internal("failed to get node locations", None)
```
octol, line 89: I see multiple places where the underlying error is discarded. I think you'll need to include them, otherwise it's going to be hard for the user and us to understand what's going wrong and fix things

nym-vpn-x/src-tauri/src/error.rs
```diff
@@ -20,100 +24,160 @@ pub enum CmdErrorSource {
...
 
-/// Enum of the localization keys for error messages
-/// displayed in the UI
-#[derive(Debug, Serialize, Deserialize, TS)]
+/// Enum of the possible specialized errors emitted by the daemon
+/// or the app backend side, to be passed to the UI layer
+#[derive(Debug, Serialize, Deserialize, TS, Clone)]
 #[ts(export)]
-pub enum I18nKey {
+pub enum ErrorKey {
```
octol, line 111: It looks like you're merging 3 kinds of errors into one global one, is there a frontend reason for this? It looks to me like you want to keep them separate, especially the credential and connection errors since they can never happen in the same situations. Also it looks like this information is still weakly encoded in the prefix-name?

octol (commented):
General take: I see a lot of places where we are throwing away the error, that is going to make our lives a lot harder I think if we can't easily identify what's going wrong for users

nym-vpn-x/src-tauri/src/error.rs
doums: I wanted a short name because this type is used a lot in the codebase, so saving few characters length is a good take.
Please note that use of abbreviation in naming is a usual thing, a few examples that come to mind in rust:
`Vec` `Iter` `Ref` `str` `mpsc` `dbg!` `info` `warn` etc etc

octol: I think I agree with @rokas-ambrazevicius here, BackendError seems easier to read while still being relatively short. The types referenced there are not comparable, as they are foundational types that all Rust developers will be familiar with, unlike this type

doums:
> General take: I see a lot of places where we are throwing away the error, that is going to make our lives a lot harder I think if we can't easily identify what's going wrong for users

> I see multiple places where the underlying error is discarded. I think you'll need to include them, otherwise it's going to be hard for the user and us to understand what's going wrong and fix things

Special care has been taken to always logs errors. So the "technicals" details of errors always show up into logs/output, and should be recorded and reported by Sentry (or it is planned to be the case).
I don't think we want to display raw technicals' error messages in the UI anymore. End user should not be exposed to the details of tech errors. But rather generics error messages, and only specific error related to the user and its interaction with the app. tldr: the end user is not a dev.

nym-vpn-x/src-tauri/src/error.rs
doums: Yes it is, the whole purpose of this cross error handling is to expose to the UI layer a unique set of explicitly defined errors and which are the only possible errors that can happens and having impact at the UI layer.
Enforcing that allows the frontend to create whatever system it needs on top of it. Eg. basing the localization of error messages.
Exposing separated errors types to the frontend ui instead of one unique type, just add more complexity for no benefits.

> Also it looks like this information is still weakly encoded in the prefix-name?

I'm not sure why you said "weakly" but since a key in that enum can represent a very specific error case, the name has to be unique and self-descriptive enough (+ rust doc added details for each variants).

octol:
> > General take: I see a lot of places where we are throwing away the error, that is going to make our lives a lot harder I think if we can't easily identify what's going wrong for users
> 
> > I see multiple places where the underlying error is discarded. I think you'll need to include them, otherwise it's going to be hard for the user and us to understand what's going wrong and fix things
> 
> Special care has been taken to always logs errors. So the "technicals" details of errors always show up into logs/output, and should be recorded and reported by Sentry (or it is planned to be the case). I don't think we want to display raw technicals' error messages in the UI anymore. End user should not be exposed to the details of tech errors. But rather generics error messages, and only specific error related to the user and its interaction with the app. tldr: the end user is not a dev.

I think ideally what we want is not to have like before with only a technical error that is only confusing for the user, nor just a generic error message without any content. Ideally we present an error message that is at the right technical level to be understandable to the user + a "details" button or expand toggle that can show the underlying error if the users explicitly wants to.

Few people will in practice run Sentry or have the ability to dig through the logs. It will help tremendously when fixing issues if the user is able to relay back to devs what went wrong

nym-vpn-x/src-tauri/src/commands/node_location.rs
doums: This is voluntary. If db write throws error for some reason, we don't want to report the error tech details to the user. He's just not the right target for it. But the error will be "friendly" reported as "Oups an internal error occured" in the UI. And the tech details of the error will be:
- logged in output/logs (so as dev we should have mats to debug)
- reported by sentry (if turned on, feature planned)

nym-vpn-x/src-tauri/src/error.rs
octol: By "weakly" here I mean weak typing, since it looks like in the error enum that you are merging to disjoint error types, but still separating by prefixing them in the name with "Connection" and "Credential"

> Yes it is, the whole purpose of this cross error handling is to expose to the UI layer a unique set of explicitly defined errors and which are the only possible errors that can happens and having impact at the UI layer.

So that's the thing, by merging them means that you can't look at the error enum and tell exactly which errors that can happen (e.g. you can't have credential import errors when connecting). I don't know how your are consuming these, but I advise you to double check this to confirm you really want a global error like this

nym-vpn-x/src-tauri/src/commands/node_location.rs
octol: See my other comment why I disagree with this. Realistically speaking not including them will mean that for almost all bug reports from users we wont be able to do anything to fix them

doums:
> a "details" button or expand toggle that can show the underlying error if the users explicitly wants to.

This not in the scope of the present PR (neither a **planned feature**). Ofc we can raise the point and check it with the team for the future.

Okay I think the PR review started to stray from the "code review" purpose. Let's discuss these points internally please.
It's not the right place to discuss feature plan in PR's review.

nym-vpn-x/src-tauri/src/error.rs
doums: Okay, since at the end this is a matter of personal preferences, I'd like to know @zaneschepke opinion
If you're all afraid of abbreviation naming 😨 I don't mind refactoring `Bkd` to `Backend` ofc

nym-vpn-x/src-tauri/src/error.rs
doums: Let's discuss it internally please

nym-vpn-x/src-tauri/src/error.rs
rokas-ambrazevicius: My suggestion is based on Swift naming conventions. 
Since Swift and Rust are both object oriented languages, so I think naming conventions between object oriented languages should be quite similar.
In Swift - naming things with long meaningful names, saves a lot of time later on.

zaneschepke:
> > > General take: I see a lot of places where we are throwing away the error, that is going to make our lives a lot harder I think if we can't easily identify what's going wrong for users
> > 
> > 
> > > I see multiple places where the underlying error is discarded. I think you'll need to include them, otherwise it's going to be hard for the user and us to understand what's going wrong and fix things
> > 
> > 
> > Special care has been taken to always logs errors. So the "technicals" details of errors always show up into logs/output, and should be recorded and reported by Sentry (or it is planned to be the case). I don't think we want to display raw technicals' error messages in the UI anymore. End user should not be exposed to the details of tech errors. But rather generics error messages, and only specific error related to the user and its interaction with the app. tldr: the end user is not a dev.
> 
> I think ideally what we want is not to have like before with only a technical error that is only confusing for the user, nor just a generic error message without any content. Ideally we present an error message that is at the right technical level to be understandable to the user + a "details" button or expand toggle that can show the underlying error if the users explicitly wants to.
> 
> Few people will in practice run Sentry or have the ability to dig through the logs. It will help tremendously when fixing issues if the user is able to relay back to devs what went wrong

This sounds like a good idea, but we would need to sync with design team as it would required some UI design changes.

zaneschepke (approved):

octol:
> > > > General take: I see a lot of places where we are throwing away the error, that is going to make our lives a lot harder I think if we can't easily identify what's going wrong for users
> > > 
> > > 
> > > > I see multiple places where the underlying error is discarded. I think you'll need to include them, otherwise it's going to be hard for the user and us to understand what's going wrong and fix things
> > > 
> > > 
> > > Special care has been taken to always logs errors. So the "technicals" details of errors always show up into logs/output, and should be recorded and reported by Sentry (or it is planned to be the case). I don't think we want to display raw technicals' error messages in the UI anymore. End user should not be exposed to the details of tech errors. But rather generics error messages, and only specific error related to the user and its interaction with the app. tldr: the end user is not a dev.
> > 
> > 
> > I think ideally what we want is not to have like before with only a technical error that is only confusing for the user, nor just a generic error message without any content. Ideally we present an error message that is at the right technical level to be understandable to the user + a "details" button or expand toggle that can show the underlying error if the users explicitly wants to.
> > Few people will in practice run Sentry or have the ability to dig through the logs. It will help tremendously when fixing issues if the user is able to relay back to devs what went wrong
> 
> This sounds like a good idea, but we would need to sync with design team as it would required some UI design changes.

Yep that sounds like a plan to me. As long as we view this PR as a stepping stone towards that

PR #446, opened 2024-05-28
refactor(vpnx): fetch countries data logic

doums:
- fetch the available countries from the gateway API (api/directory/gateways/{entry|exit}/countries) ⚡⚡
- add logic to check validity of current selected countries (if not in available list, pick a random one)
- display localized and specialized errors in the UI when countries data fetching is failing
- swap default (hardcoded) countries: entry `CH`, exit `FR`


octol (commented):
1. Doesn't this break running on non-mainnet? Imho either fix that or remove the ability to run on sandbox etc altogether
2. I would have preferred if we reused the already existing common crate for interacting with the nymvpn.com API. That would have been an excellent opportunity to add ISO lookup support there since it's something we'll need in other places as well. That crate also supports getting the list of gateways, something you likely will want to do at some point

doums:
@octol for point `2` I'm fine to use the existing crate to pull the countries list (2 http requests). But as you said it seems to require a bit of work on that crate first to add iso lookup. Which is out of the scope here. But it's a good idea for a following PR.

octol:
> Good catch for point `1`. If we are OK to drop custom (sandbox etc) network support, I can just drop it?

Been thinking a bit more about this, and I think sandbox support is less important in the new client. The daemon supports custom environments anyway so when testing we can always just resort to the vpnc client. So I would just drop sandbox for now

octol (approved):

PR #450, opened 2024-05-29
feat(legacy, vpnx): reset config file on read or parse failure

doums:
- at app startup ignore any read or parsing failures of the config file
- on error, try to clear config content and fallback to default config
- do not exit the app on config read failure
- swap default vpn mode to 5hop

### DIFF
--- a/nym-vpn-desktop/src-tauri/src/main.rs
+++ b/nym-vpn-desktop/src-tauri/src/main.rs
@@ -7,7 +7,7 @@ use anyhow::{anyhow, Result};
 use clap::Parser;
 use tauri::{api::path::config_dir, Manager};
 use tokio::sync::Mutex;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info, trace, warn};
 
 use commands::db as cmd_db;
 use commands::window as cmd_window;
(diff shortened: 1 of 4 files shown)


nym-vpn-desktop/src-tauri/src/main.rs
```diff
@@ -93,7 +93,19 @@ async fn main() -> Result<()> {
...
+        Ok(cfg) => cfg,
+        Err(e) => {
+            warn!("failed to read app config: {e}, falling back to default (empty) config");
+            debug!("clearing the config file");
+            app_config_store
+                .clear()
+                .await
+                .inspect_err(|e| error!("failed to clear the config file: {e}"))
+                .ok();
+            AppConfig::default()
```
octol, line 106: I think this is good in that it guards against the case when the file for whatever reason is corrupt or not readable, and we can then reset it.

But for the future, it doesn't handle e.g when the config file is readonly due to for example network share issues. 

Basically the correct and user friendly way to handle this whole section of code is to remove all return or error (`?`) that we can't recover from, instead pass them to the gui constructor further down and display it once the UI is up and running. We shouldn't assume that the user is launching this from a terminal

octol (approved):

PR #452, opened 2024-05-29
fix(vpnx): use connection time provided by daemon

doums:
- use the reported "established connection time" as reported by daemon
- when it's not reported as on `Connected` event, fallback to utc current time
- for any reason if the app lost the vpn connection state, on next state refresh, if vpn still connected, the connection timer will be set as expected


octol (approved):
I've been thinking about replacing the unix timestamp with something higher resolution so we get subsecond resolution. But that will be later

doums:
Actually the `Timestamp` from proto includes the nanos resolution part. But precision in seconds for the connection timer is more than sufficient imo

PR #458, opened 2024-05-31
allow probing blacklisted gateways

jstuczyn:
Before this PR can be fully integrated, we require a change inside nym-sdk to allow for injecting gateways for during initial gateway selection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/458)


octol:
Closing since it's both out of date and likely not relevant anymore

PR #472, opened 2024-06-10
add basic android ci

zaneschepke:


.github/workflows/ci-nym-vpn-android.yml
```diff
@@ -0,0 +1,32 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "nym-vpn-android/**"
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
doums, line 12: just a question, what about using our custom runners for faster runs?

doums (approved):

PR #482, opened 2024-06-11
chore(vpnx): watch vpn status

doums:
Add rpc call to `ListenToConnectionStatus`. 
In the future, the data received could be passed to the UI layer to display addtional details on current connection states.
(For now it's logged only)

### DIFF
--- a/nym-vpn-x/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-x/src-tauri/src/grpc/client.rs
@@ -17,7 +17,7 @@ use time::OffsetDateTime;
 use tokio::sync::mpsc;
 use tonic::transport::Endpoint as TonicEndpoint;
 use tonic::{transport::Channel, Request};
-use tracing::{debug, error, info, instrument, warn};
+use tracing::{debug, error, info, instrument, trace, warn};
 use ts_rs::TS;
 
 use crate::cli::Cli;
@@ -175,9 +175,9 @@ impl GrpcClient {
         Ok(())
     }
 
-    /// Watch VPN status updates
+    /// Watch VPN state updates
     #[instrument(skip_all)]
-    pub async fn watch_vpn_status(&self, app: &AppHandle) -> Result<()> {
+    pub async fn watch_vpn_state(&self, app: &AppHandle) -> Result<()> {
         let mut vpnd = self.vpnd().await?;
 
         let request = Request::new(Empty {});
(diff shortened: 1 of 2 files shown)


nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -224,6 +224,53 @@ impl GrpcClient {
...
+        let mut stream = vpnd
+            .listen_to_connection_status(request)
+            .await
+            .inspect_err(|e| {
+                error!("listen_to_connection_status failed: {}", e);
+            })?
+            .into_inner();
+
+        let (tx, mut rx) = mpsc::channel(32);
+        tokio::spawn(async move {
```
octol, line 242: It looks like you are anyway running an event loop below, so I'm not sure you need to spawn this task. Can't you just look over stream directly below where you are currently looking over rx?

doums: I spawn the task to not block current thread during watching the stream, as it needs to be iterated in a loop. Also, I need to use a channel because I'll need to use (borrow/move) the `AppHandle` once we'll handle status update (same logic as in `watch_vpn_state` function above).
So unless I missed something, no other choice than having to use a channel.

zaneschepke (approved):

nym-vpn-x/src-tauri/src/grpc/client.rs
octol: But it looks like the function is blocking on the while loop?

octol: Just try deleting the whole tokio::spawn thingie and replace 
```
while let Some(status) = rx.recv().await {
```
with
```
while let Ok(Some(status)) = stream.message().await {
```

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -224,6 +224,53 @@ impl GrpcClient {
...
+            // TODO handle status updates
+            debug!(
+                "vpn status update {:?}, {:?}",
+                status.kind(),
+                status.message
+            );
+            if !status.details.is_empty() {
+                trace!("vpn status details: {:?}", status.details);
+            }
+        }
```
octol, line 269:
```suggestion
 
        while let Ok(Some(status)) = stream.message().await {
            // TODO handle status updates
            debug!(
                "vpn status update {:?}, {:?}",
                status.kind(),
                status.message
            );
            if !status.details.is_empty() {
                trace!("vpn status details: {:?}", status.details);
            }
        }
```

doums: Thanks for the suggestion. The thing is using ` while let Ok(Some(…))` we can't handle the other possible variants i.e. `Ok(None)` or `Err(e)`, right?
Also, an `Err(e)` could be emitted one time by the server, but that does not mean the stream has finished and closed. But using the suggested `while let` pattern it will be terminated after the first non Ok(Some(..)) encoutered. Which is quite "fragile".
Even if we are just logging (for now) when these other variants kick in, I prefer to keep that for better reliability and better control on code flow.

doums: I will merge it as is, but I'm open to continue the discussion internally if you think it's needed.

octol: If you want to handle the other states you can e.g replace the while loop with `loop { select { .. } }`. The main point imho is that there is no need to spawn a separate task, since this function will anyway block while listening for messages

PR #510, opened 2024-06-19
pkg(vpnx, vpnd): add linux installer script

doums:
- all-in-one "universal" Linux installer
- install nymvpn-x  as AppImage (downloaded from GH release)
- install nym-vpnd  from prebuilt binary (downloaded from GH release), thx to the way Rust compiles, it should work on most systems
- checks sums for all downloaded artifacts (using sums provided by the releases)
- user can pick the installation directory: `/usr` or `HOME/local`
- check if any existing installation is found and stop or any files conflicts would cancel the script
- all steps are logged
- script is self-contained (yes bash still has its strength xD)
- add a workflow to public release new version of the script (on-demand for now)

### DIFF
--- a/.github/workflows/release-linux-installer.yml
+++ b/.github/workflows/release-linux-installer.yml
@@ -1,0 +1,60 @@
+# Create a new release of the Linux installer script
+
+name: release-linux-installer
+on:
+  workflow_dispatch:
+    inputs:
+      client_tag:
+        description: "Tag name of the client release"
+        required: true
+        default: nym-vpn-x-v0.1.0
+      client_version:
+        description: "Version of the client"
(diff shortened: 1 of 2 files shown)


octol (commented):
Note of caution here 

You need to be very careful that this is compatible with the package installations (deb etc) otherwise this can end up completely borking peoples machines with inconsistent systemd configurations. Especially if this does not have uninstall functionality

A casual user will not be able resolve any issues caused manually, short of reinstalling their OS

doums:
I'll add a check to scan PATH for vpnd or vpnx presences, if yes we would end the installation, inviting the user to uninstall first any existing installation.

PR #517, opened 2024-06-21
feat(vpnx): add credential expiry date  import

zaneschepke:


doums:
#### Important note:

This is a partial and incomplete implementation of the "Credential expiry" feature as it only saves the expiry date on credential import. That means it does not handle the following possible cases:
- when user successfully import a valid credential, daemon is storing that cred on its side and the cred date passed to the client is saved in the **client db**
- user for any reason uninstall the client app (only client, not daemon) -> the saved expiry date is lost, but daemon is still using the imported credential
- user install back the client app -> now the app has no way to know that a valid cred has already been imported including plenty of remaining time -> for client app no valid cred is present -> client app and daemon are unsync now on that "state"
- in the UI user import the same cred again
-> result of an error from daemon because that cred is already imported

Same problematic case would apply if user remove/reinstall only the daemon (including purging daemon local files)

nym-vpn-x/src-tauri/Cargo.toml
```diff
@@ -43,7 +43,7 @@ nym-gateway-directory = { path = "../../nym-vpn-core/crates/nym-gateway-director
 nym-config = { git = "https://github.com/nymtech/nym", rev = "b8b66fa" }
 
 [target."cfg(windows)".dependencies]
-windows = { version = "0.56.0", features = ["Win32_System_Console", "Win32_Foundation"] }
+windows = { version = "0.57.0", features = ["Win32_System_Console", "Win32_Foundation"] }
```
octol, line 46: if you bump this, please make sure to check that it works on windows

doums (approved):

PR #548, opened 2024-06-26
feat(linux-install): cleanup local app files

doums:
add local files cleanup to the installer script (on uninstall)

### DIFF
--- a/.pkg/installer.sh
+++ b/.pkg/installer.sh
@@ -26,13 +26,13 @@ B_GRY="$BLD$GRY"
 ####
 
 # nymvpn-x AppImage
-vpnx_tag=nightly-x
-vpnx_version=0.1.2-dev
+vpnx_tag=nym-vpn-x-v0.1.2
+vpnx_version=0.1.2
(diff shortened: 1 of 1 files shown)


.pkg/installer.sh
```diff
@@ -101,8 +101,17 @@ rmfile() {
   fi
 }
 
+_rmdir() {
+  if [ -d "$1" ]; then
+    sudo rm -rf "$1"
```
rokas-ambrazevicius, line 106: do we need sudo here?

doums: "normally" not for the files we currently call this function for
But in the future we could need sudo for certain files
Anyway, uninstall script need 'sudo', and at that point of script execution, sudo already has been granted by the user

rokas-ambrazevicius: If we don't need it now, let's not add it now? Let's not try to guess what the future needs, let's add it when it's needed?

doums: thx for the suggestion but I prefer to keep that function "sudo" ready

zaneschepke (approved):

PR #553, opened 2024-06-27
Upgrade whole bunch of crates in nym-vpn-core

octol:
Run cargo update on the nym-vpn-core workspace, then manually upgrade a whole bunch of crates that doesn't seem to break compilation.

- Move tonic deps to workspace
- cargo update in nym-vpn-core
- Upgrade itertools to 0.13
- Upgrade netdev to 0.29
- Upgrade pnet_packet to 0.35
- Upgrade signature to 2.2
- Upgrade tun2 to 2.0.1


nym-vpn-core/crates/nym-gateway-directory/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license.workspace = true
 
 [dependencies]
 async-trait.workspace = true
-chrono = "0.4.37"
+chrono = "0.4.38"
```
jstuczyn, line 9: why are we even using `chrono` rather than `time` like we do everywhere else?

jstuczyn (approved):

nym-vpn-core/crates/nym-gateway-directory/Cargo.toml
octol: Good question, I have no good answer. Looking at the code it might have been introduced when lifting code from the main repo or the harbourmaster repo... yes let's add that to the todo list for changing

PR #565, opened 2024-06-27
upgrade to latest stable protoc

zaneschepke:
Bump for arduino/setup-protoc to v3

Remove deb package install of protoc as it is outdated and does not support our grpc optional for expiry/credential import API.

Also, it is best for us to use the same protoc install for all workflows and platforms.

### DIFF
--- a/.github/workflows/build-nym-vpn-desktop-linux.yml
+++ b/.github/workflows/build-nym-vpn-desktop-linux.yml
@@ -42,7 +42,12 @@ jobs:
 
       - name: Install system dependencies
         run: |
-          sudo apt-get update && sudo apt-get install -y libdbus-1-dev libmnl-dev libnftnl-dev protobuf-compiler libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 librsvg2-dev file
+          sudo apt-get update && sudo apt-get install -y libdbus-1-dev libmnl-dev libnftnl-dev libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 librsvg2-dev file
+
+      - name: Install protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install rust toolchain
         uses: brndnmtthws/rust-action-rustup@v1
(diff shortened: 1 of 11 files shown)


doums (approved):

.github/workflows/build-nym-vpn-desktop-linux.yml
```diff
@@ -42,7 +42,12 @@ jobs:
 
       - name: Install system dependencies
         run: |
-          sudo apt-get update && sudo apt-get install -y libdbus-1-dev libmnl-dev libnftnl-dev protobuf-compiler libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 librsvg2-dev file
+          sudo apt-get update && sudo apt-get install -y libdbus-1-dev libmnl-dev libnftnl-dev libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 librsvg2-dev file
+
+      - name: Install protoc
```
octol, line 47: So I'm not convinced about this, the implications are that we can't anymore build on plain Ubuntu install. What are the reasons for upgrading?

octol (changes requested):

.github/workflows/build-nym-vpn-desktop-linux.yml
zaneschepke: We need to upgrade to support grpc optionals for credential import. The deb package is outdated. Additionally, I think it is better and easier to maintain if we are using the same install/setup and version on all of our workflows.

doums: It's logic to rely on the latest stable protoc version…

Not logic to rely on whatever Ubuntu is using as protoc version from its system package (which is outdated on runners)…

…to build our inter process protocol communication!

doums: If whatever package system you use is outdated, it is very straight to switch on latest stable simply by downloading it from official releases -> https://github.com/protocolbuffers/protobuf/releases/tag/v27.2

(a matter of putting `protoc`  in the PATH)

octol: Discussed extensively elsewhere, but adding here for completeness. I would prefer if we keep system `protobuf-compiler` on ubuntu.

However, a possible way forward here is if add `ubuntu-24.04` to `ci-nym-vpn-core.yml` with system protoc, and switch to manually installed protoc on ubuntu-22.04. Then we can start using `optional` while still keeping keeping the integrity of the system installed toolchain.

PR #572, opened 2024-06-28
Support for device specific keys

octol:
This PR adds support for a new data store that is intended to be used in the vpn client. More precisely:

- Create new crate `nym-vpn-store` that supports storing device specific keys.
- Create the keys as part of connecting with nym-vpnd.

The device keys are similar but orthogonal to the existing keys used by the mixnet client.

Things not covered that we'll need to add later:
- command to vpnd to query the keypair


nym-vpn-core/crates/nym-vpn-store/src/helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+pub fn keypair_exists<P: AsRef<Path>>(path: P) -> Result<bool, KeyStoreError> {
+    let device_key_paths = DeviceKeysPaths::new(path);
+    let private_key_exists = device_key_paths.private_device_key().exists();
+    let public_key_exists = device_key_paths.public_device_key().exists();
+    match (private_key_exists, public_key_exists) {
+        (true, true) => Ok(true),
+        (false, false) => Ok(false),
+        (true, false) => Err(KeyStoreError::InvalidKeyPair {
```
jstuczyn, line 38: that imo shouldn't return an error. sure, we could log a warning here, but we could just load the private key and trivially derive the public key out of it. think of ssh keys. you don't really store public keys at all. you just persist the private one and generate public on demand

nym-vpn-core/crates/nym-vpn-store/src/helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+        .await
+        .map_err(|error| KeyStoreError::Create {
+            path: path.as_ref().to_path_buf(),
+            error,
+        })
+}
+
+pub async fn create_device_keys_from_seed<P: AsRef<Path> + Clone>(
+    path: P,
+    seed: [u8; 32],
```
jstuczyn, line 78: just a sanity check: bad seed would only ever hurt the user, right? also, how do we generate the seed?

nym-vpn-core/crates/nym-vpn-store/src/key_store.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+
+use std::error::Error;
+
+use crate::DeviceKeys;
+
+pub trait KeyStore {
+    type StorageError: Error;
+
+    #[allow(async_fn_in_trait)]
```
jstuczyn, line 11: to remove this (assuming we don't export the trait), we could reduce the visibility of the trait to `pub(crate) trait KeyStore`. 

also, it's nice to not have to use `#[async_trait]` everywhere isn't it? : )

nym-vpn-core/crates/nym-vpn-store/src/persistence/on_disk.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn load_device_keypair(&self) -> Result<ed25519::KeyPair, OnDiskKeysError> {
+        let device_paths = self.paths.device_key_pair_path();
+        self.load_keypair(device_paths, "device")
+    }
+
+    fn load_keypair<T: PemStorableKeyPair>(
+        &self,
+        paths: KeyPairPath,
+        _name: impl Into<String>,
```
jstuczyn, line 67: rather than ignoring the name field, maybe we could use it to log an error message via `inspect_err`? surely it'd help in debugging, no?

nym-vpn-core/nym-vpnd/src/service/config.rs
```diff
@@ -122,3 +125,15 @@ pub(super) fn create_data_dir(data_dir: &PathBuf) -> Result<(), ConfigSetupError
...
+pub(super) async fn create_device_keys(
+    data_dir: &Path,
+) -> Result<(), nym_vpn_store::KeyStoreError> {
+    // Check if the device keys already exists, if not then create them
+    if nym_vpn_store::keypair_exists(data_dir)? {
+        nym_vpn_store::create_device_keys(data_dir).await?;
+    }
+
+    // Check that we can successfully load them
+    nym_vpn_store::load_device_keys(data_dir).await.map(|_| ())
```
jstuczyn, line 138: I assume this is a placeholder?

jstuczyn (approved):

nym-vpn-core/crates/nym-vpn-store/src/helpers.rs
octol: Afaik the thinking is that we want to be able to use a stable device specific seed.

nym-vpn-core/crates/nym-vpn-store/src/key_store.rs
octol: Unfortunately I think we'll end up using this outside this crate, probably nym-vpn-lib :(

nym-vpn-core/crates/nym-vpn-store/src/helpers.rs
octol: Good point, I'll change it

nym-vpn-core/nym-vpnd/src/service/config.rs
octol: Basically, yep

PR #618, opened 2024-07-03
fix(vpnx): credential error key

doums:


nym-vpn-core/nym-vpnd/src/service/error.rs
```diff
@@ -56,12 +56,10 @@ impl From<VpnLibImportCredentialError> for ImportCredentialError {
                         // to StorageError to capture duplicate entries. Until that change makes
```
octol, line 56: Remove the comment now that it's not relevant anymore

octol (approved):

PR #633, opened 2024-07-04
Feature/reg over mix

neacsu:


nym-vpn-core/nym-vpn-lib/src/lib.rs
```diff
@@ -135,7 +121,24 @@ impl From<NymVpn<MixnetVpn>> for SpecificVpn {
...
+#[derive(Clone, Debug)]
+pub struct MixnetClientConfig {
+    /// Path to the data directory of a previously initialised mixnet client, where the keys reside.
+    pub mixnet_data_path: Option<PathBuf>,
+
+    /// Enable Poission process rate limiting of outbound traffic.
+    pub enable_poisson_rate: bool,
+
+    /// Disable constant rate background loop cover traffic
+    pub disable_background_cover_traffic: bool,
```
rokas-ambrazevicius, line 133: `is_background_cover_traffic_enabled`
`is_poisson_rate_enabled`
`is_credendials_mode_enabled`

If this was Swift, thats how I would name things. If that's not how we do things in Rust, ignore my comment. Also, it is always a good idea to keep the consistence between enabled/disabled. If we started with enable, I'd stick with it. Do we have something in codestyle regarding variable naming conventions?

octol: Yep in general I agree naming booleans with `is_` prefix is the way to go.

For these I'm not sure. I think those names comes from being used as CLI arguments at some point?
They convey that the option is changed from their current default. 

For example background cover traffic is enabled by default, and the boolean flag here should be `false` by default. So using the above convention would mean something like `is_background_cover_traffic_disabled`. I mean it works, but like git commits using active rather than passive language just fits better for options presented to the user.

`curl --disable-foo`
`curl --is_foo_disabled`

rokas-ambrazevicius: I would recommend sticking to the positive vibe `is_background_cover_traffic_enabled`. Mixing enabled/disabled, usually makes you think twice when debugging.

Maybe we need to start a codestyle guide, for the future?

neacsu: I agree that there are some cli options who have hard to grasp naming, and a code style guide would be nice, together with a sweep over our CLIs to apply it. I won't be making that change as part of this PR though, as here it's just a matter of moving that code around

nym-vpn-core/crates/nym-authenticator-client/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(request_id)
+    }
+
+    async fn listen_for_connect_response(&self, request_id: u64) -> Result<AuthenticatorResponse> {
+        // Connecting is basically synchronous from the perspective of the mixnet client, so it's safe
+        // to just grab ahold of the mutex and keep it until we get the response.
+        let mut mixnet_client_handle = self.mixnet_client.lock().await;
+        let mixnet_client = mixnet_client_handle.as_mut().unwrap();
+
+        let timeout = tokio::time::sleep(Duration::from_secs(5));
```
durch, line 128: Do we want to make this configurable? Something to keep in mind for the future

nym-vpn-core/crates/nym-authenticator-client/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+fn check_auth_message_version(message: &ReconstructedMessage) -> Result<()> {
```
durch, line 166: I'd rather have this a proper struct (the whole message), but we can leave it for a future iteration

nym-vpn-core/crates/nym-gateway-directory/src/entries/auth_addresses.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "entry: {:?} exit: {:?}",
+            self.entry_addr.0, self.exit_addr.0
+        )
+    }
+}
+
+pub fn extract_authenticator(
```
durch, line 46: This is great, love it

durch (approved):
Two minor maintainability comments for sometimes, otherwise looks good

PR #660, opened 2024-07-10
change to uniffi callback tun statuses

zaneschepke:


nym-vpn-core/nym-vpn-lib/src/tunnel_setup.rs
```diff
@@ -363,6 +364,8 @@ pub async fn setup_tunnel(nym_vpn: &mut SpecificVpn) -> Result<AllTunnelsSetup>
     info!("Setting up route manager");
     let route_manager = setup_route_manager().await?;
 
+    crate::platform::set_listener_status(TunStatus::EstablishingConnection);
```
doums, line 367: suggestion: for better code readability

```
use crate::platform;

// nice to keep module prefix for "imported" functions
platform::set_listener_status(TunStatus::EstablishingConnection);
```

doums (approved):

rokas-ambrazevicius (approved):

nym-vpn-core/nym-vpn-lib/src/platform/mod.rs
```diff
@@ -130,6 +145,7 @@ pub struct VPNConfig {
     #[cfg(target_os = "ios")]
     pub tun_provider: Arc<dyn crate::OSTunProvider>,
     pub credential_data_path: Option<PathBuf>,
+    pub tun_status_listener: Option<Arc<dyn TunnelStatusListener>>,
```
neacsu, line 148: If I understand correctly, this is `Option` type here just so that we can have it implemented on both kotlin and swift, so to not put pressure on syncing the two sides. I think once we have the implementation stabilised on both platforms, we should probably change this to just `Arc<dyn TunnelStatusListener>` and set the value inside `LISTENER` to `Some(config.tun_status_listener)` in the `runVPN` function (a change either @zaneschepke  or @rokas-ambrazevicius  could do, whichever does the last implementation). Or is there a front-end reason for this being optional?

neacsu (approved):

nym-vpn-core/nym-vpn-lib/src/platform/mod.rs
rokas-ambrazevicius: We get tunnel updates from the network extension, so we don't need this.

zaneschepke: I made this optional so Rokas can implement if he wants, but he does not have to because he does not need it right now. 

I made it cross platform capable so he could potentially implement this in the future if we add some new features that would require a callback (like live tunnel statistics, etc). We could then just extend this same listener with another callback function to serve these additional use cases for both platforms.

PR #749, opened 2024-07-24
feat(vpnx): add optional log file support for debugging

doums:
Feature to be integrated in the UI and to allow the user manually sharing his logs for debugging purpose (bug report and so on)

- allow writing client app logs into a local file by setting the env var `LOG_FILE=1`
- logs are written in `/home/_user_/.cache/nymvpn-x/log/app.log` (Linux) `C:\Users\_username_\AppData\Local\nymvpn-x\log\app.log` (Windows)
- basic logs rotation (at app launch, the previous logs session are saved as `app.old.log`) so that the latest (or current) session and the previous session is always saved
- packages use `LOG_FILE=1` (_TODO_ deb), so when the app is launched via either desktop entry (Linux) or any desktop shortcut (via Windows) /desktop menus app logs are saved
- In app settings the "Logs" menu item opens a file dialog that lets the user access the log files (copy them and so on), when user selects one file its path is copied to clipboard + an in-app notification is sent
- the log directory is automatically created at runtime if it does not exist yet

image

### DIFF
--- a/.pkg/linux/install
+++ b/.pkg/linux/install
@@ -153,7 +153,7 @@ wrapper="#! /bin/bash
 # https://github.com/nymtech/nym-vpn-client/issues/305
 export WEBKIT_DISABLE_DMABUF_RENDERER=1
 
-RUST_LOG=info,nym_vpn_x=debug INSTALL_DIR/$target_appimage"
+LOG_FILE=1 RUST_LOG=info,nym_vpn_x=trace INSTALL_DIR/$target_appimage"
 ###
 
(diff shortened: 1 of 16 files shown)


octol:
Note that the daemon already has some capability to write to log file https://github.com/nymtech/nym-vpn-client/blob/ac3934bdecba6b0751d94d82d06b2b5438dd44fb/nym-vpn-core/nym-vpnd/src/logging.rs#L23

zaneschepke (approved):

PR #784, opened 2024-07-29
Lookup gateways using new nymvpn.com api in nym-vpn-lib

octol:
This PR primary purpose is to replace calls to nym-api + explorer-api + harbourmaster with calls to nymvpn.com, which should greatly simplify things. This initial step makes calls to nymvpn.com and then annotates them with data from the nym-api, until the IPR and authenticator addresses are added to the nymvpn.com endpoint responses. Nonetheless, this allows us to drop the explicit dependency on explore-api and harbourmaster.

After this there are now a lot of code that is only used in nym-gateway-probe, which after we also port over to the new types introduced in this PR we can remove.

- Use nymvpn.com API for fetching entry and exit gateways
- Remove requests to harbourmaster
- Remove requests to explorer-api

### DIFF
--- a/nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
+++ b/nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
@@ -497,6 +497,7 @@ impl GatewayClient {
 
     pub async fn lookup_entry_gateways(&self) -> Result<GatewayList> {
         let entry_gateways = if let Some(nym_vpn_api_client) = &self.nym_vpn_api_client {
+            info!("Fetching entry gateways from nym-vpn-api...");
             let entry_gateways = nym_vpn_api_client.get_entry_gateways().await.unwrap();
             let mut entry_gateways: Vec<_> =
                 entry_gateways.into_iter().map(Gateway::from).collect();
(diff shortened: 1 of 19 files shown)


nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
```diff
@@ -107,7 +116,7 @@ impl Config {
 
         // Since harbourmatser isn't part of the standard nym network details, we need to handle it
         // as a special case.
-        let harbour_master_url = if network.network_name == "mainnet" {
+        let harbour_master_url = if network.network_name == defaults::mainnet::NETWORK_NAME {
```
tommyv1987, line 119: if we can do the same for harbourmaster that would be great


        let harbour_master_url = if network.network_name == defaults::mainnet::NETWORK_NAME {
            Some(
                nym_vpn_api_client::MAINNET_NYM_HARBOURMASTER_URL
                    .parse()
                    .expect("mainnet default nym-harbourmaster-url not parseable"),
            )
        } else {
            std::env::var("MAINNET_NYM_HARBOURMASTER_URL").ok().map(|url| {
                url.parse()
                    .expect("MAINNET_NYM_HARBOURMASTER_URL env variable not a valid URL")
            })
        };

octol: All of this will be deleted in the follow up PR as remove any reference to harbourmaster from the vpn client

neacsu (approved):

PR #785, opened 2024-07-29
fix(vpnx-rust): licenses data generation

doums:
fix rust build pipeline

### DIFF
--- a/nym-vpn-x/src-tauri/about.toml
+++ b/nym-vpn-x/src-tauri/about.toml
@@ -3,6 +3,7 @@ accepted = [
     "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
+    "BSL-1.0",
     "CC0-1.0",
     "GPL-3.0",
     "ISC",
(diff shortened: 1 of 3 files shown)


nym-vpn-x/src-tauri/about.toml
```diff
@@ -3,6 +3,7 @@ accepted = [
     "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
+    "BSL-1.0",
```
octol, line 6: Is this https://en.wikipedia.org/wiki/Business_Source_License ? Let me quickly look into it

octol (changes requested):

zaneschepke (approved):

nym-vpn-x/src-tauri/about.toml
octol: What code is using? As far as I can tell it's not compatible with GPL-3.0?

doums: @octol I think it's https://opensource.org/license/bsl-1-0

octol: Ah good

octol (approved):

PR #818, opened 2024-08-01
chore(linux-installer): bump vpnd to 0.1.11

doums:


zaneschepke (approved):

.pkg/linux/install
```diff
@@ -39,8 +39,8 @@ vpnx_version=0.1.8
 appimage_url="https://github.com/nymtech/nym-vpn-client/releases/download/$vpnx_tag/nymvpn-x_${vpnx_version}.AppImage"
 
 # nym-vpnd prebuilt binary
-vpnd_tag=nym-vpn-core-v0.1.10
-vpnd_version=0.1.10
+vpnd_tag=nym-vpn-core-v0.1.11
```
octol, line 42: I would suggest fetch the latest so we don't have to update this manually

PR #838, opened 2024-08-05
Include probe data in vpnd gw responses

octol:


nym-vpn-core/crates/nym-vpnd/src/types/gateway.rs
```diff
@@ -138,12 +138,51 @@ impl From<nym_vpn_lib::gateway_directory::Location> for Location {
     }
 }
 
+impl From<nym_vpn_lib::gateway_directory::Entry> for Entry {
```
pronebird, line 141: Are we layering the internal lib or simply adding shims for grpc?

pronebird (commented):
seen

nym-vpn-core/crates/nym-vpnd/src/types/gateway.rs
octol: Yeah the internal daemon representation of a gateway, which I wanted to keep separate from the protobuf responses

neacsu (approved):

PR #839, opened 2024-08-05
refactor(vpnx): query countries from grpc

doums:
- switch to grpc to query countries list (entry & exit)
- drop some of the "fastest node location" feature (inactive code) since it's pretty outdated and should be refactored to be provided by grpc as well
- simplify selected node location state: drop it from the Rust side as it's useless: it is already stored in the embedded db and as state in js (avoid state duplication)


nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -338,6 +340,78 @@ impl GrpcClient {
...
+            .iter()
+            .filter_map(|gateway| {
+                // TODO comment this for now has daemon is not returning any probe data
+                // let check = gateway
+                //     .last_probe
+                //     .as_ref()
+                //     .and_then(|probe| probe.outcome.as_ref())
+                //     .and_then(|probe| probe.as_entry.as_ref())
+                //     .map(|v| {
+                //         if v.can_connect && v.can_route {
```
pronebird, line 368: Probably better if we don't stash the commented out code in our database, unless it's a code sample.

nym-vpn-x/src-tauri/src/http.rs
```diff
@@ -4,7 +4,7 @@ use once_cell::sync::Lazy;
 use reqwest::Client;
 use tracing::error;
 
-pub static HTTP_CLIENT: Lazy<Client> = Lazy::new(|| {
+pub static _HTTP_CLIENT: Lazy<Client> = Lazy::new(|| {
```
pronebird, line 7: Does it need to public? Looks like you're trying to hide it by prepending it with `_`.

pronebird (changes requested):

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -338,6 +340,78 @@ impl GrpcClient {
...
+            .iter()
+            .filter_map(|gateway| {
+                // TODO add probe data check
+                Country::try_from(gateway).ok()
+            })
+            .dedup()
+            .sorted_by(|a, b| a.name.cmp(&b.name))
+            .collect();
+
+        Ok(countries)
```
octol, line 412: If there is only a single gateway for a country, and that country is already selected as entry, then you need to exclude it from the list of countries

nym-vpn-x/src-tauri/src/http.rs
doums: I prepended it with `_` because it's not used anymore, but as it could be still useful for the future, I want to keep the declaration

nym-vpn-x/src-tauri/src/grpc/client.rs
doums: Hmm good info. At first glance I think this "defensive" logic should be handled in the daemon side and shouldn't be handled by clients, as all clients would have to implement this same logic. As far I can tell (but maybe Im missing some part of the picture here) daemon would have to filter by that logic the gateway lists before returning them as rpc result.
What do you think?

octol: The list of available countries is fundamentally different from the set of "selectable" countries. The latter can only be known where the selected state is stored, which I think currently is at the app level. Obviously the state can be moved away from the app, and possibly it should at some point, but it's not planned anytime soon

Fundamentally this is caused by the wrong state being held at the app layer. Currently it's "country" when it should be either

A. The app selects a specific Gateway from the some country selected by the user, which would be the state it could track. Then you can easily filter user choices for exit gateways.

Or

B. The app keeps no state at all, and asks the daemon to select entry and exit as separate RPC calls, which would allow the daemon to provide a list of "selectable" gateway choices. The downside here is that this would not cover the uniffi clients, which would have to implement the same logic anyway.

octol: Looking ahead a bit, and taking ecash and the new ticket books into account, I more and more think we need to implement B at some point. So the scope for this PR is more along the lines of "how large % of connect failures can the app tolerate" until that happens? No filtering is simpler to implement but will lead to a % of connect fail. Filter on the app side will drastically reduce that (but not eliminate entirely) until we have B implemented, which will be maybe in a few months.

octol: And for completeness, this check that the exit is not the same as entry happens in the daemon before it actually tries to connect, and it should have a dedicated error ID for the UI to handle. So another way to handle it is to workout a UI flow that takes this into account

octol: Thinking a bit more, I think this last solution is the best for now. It's the most bang for the buck, so to say

doums: Okay, updated to handle the new specific vpnd error

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -338,6 +340,62 @@ impl GrpcClient {
...
+
+        let countries = response
+            .get_ref()
+            .gateways
+            .iter()
+            .filter_map(|gateway| {
+                // TODO add probe data check once daemon returns some data
+                Country::try_from(gateway).ok()
+            })
+            .dedup()
```
octol, line 364: I think `dedup` assumes that it's sorted

nym-vpn-x/src-tauri/src/grpc/client.rs
```diff
@@ -338,6 +340,62 @@ impl GrpcClient {
...
+
+        let countries = response
+            .get_ref()
+            .gateways
+            .iter()
+            .filter_map(|gateway| {
+                // TODO add probe data check
+                Country::try_from(gateway).ok()
+            })
+            .dedup()
```
octol, line 392: Same as other one

nym-vpn-x/src-tauri/src/http.rs
pronebird: Why not deleting the dead code? We can always revive the old code from git history of needed.

pronebird (approved):

PR #847, opened 2024-08-06
Sequentially query entry and exit

neacsu:


nym-vpn-core/patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/nym-vpn-core/nym-vpn-lib/src/mixnet_connect.rs b/nym-vpn-core/nym-vpn-lib/src/mixnet_connect.rs
```
pronebird, line 1: Did you mean to delete this patch file?

neacsu: Ah right, my bad, thanks for the catch

PR #860, opened 2024-08-08
Setup initial workflow files build building nym-vpn-x

octol:
Time has come to update `nym-vpn-x` to use the newer structure for building releases. This is the initial set of files. They are inactive until they have been further iterated on and confirmed to produce the expected outcome

### DIFF
--- a/.github/workflows/build-nym-vpn-x-linux.yml
+++ b/.github/workflows/build-nym-vpn-x-linux.yml
@@ -1,0 +1,124 @@
+name: build-nym-vpn-x-linux
+on:
+  workflow_dispatch:
+  workflow_call:
+    secrets:
+      TAURI_PRIVATE_KEY:
+        required: true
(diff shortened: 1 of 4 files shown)


.github/workflows/build-nym-vpn-x-linux.yml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Install cargo-about
+        run: cargo install --locked cargo-about
+
+      - name: Install Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
```
doums, line 58: Surprised to see you finally switched to latest stable protoc (which is a good thing). What changed since you were blocking to keep with the outdated ubuntu system protoc? (https://github.com/nymtech/nym-vpn-client/pull/565)

PR #861, opened 2024-08-08
Call apple specific init of logs

neacsu:


octol (approved):

nym-vpn-core/crates/nym-vpnd/src/logging.rs
```diff
@@ -6,17 +6,23 @@ use tracing_appender::non_blocking::WorkerGuard;
...
-        .from_env()
-        .unwrap()
-        .add_directive("hyper::proto=info".parse().unwrap())
-        .add_directive("netlink_proto=info".parse().unwrap());
-
-    tracing_subscriber::fmt()
-        .with_env_filter(filter)
-        .compact()
-        .init();
+    #[cfg(any(target_os = "ios", target_os = "macos"))]
```
pronebird, line 9: this will not be compiled on iOS I assume?

neacsu: Ah good question, I'm not 100% sure, but if for iOS we won't be using this daemon implementation, then it's probably not necessary to specify ios target macro here.

nym-vpn-core/crates/nym-vpnd/src/cli.rs
```diff
@@ -30,33 +29,38 @@ pub(crate) struct CliArgs {
...
+    #[cfg(windows)]
     #[arg(long)]
     pub(crate) install: bool,
 
+    #[cfg(windows)]
     #[arg(long)]
     pub(crate) uninstall: bool,
 
     #[arg(long)]
     pub(crate) start: bool,
```
octol, line 48: On windows start here means send an event to start the background service already registered and handled by windows. Is it the same here? Otherwise I would suggest use the other flag

nym-vpn-core/crates/nym-vpnd/src/logging.rs
pronebird: I think so too.

nym-vpn-core/crates/nym-vpnd/src/cli.rs
pronebird: Yeah I don't quite understand what start means, maybe we should document this command?

PR #862, opened 2024-08-08
Invert MacOS as_service flag

neacsu:
As on MacOS the service is run without any flag and that can't be modified, we need to have an inverted flag, at least for this platform, which the default value (false) means running it as a service => `run_as_cli`

### DIFF
--- a/nym-vpn-core/crates/nym-vpnd/src/cli.rs
+++ b/nym-vpn-core/crates/nym-vpnd/src/cli.rs
@@ -52,7 +52,7 @@ pub(crate) struct Command {
     #[arg(long)]
     pub(crate) run_as_service: bool,
 
-    #[cfg(unix)]
+    #[cfg(target_os = "macos")]
     #[arg(long)]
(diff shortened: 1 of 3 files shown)


nym-vpn-core/crates/nym-vpnd/src/cli.rs
```diff
@@ -48,8 +48,13 @@ pub(crate) struct Command {
     #[arg(long)]
     pub(crate) start: bool,
 
+    #[cfg(windows)]
     #[arg(long)]
     pub(crate) run_as_service: bool,
+
+    #[cfg(unix)]
+    #[arg(long)]
+    pub(crate) run_as_cli: bool,
```
octol, line 57: Maybe make this only for mac? Otherwise you'll need to edit all the systemd stuff and vpnx installers too

octol (approved):

pronebird:
Can't we add a program argument in our plist in `/Library/LaunchDaemons` in order to avoid additional CLI flag?

PR #871, opened 2024-08-12
Update uniffi API to use gateway-directory and not vpn api directly

octol:
Update uniffi functions fetching the country lists to use the gateway-directory and not the vpn api client directly

NOTE: this change requires updating the android and ios client


octol:
Tagging @rokas-ambrazevicius  and @zaneschepke for adding the finishing touches

rokas-ambrazevicius (approved):

zaneschepke (approved):

PR #885, opened 2024-08-12
Restore low-latency entry gateway option

octol:
Restore the option to select the entry gateway by probing the latency

Depends on
- https://github.com/nymtech/nym/pull/4759
- https://github.com/nymtech/nym-vpn-client/pull/929


nym-vpn-core/crates/nym-gateway-directory/Cargo.toml
```diff
@@ -6,21 +6,22 @@ license.workspace = true
 
 [dependencies]
 async-trait.workspace = true
-chrono = "0.4.38"
+chrono.workspace = true
```
jstuczyn, line 9: this is definitely out of scope (so don't change it now), but why are we pulling in chrono? we're using `time` literally everywhere else : )

nym-vpn-core/crates/nym-gateway-directory/src/entries/gateway.rs
```diff
@@ -9,13 +9,20 @@ use tracing::error;
...
+type Perfomance = f64;
+
 #[derive(Clone, Debug)]
 pub struct Gateway {
     pub identity: NodeIdentity,
     pub location: Option<Location>,
     pub ipr_address: Option<IpPacketRouterAddress>,
     pub authenticator_address: Option<AuthAddress>,
     pub last_probe: Option<Probe>,
+    pub host: Option<nym_topology::NetworkAddress>,
```
jstuczyn, line 22: does it make sense that the whole struct is bunch of `Option` fields? isn't it the case that once you grab the data you either have most of them or nothing at all?

also, I wonder, from the usability point of view, would it perhaps make more sense to leave host as a `String` instead?

nym-vpn-core/crates/nym-gateway-directory/src/entries/gateway.rs
```diff
@@ -37,6 +44,20 @@ impl Gateway {
     pub fn has_ipr_address(&self) -> bool {
         self.ipr_address.is_some()
     }
+
+    pub fn clients_address_no_tls(&self) -> Option<String> {
+        match (&self.host, &self.clients_ws_port) {
+            (Some(host), Some(port)) => Some(format!("ws://{}:{}", host, port)),
```
jstuczyn, line 50: for example with this, if you have `host`, you **must** have the `ws` port. there are no situations under which you can have one but not the other

nym-vpn-core/crates/nym-gateway-directory/src/entries/gateway.rs
```diff
@@ -179,12 +204,20 @@ impl TryFrom<nym_validator_client::models::DescribedGateway> for Gateway {
                     .inspect_err(|err| error!("Failed to parse authenticator address: {err}"))
                     .ok()
             });
+        let gateway = nym_topology::gateway::Node::try_from(gateway).ok();
+        let host = gateway.clone().map(|g| g.host);
+        let clients_ws_port = gateway.clone().map(|g| g.clients_ws_port);
```
jstuczyn, line 209: nitpicking: you don't have to be cloning the whole `gateway`, since `u16` is `Copy`

nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
```diff
@@ -200,7 +198,9 @@ impl GatewayClient {
                     .inspect_err(|err| error!("Failed to parse gateway: {err}"))
                     .ok()
             })
-            .collect();
+            .collect::<Vec<_>>();
+        let skimmed_gateways = self.lookup_skimmed_gateways().await?;
+        append_performance(&mut gateways, skimmed_gateways);
```
jstuczyn, line 203: do we actually need the data from described gateways? if so, what is that so that we could in the future in the `semi-skimmed` endpoint : )

jstuczyn (approved):

nym-vpn-core/crates/nym-gateway-directory/Cargo.toml
octol: Ah good to know, I've forgotten the reason why this is used, likely because it's spread here from explorer-api

nym-vpn-core/crates/nym-gateway-directory/src/entries/gateway.rs
octol: It's `Option` because you don't have them currently in the nym-vpn-api endpoints yet, so we append it at a later stage. But yeah some of these will go away in the future

nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
octol: It's for the IPR and authenticator addresses. But ideally we shouldn't need either, it should all come from the nym-vpn-api and this is a bandaid until it's added over there :)

PR #892, opened 2024-08-13
Fork minimal subset of old credential crates and switch to latest develop

octol:
Credentials have been reworked in the platform repo with the advent of ecash, making the vpn client in it's current form incompatible. This PR forks a minimal subset of these crates so that we can provide a smooth migration over to the new ecash system. It also updates to the latest rev of the platform repo.


octol:
I've realized that this is currently failing since the mixnet client is pulling in the new ecash enabled credential-storage crate, which applies an incompatible db migration

octol:
https://github.com/nymtech/nym-vpn-client/pull/929/ is a better solution so closing this

PR #913, opened 2024-08-16
Add two-hop WireGuard for iOS

pronebird:
- Add two-hop WireGuard integration for iOS
- Rework OSTunProvider to match the system API available on iOS for configuring tunnel interface, routes, dns etc..
- Add custom runner for wireguard on iOS. I am hoping we can harmonize the overall implementation between desktop and mobile but I have no energy to do that right now as the scope is only growing.


nym-vpn-apple/NymMixnetTunnel/PacketTunnelProvider.swift
```diff
@@ -4,100 +4,277 @@ import NymLogger
...
             throw PacketTunnelProviderError.invalidSavedConfiguration
         }
 
-        let callback: () -> Void = { [weak self] in
-            guard let config = self?.mixnetTunnelProvider.nymConfig
-            else {
-                return
-            }
+        var vpnConfig = try mixnetConfig.asVpnConfig(tunProvider: self, tunStatusListener: self)
+        // todo: figure out WHAT always defaults this setting to .randomLowLatency
```
rokas-ambrazevicius, line 49: https://github.com/nymtech/nym-vpn-client/blob/94ba6cda9598710f42c8f17935de9dbb7642dadd/nym-vpn-apple/Services/Sources/Services/ConnectionManager/ConnectionStorage.swift#L29

Changing this to `.random`

rokas-ambrazevicius: or here https://github.com/nymtech/nym-vpn-client/blob/97db4f4eb056d62fc5e725acda6f0423d23f366c/nym-vpn-apple/Services/Sources/Services/TunnelMixnet/Config/MixnetConfig.swift#L67

pronebird: Rokas have you already fixed that right?

rokas-ambrazevicius: Should be working in any case after this was merged:
https://github.com/nymtech/nym-vpn-client/pull/885

zaneschepke (approved):

zaneschepke (approved):

zaneschepke (approved):

zaneschepke (approved):

pronebird:
@octol @neacsu @zaneschepke @rokas-ambrazevicius please leave your review early next week, preferably on Monday because we gotta move on with this and @zaneschepke is very close to bringing support for wg on android too.

.github/workflows/ci-nym-vpn-core.yml
```diff
@@ -202,7 +202,8 @@ jobs:
         if: contains(matrix.os, 'macos') || contains(matrix.os, 'mac-m1')
         working-directory: nym-vpn-core
         run: |
-          cargo build --verbose --target aarch64-apple-ios -p nym-vpn-lib
+          rustflags="-L ${GITHUB_WORKSPACE}/build/lib/aarch64-apple-ios"
+          RUSTFLAGS=$rustflags cargo build --verbose --target aarch64-apple-ios -p nym-vpn-lib
```
octol, line 206: General comment _not_ for this PR:
The platform stuff is really reaching a point where we should try to take a step back and rework it. Also I'm kinda think that this part here is the result of building both mac and ios during a single job while splitting them into separate seems likely more appropriate (like how android is separate from linux).

nym-vpn-apple/MixnetLibrary/Package.swift
```diff
@@ -23,8 +23,8 @@ let package = Package(
         ),
         .binaryTarget(
             name: "NymVpnLib",
-            url: "https://github.com/nymtech/nym-vpn-client/releases/download/nym-vpn-core-v0.1.16/nym-vpn-core-v0.1.16_ios_universal.zip",
-            checksum: "41a5ecb8b0ab7cc2e6130179863761d12bcc988eda6c269cb00bd01a43449741"
+            url: "https://github.com/nymtech/nym-vpn-client/releases/download/nym-vpn-core-nightly/nym-vpn-core-v0.1.17-dev_ios_universal.zip",
+            checksum: "66a830cd46b912df1c11704b487f7a661299ff4583d142770ae04c0ceeaaad5c"
```
octol, line 27: This will break after the next nightly build?

nym-vpn-core/nym-vpn-lib/src/lib.rs
```diff
@@ -47,3 +51,10 @@ pub use crate::{
         SpecificVpn,
     },
 };
+
+pub const DEFAULT_DNS_SERVERS: [IpAddr; 4] = [
```
octol, line 55: nitpick: separate config.rs file or similar?

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
```diff
@@ -0,0 +1,64 @@
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+
+/// The type alias for the receiving end of path observer.
+pub type DefaultPathReceiver = UnboundedReceiver<OSDefaultPath>;
+
+/// Observer type that wraps network path changes into a channel.
+#[derive(Debug)]
+pub struct DefaultPathObserver {
```
octol, line 8: I like this, wrapping channels in a strong type. Imho we should do this more often

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
```diff
@@ -0,0 +1,64 @@
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
```
octol, line 1: Side comment: for unbounded receivers/senders there is no confusion, but for bounded senders/receivers then the type has the same name for mpsc/broadcast/oneshot etc so it's nice to not import the whole path so it's easier to see in the code what channel it is

nym-vpn-core/nym-vpn-lib/src/mobile/runner.rs
```diff
@@ -0,0 +1,427 @@
...
+    shutdown_token: CancellationToken,
+}
+
+impl WgTunnelRunner {
+    pub fn new(config: VPNConfig, shutdown_token: CancellationToken) -> Result<Self> {
+        let user_agent = UserAgent::from(nym_bin_common::bin_info_local_vergen!());
+        tracing::info!("User agent: {user_agent}");
+
+        let generic_config = GenericNymVpnConfig {
+            mixnet_client_config: MixnetClientConfig {
```
octol, line 76: Unrelated side comment: we should implement `Default` for `MixnetClientConfig` in the future

nym-vpn-core/nym-vpn-lib/src/mobile/runner.rs
```diff
@@ -0,0 +1,427 @@
...
+                // Put in some manual error handling, the correct long-term solution is that handling
+                // errors and diconnecting the mixnet client needs to be unified down this code path
+                // and merged with the mix tunnel one.
+                mixnet_client.disconnect().await;
+                return Err(err);
+            }
+        };
+
+        let bandwidth_controller =
+            BandwidthController::new(mixnet_client.clone(), self.task_manager.subscribe());
```
octol, line 133: Side comment for the future:
Imho we should switch BandwidthController to use CancellationToken. Also, BandwidthController calls `mixnet_client.disconnect()` I just spotted, not sure what the implication of that is ...

nym-vpn-core/nym-vpn-lib/src/mobile/runner.rs
```diff
@@ -0,0 +1,427 @@
...
+        let SelectedGateways { entry, exit } = self.select_gateways().await?;
+        let mixnet_client = self.start_mixnet_client(&entry).await?;
+
+        let auth_addresses = match self.setup_auth_addresses(&entry, &exit) {
+            Ok(auth_addr) => auth_addr,
+            Err(err) => {
+                // Put in some manual error handling, the correct long-term solution is that handling
+                // errors and diconnecting the mixnet client needs to be unified down this code path
+                // and merged with the mix tunnel one.
+                mixnet_client.disconnect().await;
```
octol, line 127: I have a strong suspicion that this is a no-op since we pass in an external `task_manager.subscribe()` when creating the mixnet client, and the way to disconnect the mixnet_client is to call `task_manager.signal_shutdown()`

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl OSDefaultPathObserver for DefaultPathObserver {
+    fn on_default_path_change(&self, new_path: OSDefaultPath) {
+        if self.tx.send(new_path).is_err() {
+            tracing::warn!("Failed to send default path change.");
+        }
+    }
+}
+
+#[derive(uniffi::Enum, Debug)]
```
octol, line 26: What's the thinking when deciding between placing stuff in this mod vs `platform`? Is `platform` something that will be phased out in time?

octol (approved):
Looking great! Well done!

I left some side comments for stuff I noticed while reading the code, some which refers to code not touched by this PR. Feel free to do what you want with them

nym-vpn-apple/MixnetLibrary/Package.swift
rokas-ambrazevicius: This will break. I believe this is a temp fix to get the Testflight out the door.

octol: If you need a custom build of the vpn lib that is non-temporary you can use a named tag like `nym-vpn-core-ios-testflight`. But maybe if we merge this on Monday we can create a new release core then and use that?

rokas-ambrazevicius: That would be ideal 👍

.github/workflows/ci-nym-vpn-core.yml
pronebird: Yeah that's legit. Let's add a card for this and attack it in a separate PR.

nym-vpn-apple/MixnetLibrary/Package.swift
pronebird: we'll fix this with a follow up pr

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
pronebird: Correct. Bounded are different story.

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
pronebird: Note that this struct implements `OSDefaultPathObserver` trait, making it a conduit for messages received via FFI boundary.

nym-vpn-core/nym-vpn-lib/src/mobile/ios/default_path_observer.rs
pronebird: Simply a logical grouping of related types. Everything under `mobile/ios` in essence is a platform-specific stuff. I didn't want to move things too apart since it breaks connection between types.

nym-vpn-core/nym-vpn-lib/src/mobile/runner.rs
pronebird: Yeah what BandwidthController does is it seems that it just waits for task manager to shutdown before killing the mixnet client. I am not quite sure why it's done this way, this could be re-arranged.

nym-vpn-core/nym-vpn-lib/src/lib.rs
pronebird: I'll do that in a separate PR to unblock people.

nym-vpn-core/nym-vpn-lib/src/mobile/runner.rs
pronebird: It might be, I ran out of time to investigate this any closer.

PR #928, opened 2024-08-19
Use lighter linker for windows build

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/928)


octol (approved):
Builds and links when I tested locally

pronebird:
What's the impact of this change?

neacsu:
> What's the impact of this change?

CI rust builds on windows run into OOM, when the ci machine has been run for a longer time. This linker is supposed to use less memory, but in order to reproduce it, we need the machine to start OOMing again and see if this PR fixes that.
PS: We already tried bumping up the RAM on the windows machine to 32GB, put that still isn't enough

PR #937, opened 2024-08-20
feat(vpnx): add network env file to config

doums:
Add to the app config file a new field `network_env_file` to set a custom network env

### DIFF
--- a/nym-vpn-x/src-tauri/src/fs/config.rs
+++ b/nym-vpn-x/src-tauri/src/fs/config.rs
@@ -4,6 +4,8 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Default, Serialize, Deserialize, Debug, Clone)]
 pub struct AppConfig {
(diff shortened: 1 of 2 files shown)


octol (commented):
How does this interact with stuff that is saved on disk? Is it separate like the daemon does it or is there a risk that stuff gets stored in locations where it now becomes incompatible with mainnet?

doums:
The only data which is network dependent, stored on client side is:
- entry and exit country -> not an issue since if an invalid country is selected, it is automatically updated to a random available country (+ user is notified)
- any saved credential expiry date -> not an issue, user would have to import a new credential after switching network anyway, the only "risk" is UI showing cred expiry date for a cred valid for another network
Note: this should be fixed once daemon expose the required API to query cred data (linked to the current network), which is a pending/planned task afaik

zaneschepke (approved):

PR #950, opened 2024-08-21
Send shutdown sequentially

neacsu:
Should fix the libwg panics on shutdown


pronebird (approved):
This makes a lot of sense. I should also look into fixing this on libwg side.

octol (approved):

PR #963, opened 2024-08-22
Update logs view and fix some issues

rokas-ambrazevicius:
Update logs view to Figma design.
Fix an issue where fileHandle is not found after removal of logs.

### DIFF
--- a/nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
+++ b/nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
@@ -27,7 +27,7 @@ public class FileLogHandler: LogHandler {
         guard let logsDirectory else { return nil }
 
         try? fileManager.createDirectory(at: logsDirectory, withIntermediateDirectories: true, attributes: nil)
-        let logFileURL = logsDirectory.appendingPathComponent("Log.log")
+        let logFileURL = logsDirectory.appendingPathComponent(Constants.logFileName.rawValue)
         return logFileURL
     }
 
@@ -63,10 +63,20 @@ public class FileLogHandler: LogHandler {
         let data = Data(logLine.utf8)
 
         ioQueue.async {
-            if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
-                self.fileHandle = try? FileHandle(forWritingTo: logFileURL)
-            }
-            try? self.fileHandle?.write(contentsOf: data)
+            do {
+                if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
+                    self.fileHandle = try FileHandle(forWritingTo: logFileURL)
+                }
+                try self.fileHandle?.write(contentsOf: data)
+            } catch CocoaError.fileNoSuchFile {
+                guard let logFileURL = FileLogHandler.logFileURL,
+                      !FileManager.default.fileExists(atPath: logFileURL.relativePath)
+                else {
+                    return
+                }
+                FileManager.default.createFile(atPath: logFileURL.relativePath, contents: nil, attributes: nil)
+                self.fileHandle = nil
+            } catch {}
         }
     }
 }
--- a/nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
+++ b/nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
@@ -4,37 +4,15 @@ import Constants
 
 public class FileLogHandler: LogHandler {
     private let label: String
+    private let logFileManager: LogFileManager
 
-    public init(label: String) {
+    public init(label: String, logFileManager: LogFileManager = LogFileManager.shared) {
         self.label = label
+        self.logFileManager = logFileManager
     }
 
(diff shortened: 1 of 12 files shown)


nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
```diff
@@ -63,10 +63,24 @@ public class FileLogHandler: LogHandler {
...
-            if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
-                self.fileHandle = try? FileHandle(forWritingTo: logFileURL)
+            do {
+                if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
+                    self.fileHandle = try FileHandle(forWritingTo: logFileURL)
+                }
+                try self.fileHandle?.write(contentsOf: data)
+            } catch {
+                // The file “Logs.log” doesn’t exist.
+                let nsError = error as NSError
```
pronebird, line 73: Perhaps match against `CocoaError.fileNoSuchFile`

pronebird: Alternatively just do `FileManager.default.createFile(..)` before `self.fileHandle = try FileHandle(forWritingTo: logFileURL)`, it shouldn't do anything if file exists I think.

rokas-ambrazevicius: > Alternatively just do `FileManager.default.createFile(..)` before `self.fileHandle = try FileHandle(forWritingTo: logFileURL)`, it shouldn't do anything if file exists I think.

If a file already exists at path, this method overwrites the contents of that file if the current process has the appropriate privileges to do so.

https://developer.apple.com/documentation/foundation/filemanager/1410695-createfile

pronebird: Uh ok. Then check if it exists I guess.

nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
```diff
@@ -63,10 +63,20 @@ public class FileLogHandler: LogHandler {
...
+                }
+                try self.fileHandle?.write(contentsOf: data)
+            } catch CocoaError.fileNoSuchFile {
+                guard let logFileURL = FileLogHandler.logFileURL,
+                      !FileManager.default.fileExists(atPath: logFileURL.relativePath)
+                else {
+                    return
+                }
+                FileManager.default.createFile(atPath: logFileURL.relativePath, contents: nil, attributes: nil)
+                self.fileHandle = nil
```
pronebird, line 78: You need to open log file after creating it and log the line. Otherwise you basically drop the logged line.

pronebird: Maybe we should create the file beforehand to avoid duplicating the code?


```swift
if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
    if !<file exists?> {
        FileManager.default.createFile(atPath: logFileURL.path(false), contents: nil, attributes: nil)
    }
    self.fileHandle = try? FileHandle(forWritingTo: logFileURL)
    try self.fileHandle?.write(contentsOf: data)
}
```

pronebird (changes requested):

nym-vpn-apple/Services/Sources/Services/NymLogger/FileLogHandler.swift
pronebird: Maybe we should create the file beforehand to avoid duplicating the code?


```swift
if self.fileHandle == nil, let logFileURL = FileLogHandler.logFileURL {
    if !<file exists?> {
        FileManager.default.createFile(atPath: logFileURL.path(false), contents: nil, attributes: nil)
    }
    self.fileHandle = try? FileHandle(forWritingTo: logFileURL)
    try self.fileHandle?.write(contentsOf: data)
}
```

nym-vpn-apple/Services/Sources/Services/NymLogger/LogFileManager.swift
```diff
@@ -0,0 +1,59 @@
...
+import Constants
+
+public final class LogFileManager {
+    public static let shared = LogFileManager()
+
+    init() {
+        configure()
+    }
+
+    public var fileHandle: FileHandle?
```
pronebird, line 11: `fleHandle` leaks outside of `LogFileManager`

If we are going to share it with other `FileLogHandler` types then I think it would be sensible to handle all io in here and make sure we use a dispatch queue to serialize calls.

As a template what comes to my mind:

```swift
private let queue = DispatchQueue(label: "..")

public func write(str: String) {
  queue.async { 
     // open or create file if file handle is nil
     self.fileHandle = ...
  }
}

public func deleteFile() {
  queue.async {
    // delete file
    // release file handle
    self.fileHandle = nil
  }
}
```

pronebird (changes requested):

pronebird (approved):

PR #965, opened 2024-08-22
Extract out nym-wg-gateway-client crate

octol:
Extract out `nym-wg-gateway-client` into its own crate, as it's a mostly independent module

- **Remove bare AddrParseError**
- **Rename config to wireguard_config mod**
- **Restrict pub -> pub(crate) across the board**
- **Make function private**
- **Extract out nym-wg-gateway-client crate**


dynco-nym (approved):

octol:
Thanks guys for the feedback, I'll fold that into the follow-up PR :)

PR #967, opened 2024-08-22
Add wireguard-go wrapper, netstack and two hop forwarder

pronebird:
1. go: Add netstack support to enable running WG over virtual tcp/ip stack 
2. go: Add udp forwarder to be able to inject WireGuard into WireGuard.
3. go: Renamed tunnel container into container with generic element inside to be able to store different things inside not just tunnels.
4. Add bindings via nym-wg-go. I am aware of some duplication of PrivateKey/PublicKey types etc which we already have in nym-crypto but a. they lack a few things b. the display impl. prints keys in base58 which is fairly in WireGuard which prefers base64. At least during debugging and in general base64  can be copied verbatim between configs. So for that reason I rolled my own types, but we can probably take another stab at it and see if we can reuse nym-crypto.


zaneschepke (approved):

octol (commented):
Very nice, I like it :)
So for some reason it breaks wg mode on my linux box, not sure if this is due to one of the many go dependencies that are updated or the code changes

pronebird:
> Very nice, I like it :)
> 
> So for some reason it breaks wg mode on my linux box, not sure if this is due to one of the many go dependencies that are updated or the code changes 

I will look into that today. The libwg interface should be the same though, so not entirely sure what went wrong.

.github/workflows/build-nym-vpn-android.yml
```diff
@@ -104,9 +104,9 @@ jobs:
             x86_64-linux-android \
             i686-linux-android
 
-      - name: Add cargo-ndk
+      - name: Install cargo deps
         run: |
-          cargo install cargo-ndk
+          cargo install cargo-ndk cargo-license
```
octol, line 109: Suggestion: use --locked

.github/workflows/ci-nym-vpn-core.yml
```diff
@@ -120,6 +130,12 @@ jobs:
         run: |
           ./wireguard/build-wireguard-go.sh
 
+      - name: Build wireguard (iOS)
+        if: contains(matrix.os, 'macos') || contains(matrix.os, 'mac-m1')
+        shell: bash
+        run: |
+          ./wireguard/build-wireguard-go.sh --ios
```
octol, line 137: In the build script it seems like ios and macos are different code paths, but here this command is applied for both?

.github/workflows/ci-nym-vpn-core.yml
```diff
@@ -93,6 +93,16 @@ jobs:
           echo "TRIPLET=$triplet" >> $GITHUB_ENV
           echo "RUSTFLAGS=-L ${GITHUB_WORKSPACE}/build/lib/${triplet}" >> $GITHUB_ENV
 
+      - name: set rust flags
+        working-directory: nym-vpn-core
+        shell: bash
+        run: |
+          case  "$(uname -s)" in
```
octol, line 100: Looks like work in progress but I think it's probably simpler to incorporate any env variable changes in the pre-existing block?

.github/workflows/ci-nym-vpn-core.yml
pronebird: This is a mistake. Let me fix this

pronebird: Removed

.github/workflows/build-nym-vpn-android.yml
pronebird: @zaneschepke what do you think?

.github/workflows/ci-nym-vpn-core.yml
pronebird: Actually. This is correct because we build wireguard-go for 3 platforms here:

- Desktop
- Android
- iOS (added in this PR)

.github/workflows/ci-nym-vpn-core.yml
pronebird: I don't recall adding this. @zaneschepke was it your work?

octol (approved):
Tested locally and works for me on linux :)

.github/workflows/ci-nym-vpn-core.yml
octol: Yep looks good!

PR #976, opened 2024-08-23
Use bin_info_local_vergen and remove re-export of nym-bin-common

octol:
To make sure the bin_info data passed to UserAgent comes from the local callsite, use `bin_info_local_vergen` instead of `bin_info`. The issue comes from the fact that the dependency crate lives in the platform repo where we can just use `bin_info` since the git commit SHA1 is the same anyway.

Remove re-export of `nym-bin-common`


pronebird (approved):
Good work!

PR #985, opened 2024-08-26
De-duplicate some task handling in nym-vpn-lib

octol:
This is the start of an attempt to simplify task handling in the vpn lib and move some of it away from the lib to the binary crate. This is PR is not a complete solution but a step in the right direction, I hope.

- Switch `nym-vpn-cli` to use to use the same task handling as the daemon and uniffi clients, so that we can de-duplicate some of the code in `nym-vpn-lib` that handles waiting for the running task and interrups and delete it.

- Put in a 1 sec sleep as a hack until the task handling in `nym-vpnd` is unified with `nym-vpn-lib`.

- Stop using `nym_task::wait_for_signal_and_error` to try to push signal handling one level up. This is not the endgoal, but should bring us one step closer.

- Handle a few failures that didn't disconnect the mixnet client in the wireguard path, and put in a note about what's missing.

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-cli/src/error.rs
+++ b/nym-vpn-core/crates/nym-vpn-cli/src/error.rs
@@ -27,8 +27,8 @@ pub(crate) enum Error {
     #[error("failed to parse encoded credential data")]
     FailedToParseEncodedCredentialData(#[source] bs58::decode::Error),
 
-    #[error(transparent)]
-    Boxed(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
+    #[error("{0}")]
+    VpnRun(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
 
     #[error("vpn task stopped unexpectedly")]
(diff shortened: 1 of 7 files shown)


nym-vpn-core/crates/nym-vpn-cli/src/error.rs
```diff
@@ -26,6 +26,12 @@ pub(crate) enum Error {
 
     #[error("failed to parse encoded credential data")]
     FailedToParseEncodedCredentialData(#[source] bs58::decode::Error),
+
+    #[error("{0}")]
```
pronebird, line 30: In some of your PRs you used `#[transparent]` instead. Does it not work for a boxed error?

pronebird (approved):

nym-vpn-core/crates/nym-vpn-cli/src/error.rs
octol: It only works when you have `#[from]`, which in this case I wanted to avoid

PR #986, opened 2024-08-26
Remove the ctrl-c signal handler from nym-vpn-lib

octol:
Remove the Ctrl-C signal handler from `nym-vpn-lib` and instead start it from the cli crate, which in turns sends a stop signal that gets the whole shutdown procedure rolling

NOTE: there is still a second signal handler hidden in force to exit step

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-cli/src/main.rs
+++ b/nym-vpn-core/crates/nym-vpn-cli/src/main.rs
@@ -5,6 +5,7 @@ use std::{fs, path::PathBuf};
 
 use clap::Parser;
 use commands::{CliArgs, ImportCredentialTypeEnum};
+use futures::channel::mpsc;
 use nym_vpn_lib::{
     bin_common::bin_info,
     gateway_directory::{Config as GatewayConfig, EntryPoint, ExitPoint},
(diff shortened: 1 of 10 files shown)


nym-vpn-core/crates/nym-vpn-cli/src/main.rs
```diff
@@ -5,6 +5,7 @@ use std::{fs, path::PathBuf};
 
 use clap::Parser;
 use commands::{CliArgs, ImportCredentialTypeEnum};
+use futures::channel::mpsc;
```
pronebird, line 8: We could probably default to using [tokio::sync::mpsc](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html) when possible

nym-vpn-core/crates/nym-vpn-cli/src/main.rs
```diff
@@ -178,27 +179,21 @@ async fn run_vpn(args: commands::RunArgs, data_path: Option<PathBuf>) -> Result<
...
-        Ok(nym_vpn_lib::NymVpnExitStatusMessage::Failed(err)) => {
-            debug!("VPN exited with error: {:?}", err);
-            Err(Error::VpnRun(err))
-        }
-        Err(err) => {
-            debug!("VPN exited with error: {:?}", err);
-            Err(Error::UnexpectedStop)
-        }
-    }
+fn register_signal_handler(vpn_ctrl_tx: mpsc::UnboundedSender<nym_vpn_lib::NymVpnCtrlMessage>) {
```
pronebird, line 187: We might want to listen for `SIG_TERM` too as those are common:

```rs
use tokio::signal::unix::{signal, SignalKind};
let mut signal_stream = signal(SignalKind::terminate())?;
signal_stream.recv().await;
vpn_ctrl_tx.unbounded_send(..).unwrap();
```

pronebird (approved):
Left a few comments otherwise ok.

PR #1001, opened 2024-08-29
Apple: Update settings order

rokas-ambrazevicius:


nym-vpn-apple/Settings/Sources/Settings/SettingsViewModel.swift
```diff
@@ -14,7 +14,6 @@ public class SettingsViewModel: SettingsFlowState {
         [
             connectionSection(),
             themeSection(),
-            logsSection(),
```
pronebird, line 17: Are we removing logs?

pronebird (approved):

PR #1026, opened 2024-09-02
ci: add ios target

pronebird:
- Configure iOS target separately to avoid mixing Mac and iOS 
- Separate uniffi-bindgen from vpn-lib to avoid building the entire thing only to run bindgen
- Xcode: automatically download env files when they aren't present in the file system


octol (approved):

nym-vpn-core/uniffi-bindgen/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
octol, line 1: Please move the new crate to `crates`

octol (approved):

PR #1038, opened 2024-09-04
cargo +nightly fmt -- --config imports_granularity=Crate,group_imports=StdExternalCrate

octol:
Run `cargo +nightly fmt -- --config imports_granularity=Crate,group_imports=StdExternalCrate`


pronebird:
I like it, the only thing I'd add .rustfmt config because this would enable us to run `cargo +nightly fmt` and stable will only throw a few warnings.

pronebird (approved):

PR #1051, opened 2024-09-04
Handle a bunch more wg related errors

octol:


proto/nym/vpn.proto
```diff
@@ -267,6 +267,18 @@ message Error {
...
+    OUT_OF_BANDWIDTH_WHEN_SETTING_UP_TUNNEL = 26;
+
+    // Failed to bring up the wg interface due to auth failure
+    FAILED_TO_BRING_INTERFACE_UP_WG_AUTH_FAILED = 27;
+
+    // Failed to bring up the wg interface
+    FAILED_TO_BRING_INTERFACE_UP_WG_DOWN = 28;
+
+    // Failed to bring up the wg interface, tunnel was closed
+    FAILED_TO_BRING_INTERFACE_UP_WG_EVENT_TUNNEL_CLOSE = 29;
```
neacsu, line 281: Those errors seem a bit too specific to talpid imo. Is there any benefit in exposing them outside of vpnd? They sound like very internal error types that wouldn't help UIs with anything.
I think, in case there is any action we could take based on this error differentiation, it should be limited inside vpnd

octol: Yeah agree, I'm going to continue iterating on errors in the vpn lib in a series of upcoming PRs

PR #1061, opened 2024-09-05
Update env vars on build

rokas-ambrazevicius:
Updates env vars automatically if env files are empty.


pronebird (commented):
We already have it defined under `Build > pre-actions` for NymVPN accessible via Product > Scheme > Edit scheme in Xcode or cmd + shift + < shortcut

PR #1068, opened 2024-09-06
Apple: Add environment selector

rokas-ambrazevicius:
IMG_3724

Add environment selector for iOS.

Press and hold version to access the menu.


doums (approved):
lovely stash

PR #1070, opened 2024-09-06
Move select gw and error to sub-mod where it's used

octol:
- Rename to SelectGatewaysError
- Move select gw error to where it's used
- Push select gateways to sub module


octol:
So I'm in two minds about this one. I moved the select gateway functionality to the one place where it's used to improve locality, but fixing CI for mobile I saw it's also used in the mobile runner...

PR #1072, opened 2024-09-06
Apple: Store current env selection in shared app settings

rokas-ambrazevicius:
Store current env in app settings, for tunnel to be able to set values from correct source as well.

### DIFF
--- a/nym-vpn-apple/Services/Sources/Services/ConfigurationManager/ConfigurationManager.swift
+++ b/nym-vpn-apple/Services/Sources/Services/ConfigurationManager/ConfigurationManager.swift
@@ -1,13 +1,21 @@
 import Foundation
+import AppSettings
 import Constants
 import Logging
 
 public final class ConfigurationManager {
+    private let appSettings: AppSettings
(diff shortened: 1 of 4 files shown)


doums (approved):

zaneschepke (approved):

nym-vpn-apple/Services/Sources/Services/ConfigurationManager/ConfigurationManager.swift
```diff
@@ -1,13 +1,21 @@
...
 import Logging
 
 public final class ConfigurationManager {
+    private let appSettings: AppSettings
     private let logger = Logger(label: "Configuration Manager")
 
-    private var currentEnv: Env = .mainnet
+    // Source of truth in AppSettings.
+    // We need to set same settings in tunnel extension as well.
+    private var currentEnv: Env = .mainnet {
```
pronebird, line 12: Maybe then you should simply define a getter and setter that both proxy to `appSettings.currentEnv`?

PR #1073, opened 2024-09-06
Handle a few more error cases in vpnd

octol:
Handle a few more error cases in `nym-vpnd`


proto/nym/vpn.proto
doums, line 1: Should we expect other short-term changes in these new errors?

doums (approved):

proto/nym/vpn.proto
octol: The ones that are unstable have a comment to indicate it

PR #1094, opened 2024-09-09
Use golang from runner image for linux builds

octol:
Now that we generate docker runners programmatically, let's set it up there instead


Bykow (approved):
Nice

PR #1096, opened 2024-09-09
refactor(vpn-x): migrate to tauri v2

doums:
- [x] code migration
- [x] Linux workflows updated


zaneschepke (approved):
🚀

nym-vpn-x/src-tauri/src/fs/path.rs
```diff
@@ -1,24 +1,71 @@
...
-        .map(|mut p| {
-            p.push(APP_DIR);
-            p.push(LOG_DIR);
-            p
-        })
-        .and_then(|p| check_dir(&p).map(|_| p).ok());
+/// The app config directory
+/// - Linux: `$XDG_CONFIG_HOME/nymvpn-x` or `$HOME/.config/nymvpn-x`
+/// - Windows: `{FOLDERID_RoamingAppData}\nymvpn-x`, example  `C:\Users\Pierre\AppData\Roaming\nymvpn-x`
+pub static APP_CONFIG_DIR: Lazy<Option<PathBuf>> = Lazy::new(|| {
```
pronebird, line 12: Typically you'd access the same info from tauri using `App` instance:

```rs
let config_dir = app.path().app_config_dir()?;
```

Maybe there is a reason why this is static, but if you can obtain these paths from tauri and then push them down to dependencies, you may be able to get rid of dependency on `dirs` crate.

https://docs.rs/tauri/2.0.0-rc.11/tauri/path/struct.PathResolver.html

pronebird (changes requested):

nym-vpn-x/src-tauri/src/main.rs
```diff
@@ -201,7 +201,8 @@ async fn main() -> Result<()> {
                 }
             });
 
-            let handle = app.handle();
+            // TODO check if clone works
+            let handle = app.handle().clone();
```
pronebird, line 205: It does and some objects like Tray and others have a method to obtain it via `app_handle()` too, which may come very handy.

nym-vpn-x/src-tauri/src/tray.rs
```diff
@@ -0,0 +1,149 @@
...
+        show_window(tray_icon.app_handle(), false).ok();
+    }
+}
+
+#[instrument(skip(app))]
+fn on_menu_event(app: &AppHandle, event: MenuEvent) {
+    trace!("menu event [{}]", event.id.0);
+
+    match event.id().as_ref() {
+        "ShowHide" => {
```
pronebird, line 51: Make these constants.

nym-vpn-x/src-tauri/src/fs/path.rs
doums: Yes I know. 
In short, I need to access these dirs before tauri app is instantiated for a bunch of initialization which happen before tauri app setup.

And so this is why I'm using dirs crate to resolve them (note that underthewood this is what tauri is using too, so it's the same https://github.com/tauri-apps/tauri/discussions/7757).
Also note that, in tauri v1, the dirs API was exposed outside the app handle. This is not the case anymore. Reason why I had to add `dirs` in deps.

nym-vpn-x/src-tauri/src/tray.rs
doums: Why we'd need that, what value it brings to? Those variants are only relevant in this context of pattern matching (aka the `match` right?).

Maybe you missed the enum above?
```
#[derive(AsRefStr, Debug)]
enum MenuItemId {
    ShowHide,
    Quit,
}
```

nym-vpn-x/src-tauri/src/main.rs
doums: > some objects like Tray and others have a method to obtain it via app_handle() too, which may come very handy

yep I know

Indeed, in tauri v1, idk why but it was not needed to clone the handle ref (and taking owership) see https://github.com/nymtech/nym-vpn-client/blob/45297614d96c3e0b27bf9ccf0bd08215e7009ffb/nym-vpn-x/src-tauri/src/main.rs#L204

now in tauri v2 something as changed.
```
error[E0521]: borrowed data escapes outside of closure
   --> src/main.rs:179:26
    |
157 |         .setup(move |app| {
    |                      ---
    |                      |
    |                      `app` is a reference that is only valid in the closure body
    |                      has type `&'1 mut tauri::App`
...
179 |             let handle = app.handle();
    |                          ^^^^^^^^^^^^
    |                          |
    |                          `app` escapes the closure body here
    |                          argument requires that `'1` must outlive `'static`
```

And at first glance it looked suspicious so I left few TODO comments, but yea it seems to work as expected, so I'll remove them.

nym-vpn-x/src-tauri/src/tray.rs
octol: If it's effectively matching over the menu item enum, then a possible way here would be to construct the enum from the event

.github/workflows/build-nym-vpn-x-linux.yml
```diff
@@ -37,8 +37,9 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt-get update && sudo apt-get install -y libdbus-1-dev libmnl-dev libnftnl-dev \
-            libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev jq \
-            libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 unzip librsvg2-dev file
+            libwebkit2gtk-4.1-dev build-essential curl wget libssl-dev jq \
+            libgtk-3-dev squashfs-tools libayatana-appindicator3-dev make libfuse2 unzip librsvg2-dev file \
+            libsoup-3.0-dev libjavascriptcoregtk-4.1-dev
```
octol, line 42: Once we have merged this, I will see if I can add these to the runner image

.github/workflows/ci-nymvpn-x-rust.yml
```diff
@@ -7,7 +7,6 @@ on:
       - ".github/workflows/ci-nymvpn-x-rust.yml"
       - "nym-vpn-x/src-tauri/**"
       - "nym-vpn-core/crates/nym-vpn-proto/**"
-      - "nym-vpn-core/crates/nym-gateway-directory/**"
       - "nym-vpn-core/crates/nym-vpnd/**"
```
octol, line 10: I wonder if we can skip this path trigger too, changes in the daemon should not affect the UI client at this point (exception being proto)

nym-vpn-x/src-tauri/src/fs/path.rs
octol: This is an interesting one. I think it touches on a broader thing where it would be pretty great if actually this can be swapped around. Bringing up the UI before reading config files and stuff like, so that we can present errors related to this in the UI. I saw the startup_error window and I think that's a good start, and we should capture all errors in the UI

That would be out of scope of this PR though.

pronebird: We spoke about that over the call. Basically if we move initialization into Tauri's `setup(|app| { ... })` then we can access paths directly from tauri's facilities. @doums said he will consider a refactor down the road.

doums: Yea actually, and thanks to Andrej for pointing it; there is a way to initiate those states in the tauri app setup. In the required non-optional way.
That would allow dropping that dedicated app 'startup_error' and handle all startup failure directly in the main app. But as you said, this is out of the scope here. But good to keep in mind in future iteration.

.github/workflows/ci-nymvpn-x-rust.yml
doums: yes and proto changes are included as trigger anyway, so useless imo
let's drop it

nym-vpn-x/src-tauri/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "nym-vpn-x"
+name = "nymvpn-x"
```
octol, line 2: So I see a lot of renaming nym-vpn-x -> nymvpn-x in the PR, not just here. If this is only done due to the Windows installer, then the correct solution would be to tweak the `installer.nsi` file which sets all these things

nym-vpn-x/src-tauri/src/tray.rs
pronebird: @doums because compiler will not warn you if tomorrow you or someone else remove or rename any of `MenuItemId` variants. If you were to match against `MenuItemId` then the compiler would error.

nym-vpn-x/src-tauri/Cargo.toml
octol, line 2: Looking at this and @pronebird pointing me to https://github.com/tauri-apps/tauri/issues/9668 I think we shouldn't rename the crate after all. Let's stick to the current one and try to sort out the windows installer in a follow up PR

octol (changes requested):

nym-vpn-x/src-tauri/Cargo.toml
doums: @octol the nsis installer template is not the only impacted area, please consider this:

>Jon sticking with `nym-vpn-x` will have these consequences:
- tauri will use and create `nym-vpn-x` as app dirs (eg config cache .local/state etc) instead of the current and expected `nymvpn-x`
- will need to code refactor to switch to `nym-vpn-x` and drop `nymvpn-x` 
- need to update linux installer scripts
-  need to update aur packages (maybe deb package ?),
- on new app version install, these dirs will be created along the old ones in the user file system
- nsis installer will have to be customized
- a lot of changes required in the gh build workflow as final emitted binaries changed from `nymvpn-x` to  `nym-vpn-x`
(+ I'm probably missing some other impacts)

Ofc in a next iteration, we could handle this global name changing, which is planned anyway

pronebird: Yeah I am not sure if hooks can be used for that, maybe there is another way to rename the installer package if that's the goal we have here. 

Unrelated to this PR: however I think we shouldn't copy `installer.nsh` unless we need heavy modifications to it because upstream fixes to this large file (1000 lines of code) might save us a lot of trouble. I know we will not be manually syncing it. Hooks can be used to add extra scripts if we need to.

octol: Are we sure about this, when I tried locally using the old crate name I saw it produced appimage and deb package with the old name nymvpn-x

doums: hmm weirdo, I experienced this on both linux and windows (that's why motivated me to rename rust package name)

doums: Just tried in dev mode with `nym-vpn-x`
image

nym-vpn-x/src-tauri/tauri.conf.json
```diff
@@ -1,84 +1,57 @@
...
-          "installerIcon": "icons/nsis_icon.ico",
-          "headerImage": "icons/nsis_header.bmp",
-          "sidebarImage": "icons/nsis_sidebar.bmp"
-        }
       }
     },
-    "security": {
-      "csp": null
-    },
+    "createUpdaterArtifacts": "v1Compatible"
```
octol, line 42: Do we need this right now? It fails the build for me locally

nym-vpn-x/src-tauri/Cargo.toml
doums: image

octol: where does the hsts-storage file come from?

nym-vpn-x/src-tauri/tauri.conf.json
doums: I wasn't sure, this has been set by the tauri auto migration command.
But I think we'll need to restore the updater stuff anyway in a next step.

nym-vpn-x/src-tauri/Cargo.toml
doums: Double-checked, running the generated appimage, it uses `.local/state/nym-vpn-x` as app data dir

octol: image

pronebird (changes requested):
I think that PR is really nice. I would only ask to use the constants when matching menu items please. We already have an enum type so it should be trivial to change.

nym-vpn-x/src-tauri/Cargo.toml
octol: that's what I get with `nym-vpn-x`

nym-vpn-x/src-tauri/tauri.conf.json
octol: ok, but let's remove it for now so it builds out of the box

nym-vpn-x/src-tauri/Cargo.toml
doums: yea bundle names respect `productName` but when you run them (the AppImage for instance), still uses `nym-vpn-x` as app dirs (cache config data and so on)

octol (approved):

pronebird (approved):
Outstanding!

PR #1099, opened 2024-09-10
Apple: Display notification on disconnect

rokas-ambrazevicius:
Display a notification on failed connection.

### DIFF
--- a/nym-vpn-apple/Services/Sources/Services/AppSettings/AppSettings.swift
+++ b/nym-vpn-apple/Services/Sources/Services/AppSettings/AppSettings.swift
@@ -43,10 +43,6 @@ public final class AppSettings: ObservableObject {
     public var connectionType: Int?
     @AppStorage(AppSettingKey.lastConnectionIntent.rawValue)
     public var lastConnectionIntent: String?
-    @AppStorage(AppSettingKey.didConnectSuccessfullyOnce.rawValue)
-    public var didConnectSuccessfullyOnce = false
-    @AppStorage(AppSettingKey.didAskForNotificationPermission.rawValue)
-    public var didAskForNotificationPermission = false
 
     @AppStorage(
         AppSettingKey.currentEnv.rawValue,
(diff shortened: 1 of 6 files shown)


pronebird:
I raise a discussion on this. I don't think we really need to do this.

nym-vpn-apple/Services/Sources/Services/NotificationsManager/NotificationsManager.swift
```diff
@@ -0,0 +1,95 @@
...
+    func requestNotificationPermission() {
+        UNUserNotificationCenter.current().requestAuthorization(
+            options: [.alert, .badge, .sound]
+        ) { [weak self] granted, _ in
+            self?.permissionGranted = granted
+        }
+    }
+
+    func askForPermissionIfNeeded() {
+        guard !appSettings.didAskForNotificationPermission else { return }
```
pronebird, line 89: Maybe we could rely on `authorizationStatus` to determine if we need to ask user for permission, for instance when it's only `.notDetermined`?

nym-vpn-apple/Services/Sources/Services/AppSettings/AppSettings.swift
```diff
@@ -43,12 +43,16 @@ public final class AppSettings: ObservableObject {
     public var connectionType: Int?
     @AppStorage(AppSettingKey.lastConnectionIntent.rawValue)
     public var lastConnectionIntent: String?
+    @AppStorage(AppSettingKey.didConnectSuccessfullyOnce.rawValue)
```
pronebird, line 46: This property is unused therefore can be removed.

pronebird (changes requested):

nym-vpn-apple/Services/Sources/Services/AppSettings/AppSettings.swift
```diff
@@ -78,4 +78,6 @@ enum AppSettingKey: String {
     case connectionType
     case lastConnectionIntent
     case currentEnv
+    case didConnectSuccessfullyOnce
```
pronebird, line 81: I don't think that neither `didConnectSuccessfullyOnce` nor `didAskForNotificationPermission` are used.

pronebird (changes requested):

pronebird (approved):

PR #1106, opened 2024-09-11
Stop route manager and remove routes

pronebird:
This is a band aid fix to remove routes and stop route manager when exiting main tunnel run loop.


octol (approved):

pronebird:
Closing because it looks like there is a problem with netlink on ubuntu24 and this PR does not fix it.

PR #1115, opened 2024-09-12
Use new vpn-api gateway directory

octol:
Use the new gateway directory endpoints in the nym-vpn-api

- [x] Add vpn-api client calls to fetch vpn gateways and countries
- [x] Remove calls to nym-api and polyfill step
- [x] Add vpn gateway list
- [x] Add vpn country list
- [x] Add to daemon API
- [x] Add to uniffi API

### DIFF
--- a/proto/nym/vpn.proto
+++ b/proto/nym/vpn.proto
@@ -391,6 +391,19 @@ message ListExitGatewaysResponse {
   repeated ExitGateway gateways = 1;
 }
 
+message VpnGateway {
+  Gateway id = 1;
+  Location location = 2;
+}
+
+message ListVpnGatewaysRequest {
+  Threshold min_gateway_performance = 1;
+}
+
+message ListVpnGatewaysResponse {
+  repeated VpnGateway gateways = 1;
+}
+
 message ListEntryCountriesRequest {
   Threshold min_gateway_performance = 1;
 }
@@ -404,6 +417,14 @@ message ListExitCountriesRequest {
 }
 
 message ListExitCountriesResponse {
+  repeated Location countries = 1;
+}
+
+message ListVpnCountriesRequest {
+  Threshold min_gateway_performance = 1;
+}
+
+message ListVpnCountriesResponse {
   repeated Location countries = 1;
 }
 
@@ -476,8 +497,10 @@ service NymVpnd {
 
   rpc ListEntryGateways (ListEntryGatewaysRequest) returns (ListEntryGatewaysResponse) {}
   rpc ListExitGateways (ListExitGatewaysRequest) returns (ListExitGatewaysResponse) {}
+  rpc ListVpnGateways (ListVpnGatewaysRequest) returns (ListVpnGatewaysResponse) {}
   rpc ListEntryCountries (ListEntryCountriesRequest) returns (ListEntryCountriesResponse) {}
   rpc ListExitCountries (ListExitCountriesRequest) returns (ListExitCountriesResponse) {}
+  rpc ListVpnCountries (ListVpnCountriesRequest) returns (ListVpnCountriesResponse) {}
 
   // Unstable
   rpc StoreAccount (StoreAccountRequest) returns (StoreAccountResponse) {}
--- a/.github/workflows/ci-nym-vpn-android.yml
+++ b/.github/workflows/ci-nym-vpn-android.yml
@@ -35,9 +35,6 @@ jobs:
       - name: Setup Android SDK
         uses: android-actions/setup-android@v3
 
-      - name: Install NDK
-        run: sdkmanager "ndk;26.3.11579264"
-
       - name: Run ktlint
         run: ./gradlew ktlintCheck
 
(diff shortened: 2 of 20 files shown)


proto/nym/vpn.proto
```diff
@@ -476,8 +497,10 @@ service NymVpnd {
 
   rpc ListEntryGateways (ListEntryGatewaysRequest) returns (ListEntryGatewaysResponse) {}
   rpc ListExitGateways (ListExitGatewaysRequest) returns (ListExitGatewaysResponse) {}
+  rpc ListVpnGateways (ListVpnGatewaysRequest) returns (ListVpnGatewaysResponse) {}
   rpc ListEntryCountries (ListEntryCountriesRequest) returns (ListEntryCountriesResponse) {}
   rpc ListExitCountries (ListExitCountriesRequest) returns (ListExitCountriesResponse) {}
+  rpc ListVpnCountries (ListVpnCountriesRequest) returns (ListVpnCountriesResponse) {}
```
doums, line 503: 1. How the country lists are supposed to be relative to (filtered by) the current selected VpnMode? As I don't see any changes in the request parameters, e.g. passing the target vpn mode as param
2. `ListVpnCountries` As we already have dedicated methods to get country list for both entry or exit, what is the purpose of this new method? Is it just a free addition to the service?

octol: These are new lists, and entry and exit lists for vpn are the same. The countries will not be the same for mixnet and vpn

doums: I'm not sure to understand it, so let-me know if not:
1. In 2hop-wg mode client is now expected to migrate on `ListVpnCountries` to query the list, it returns a list that is valid for both entry and exit gw
2.  In Mixnet mode, client is now expected to use `ListEntryCountries` and `ListExitCountries` to query the lists corresponding to entry and exit gw

Is that correct?

octol: that is correct

zaneschepke: Everything looks great! The only part I would want to potentially change is the naming. I don't think VpnCountries is very intuitive without an explanation. I think this should be something like WgCountries or TwoHopCountries to make it more clear what this is.

zaneschepke (approved):

proto/nym/vpn.proto
rokas-ambrazevicius: vpn countries kind of make sense, but then the other two are harder to understand without knowing context

rokas-ambrazevicius: The list countries requests look very similar. As well as responses.
Maybe it would be nice to have something like `countries(for type: CountryType)`?

```
    enum CountryType {
        MIXNET_ENTRY = 0;
        MIXNET_EXIT = 1;
        VPN = 2;
    }
```

zaneschepke: I think we could improve this in the future with something like what Rokas suggested and have a single uniffi function for getting countries, but for right now I think we can keep this as is and refactor later as it is a pain to regenerate bindings, etc.

doums: I agree with Rokas. Based on all 3 methods are sharing the same request and response types. We should strive to simplify that country API, so something like:

```proto
message ListCountriesRequest {
  enum Context {
    CONTEXT_UNSPECIFIED = 0;
    MIXNET_ENTRY = 1;
    MIXNET_EXIT = 2;
    TWO_HOP = 3;
  }
  Context context = 1;
  Threshold min_gateway_performance = 2;
}

message ListCountriesResponse {
  repeated Location countries = 1;
}

  // […]
  rpc ListCountries (ListCountriesRequest) returns (ListCountriesResponse) {}
```

But this includes some limitations:
- in the future we must be sure to not expect to have different needs by context for the request types and response types (e.g. for mixnet query we need to support a new request param that is only valid for mixnet mode)
- these new changes are breaking and not backward compatible (i.e. clients would have to migrate their grpc client implementation to be able to support the new daemon version)

rokas-ambrazevicius: I think it is better to have good foundation and improve the flow of introducing breaking things now, rather than start breaking them in production.
It is possible to make things backwards compatible.
I think we should be able to automatically update daemon without issues even if it is not backwards compatible.

rokas-ambrazevicius (approved):

proto/nym/vpn.proto
octol: Some comments here, the functions (both grpc and uniffi) where chosen to not break existing code and apps, so that we can merge this straightaway and not have to wait until all apps have integrated it. Once all apps have caught up we can unify and simplify by breaking compatibility in a single PR (for all apps). 

This is in particular relevant for the uniffi API which is currently a bit funky

octol: also, how the performance parameter will be used is still not really nailed down, so it's not guaranteed it will be the same for different modes. That's what I'm working on now for next week

doums (approved):

proto/nym/vpn.proto
zaneschepke: Dope! Sounds like a plan. Let's merge it!

octol:
Great feedback guys, let's try to incorporate it soon as possible. I'm gonna merge this as is, but let's unify and simplify the API in follow up PRs so that clients that are ready to, can that start using this straight away

PR #1125, opened 2024-09-14
pkg(vpn-x,aur): update nymvpn-x aur package

doums:


octol (approved):

octol (commented):
This reminds me, the deb file dependency on nym-vpnd should probably also have a version specified, so that we depend on >= 0.2.1

PR #1135, opened 2024-09-16
refactor(vpn-x): use fs rename

doums:


pronebird (approved):

nym-vpn-x/src-tauri/src/log.rs
```diff
@@ -44,6 +38,8 @@ pub async fn setup_tracing(log_file: bool) -> Result<Option<WorkerGuard>> {
         let log_dir = APP_LOG_DIR
             .clone()
             .ok_or(anyhow!("failed to get log dir"))?;
+        let log_file = log_dir.join(LOG_FILE);
+        println!("logging to file: {}", log_file.display());
```
octol, line 42: Can we not add this to the log instead of using `println`?  Log the location after the call to `tracing_subscriber::fmt()`

PR #1144, opened 2024-09-16
Finish support of the new gateway directory API

octol:
Finish support of the new gateway directory API on nymvpn.com.
Support setting the mixnet and vpn performance threshold parameters for the gateway directory.
Update grpc API.

The changes to the uniffi API is split out into https://github.com/nymtech/nym-vpn-client/pull/1155

**NOTE**: breaks the grpc API for getting gateways and countries

### DIFF
--- a/proto/nym/vpn.proto
+++ b/proto/nym/vpn.proto
@@ -95,7 +95,8 @@ message ConnectRequest {
   bool disable_background_cover_traffic = 7;
   bool enable_credentials_mode = 8;
   Threshold min_mixnode_performance = 9;
-  Threshold min_gateway_performance = 10;
+  Threshold min_gateway_mixnet_performance = 10;
+  Threshold min_gateway_vpn_performance = 11;
 }
 
 message ConnectResponse {
--- a/proto/nym/vpn.proto
+++ b/proto/nym/vpn.proto
@@ -354,9 +354,18 @@ message AsExit {
   bool can_route_ip_external_v6 = 5;
 }
 
+message WgProbeResult {
+  bool can_register = 1;
+  bool can_handshake = 2;
+  bool can_resolve_dns = 3;
+  float ping_hosts_performance = 4;
+  float ping_ips_performance = 5;
+}
+
 message ProbeOutcome {
   AsEntry as_entry = 1;
   AsExit as_exit = 2;
+  WgProbeResult wg = 3;
 }
 
 message Probe {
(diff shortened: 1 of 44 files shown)


proto/nym/vpn.proto
```diff
@@ -353,78 +354,55 @@ message AsExit {
...
+message ListGatewaysResponse {
+  repeated GatewayResponse gateways = 1;
 }
 
-message ListVpnCountriesRequest {
-  Threshold min_gateway_performance = 1;
+message ListCountriesRequest {
+  GatewayType kind = 1;
+  Threshold min_mixnet_performance = 2;
+  Threshold min_vpn_performance = 3;
```
doums, line 402: These two fields should be both marked as `optional`

proto/nym/vpn.proto
```diff
@@ -353,78 +354,55 @@ message AsExit {
...
+  EXIT = 2;
+  VPN = 3;
 }
 
-message ListExitGatewaysResponse {
-  repeated ExitGateway gateways = 1;
+message ListGatewaysRequest {
+  GatewayType kind = 1;
+  Threshold min_mixnet_performance = 2;
+  Threshold min_vpn_performance = 3;
```
doums, line 392: These two fields should be both marked as `optional`

proto/nym/vpn.proto
doums: It is an `Option` in rust implementations yes, but is it the case with other languages' implementation, like in swift?

proto/nym/vpn.proto
```diff
@@ -353,78 +355,57 @@ message AsExit {
...
   Probe last_probe = 3;
 }
 
-message ListExitGatewaysRequest {
-  Threshold min_gateway_performance = 1;
+enum GatewayType {
+  GATEWAY_TYPE_UNSPECIFIED = 0;
+  ENTRY = 1;
+  EXIT = 2;
+  VPN = 3;
```
doums, line 387: Suggestion:

```
MIXNET_ENTRY
MIXNET_EXIT
WG
```

doums (approved):

proto/nym/vpn.proto
octol: yes good suggestion

doums:
Changed the proto to match

```
MIXNET_ENTRY
MIXNET_EXIT
WG
```

zaneschepke (approved):

proto/nym/vpn.proto
doums: @rokas-ambrazevicius are proto non-scalar types generated in Swift code as "optional" or "nullable" without being labeled as `optional`?

PR #1148, opened 2024-09-16
APPLE: add daemon version check and update

rokas-ambrazevicius:
Add version check for daemon. If version is lower than required one - update daemon.

### DIFF
--- a/nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
+++ b/nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
@@ -335,8 +335,9 @@ private extension HomeViewModel {
 
     func installHelperIfNeeded() async -> Bool {
         var isInstalledAndRunning = helperManager.isHelperAuthorizedAndRunning()
+        let needsUpdate = await needsHelperUpdate()
         // TODO: check if possible to split is helper running vs isHelperAuthorized
-        guard isInstalledAndRunning
+        guard isInstalledAndRunning && !needsUpdate
         else {
             do {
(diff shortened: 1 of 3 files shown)


doums (approved):

nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
```diff
@@ -350,6 +351,16 @@ private extension HomeViewModel {
         return isInstalledAndRunning
     }
 
+    func needsHelperUpdate() async -> Bool {
+        do {
+            let version = try await grpcManager.version()
+            guard helperManager.requiredVersion != version else { return false }
```
pronebird, line 357: This guard seems unnecessary, could just `return helperManager.requiredVersion != version`

nym-vpn-apple/ServicesMacOS/Sources/GRPCManager/GRPCManager.swift
```diff
@@ -40,6 +40,24 @@ public final class GRPCManager: ObservableObject {
...
     }
 
+    // MARK: - Info -
+
+    public func version() async throws -> String {
+        try await withCheckedThrowingContinuation { continuation in
+            let call = client.info(Nym_Vpn_InfoRequest(), callOptions: CallOptions(logger: logger))
+
+            call.response.whenComplete { result in
+                switch result {
```
pronebird, line 50: You can return result directly with [`resume(with: Result<>)`](https://developer.apple.com/documentation/swift/checkedcontinuation/resume(with:)-3gh60)

PR #1149, opened 2024-09-16
refactor(desktop): rename to nym-vpn-app

doums:


octol (commented):
This is great!
Let me have a proper look through to see if there are any gotchas. Also, I'm thinking we can add a "conflicts" rule to the deb package with the old one so make the transition automatic and smooth for Ubuntu users. I can look into that

octol (approved):

PR #1156, opened 2024-09-17
APPLE: fix daemon not getting installed

rokas-ambrazevicius:


nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
```diff
@@ -352,6 +352,7 @@ private extension HomeViewModel {
     }
 
     func needsHelperUpdate() async -> Bool {
+        guard helperManager.isHelperAuthorizedAndRunning() else { return false }
```
pronebird, line 355: So what happens if helper tool is out of date but not running?

rokas-ambrazevicius: It will get reinstalled.

PR #1181, opened 2024-09-18
Handle another set of errors in the daemon

octol:
- **Handle OutOfBandwidth**
- **Handle a bunch of wg gateway client errors**


rokas-ambrazevicius (approved):

pronebird (commented):
> Reviewed 14 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181)* status: 14 of 16 files reviewed, 3 unresolved discussions (waiting on @doums, @neacsu, @octol, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-authenticator-client/src/error.rs` line 16 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79qZBtEjo2fR7vM9CI:-O79qZBtEjo2fR7vM9CJ:b-45v8m3) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/nym-vpn-core/crates/nym-authenticator-client/src/error.rs#L16)):*
> ```XML
> 
>     #[error("failed to send mixnet message")]
>     FailedToSendMixnetMessage(#[source] nym_sdk::Error),
> ```

Nit: feels excessive to include "failed" in the name, it's an error variant so it's a failure anyway, perhaps `SendMixnetMessage` would be shorter.
___
*[`proto/nym/vpn.proto` line 192 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79rNEt4UVipOSQUnZ6:-O79rNEt4UVipOSQUnZ7:b-npvtem) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/proto/nym/vpn.proto#L192)):*
> ```proto
> 
>   // Detailed message for logging and debugging
>   string internal = 2;
> ```

Does it mean that this field is only used to pass a custom message when kind == INTERNAL? Otherwise I think `message` was a good descriptive name for a string message that's not necessarily localized.
___
*[`proto/nym/vpn.proto` line 209 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79rsgx84dY8sjI5x_I:-O79rsgx84dY8sjI5x_J:bt2iwvk) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/proto/nym/vpn.proto#L209)):*
> ```proto
>     // An internal error that indicates a programmer error and should not
>     // happen. If it did an invariant was probably broken at some point.
>     INTERNAL = 33;
> ```

Nit: we might want to order errors by their corresponding raw value, right now it's mixed.

octol (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181)* status: 14 of 16 files reviewed, 3 unresolved discussions (waiting on @doums, @neacsu, @pronebird, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-authenticator-client/src/error.rs` line 16 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79qZBtEjo2fR7vM9CI:-O7AfB391TLbOynDdoV7:bykvmx6) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/nym-vpn-core/crates/nym-authenticator-client/src/error.rs#L16)):*

Nit: feels excessive to include "failed" in the name, it's an error variant so it's a failure anyway, perhaps `SendMixnetMessage` would be shorter.
</blockquote></details>

This is an interesting one. Clippy certainly has opinions about it when you end up having a "FailedTo" prefix to all enum cases. 
I think it often signals a subtle difference in that it emphasizes the attempted action that triggered the error, and not the error that occurred. It's a higher level of abstraction basically. That said, sometimes if the error case already is a verb, the "FailedTo" prefix becomes redundant
I'll remove it here
___
*[`proto/nym/vpn.proto` line 192 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79rNEt4UVipOSQUnZ6:-O7Ag_8R3bGXioPpq_VW:b-kny1ux) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/proto/nym/vpn.proto#L192)):*

Does it mean that this field is only used to pass a custom message when kind == INTERNAL? Otherwise I think `message` was a good descriptive name for a string message that's not necessarily localized.
</blockquote></details>

Nah it's always passed. Though what it contains varies a bit, but usually it's just the string representation of the top-level error that occurred, but not always. The reasoning for renaming it was to drive home to point that we should not be using this in the UI. If there is a more detailed description of the error that we can present in the UI (unlocalized), I think it should be included in `details`
___
*[`proto/nym/vpn.proto` line 209 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79rsgx84dY8sjI5x_I:-O7AhCZZ16dQ6xd7Veau:b-rfmqg6) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/proto/nym/vpn.proto#L209)):*

Nit: we might want to order errors by their corresponding raw value, right now it's mixed.
</blockquote></details>

Currently they are ordered by logical grouping, and loosely sorted by the order they can happen. Imho it makes sense to keep previous the numbers when we append additional error cases so that we don't invalidate any clients consuming this. Since the  numbers are checked at compile time anyway there is no risk of overlap and they can basically be random.

We could of course reset them periodically so they are sequential, but imho that would be better in a separate PR I think, possibly with consuming clients changed in the same PR

octol:
*[`proto/nym/vpn.proto` line 192 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1181#-O79rNEt4UVipOSQUnZ6:-O7AmJdB5NX4NtbBw9l9:b8dak7u) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2f479f469976d42dae559d06ccb2c616a667d53f/proto/nym/vpn.proto#L192)):*
<details><summary><i>Previously, octol (Jon Häggblad) wrote…</i></summary><blockquote>

Nah it's always passed. Though what it contains varies a bit, but usually it's just the string representation of the top-level error that occurred, but not always. The reasoning for renaming it was to drive home to point that we should not be using this in the UI. If there is a more detailed description of the error that we can present in the UI (unlocalized), I think it should be included in `details`
</blockquote></details>

ok I changed it back to `message`

PR #1186, opened 2024-09-20
Fix handling errors sent in channels

octol:
Primarily this PR fixes that `OutOfBandwidth` errors were being snuck through without being properly caught by the daemon.

Fix bandwidth messages sent from the mixnet client not being caught. This was caused by not matching on the correct type, due to importing the same crate twice to support the freepass credential handling until ecash is merged.

Also wrap the errors sent through the status channel in a struct suitable for catching by the listener.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1186)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5559,6 +5559,7 @@ dependencies = [
  "http 0.2.12",
  "maplit",
  "nym-bandwidth-controller 0.1.0 (git+https://github.com/nymtech/nym?rev=3f922cc0)",
+ "nym-bandwidth-controller 0.1.0 (git+https://github.com/nymtech/nym?rev=b7baff1)",
  "nym-bin-common 0.6.0 (git+https://github.com/nymtech/nym?rev=b7baff1)",
  "nym-task",
  "nym-validator-client 0.1.0 (git+https://github.com/nymtech/nym?rev=b7baff1)",
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5445,6 +5445,7 @@ dependencies = [
  "nym-authenticator-client",
  "nym-authenticator-requests",
  "nym-bandwidth-controller 0.1.0 (git+https://github.com/nymtech/nym?rev=3f922cc0)",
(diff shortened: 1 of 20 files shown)


rokas-ambrazevicius (approved):

doums (approved):

neacsu (approved):
> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1186)* status: all files reviewed, 1 unresolved discussion (waiting on @octol, @pronebird, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/status_listener.rs` line 88 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1186#-O7EF7qz9XkgQnkFpEPd:-O7EF7qz9XkgQnkFpEPe:b-ncj48n) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0c3cda9319efec740ebcaf0d6ba2c101b96905da/nym-vpn-core/crates/nym-vpnd/src/service/status_listener.rs#L88)):*
> ```XML
>         } else {
>             tracing::warn!("VPN status: unknown: {msg}");
>             tracing::warn!("{msg:#?}");
> ```

Although it doesn't happen just here, I think we should avoid doubling warning (or error) logs for the same thing, and even more one right after the other

octol (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1186)* status: all files reviewed, 1 unresolved discussion (waiting on @neacsu, @pronebird, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/status_listener.rs` line 88 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1186#-O7EF7qz9XkgQnkFpEPd:-O7ELJqQ8IUQhrUAFD04:buti79t) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0c3cda9319efec740ebcaf0d6ba2c101b96905da/nym-vpn-core/crates/nym-vpnd/src/service/status_listener.rs#L88)):*

Although it doesn't happen just here, I think we should avoid doubling warning (or error) logs for the same thing, and even more one right after the other
</blockquote></details>

ah good catch yeah I think that's leftover from testing, let me fix that

PR #1188, opened 2024-09-20
chore(tauri-app): handle more daemon errors

doums:
Add localization and expose latest addition to proto errors

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1188)


octol (approved):
Looking good. Just don't forget to also include the relevant details fields when diplaying in the UI when it's relevant (like which gateway it was). Often there is also a `reason` entry in the details map that while not localized, still will be informative under a expandable "details" field in the UI

doums:
> Looking good. Just don't forget to also include the relevant details fields when diplaying in the UI when it's relevant (like which gateway it was). Often there is also a `reason` entry in the details map that while not localized, still will be informative under a expandable "details" field in the UI

yes that's the current behavior and default, the `details` map is forwarded right to the UI (and displayed when present)

https://github.com/nymtech/nym-vpn-client/blob/a8a56b4d571f3fa9010dae4721c992667ebf42da/nym-vpn-app/src-tauri/src/error.rs#L120

https://github.com/nymtech/nym-vpn-client/blob/a8a56b4d571f3fa9010dae4721c992667ebf42da/nym-vpn-app/src/pages/home/ConnectionStatus.tsx#L66

PR #1189, opened 2024-09-20
fix: android mixnet regression

zaneschepke:
- add bypass for gateway_ws_fd
- add allowedIps calculator to exclude entry gateway IP from the tunnel
- fixes regressions where vpnservice does not shutdown properly

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1189)


pronebird (commented):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1189)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @zaneschepke)

nym-vpn-core/crates/nym-vpn-lib/src/vpn/mixnet.rs
```diff
@@ -141,6 +141,14 @@ impl NymVpn<MixnetVpn> {
             entry_mixnet_gateway_ip,
             default_lan_gateway_ip,
         );
+
+        #[cfg(target_os = "android")]
+        {
+            if let Some(raw_fd) = mixnet_client.gateway_ws_fd().await {
+                self.android_tun_provider.clone().bypass(raw_fd);
```
neacsu, line 148: From what I remember, this is something we were doing before too, in a similar place. I think the idea of @pronebird is to move this kind of platform specific code into the same place, as to be easier to maintain and more stable. Maybe also check with him if this is the appropriate place to call the `bypass` function

zaneschepke: Yeah, at some point I think it was accidentally removed. Yeah, I agree it should be moved in the future. I can check, but this is essentially the same spot we had this before.

nym-vpn-android/nym-vpn-client/src/main/java/net/nymtech/vpn/util/AllowedIpsCalculator.kt
```diff
@@ -0,0 +1,195 @@
+package net.nymtech.vpn.util
+
+import kotlin.math.floor
+import kotlin.math.ln
+import kotlin.math.pow
+
+// TODO this needs ipv6 support added
+object AllowedIpsCalculator {
```
octol, line 8: Would have thought something like this would already be implemented in libraries, or at least the building blocks to implement it without the detailed manual manipulations

octol (approved):

nym-vpn-android/nym-vpn-client/src/main/java/net/nymtech/vpn/util/AllowedIpsCalculator.kt
zaneschepke: Same! I was surprised to find there was nothing. I think I will eventually refactor it and publish it as a library. There was only a couple of repos that generate this allowedIps list (which I based this off of), and most are in python.

zaneschepke (commented):
> Reviewed 2 of 4 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1189)* status: all files reviewed, 2 unresolved discussions (waiting on @octol and @pronebird)

PR #1218, opened 2024-09-26
android: typed navigation migration

zaneschepke:
- Migrates android to jetpacks new typed navigation
- Fixes bug where user could have a selected country that is not part of the list of available countries
- Shut down VPN on out of bandwidth message

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1218)


rokas-ambrazevicius (approved):

nym-vpn-android/nym-vpn-client/src/main/java/net/nymtech/vpn/util/Constants.kt
```diff
@@ -6,7 +6,7 @@ object Constants {
 	const val NYM_VPN_LIB = "nym_vpn_lib"
 
 	// Add Rust environment vars for lib
-	const val DEFAULT_COUNTRY_ISO = "DE"
+	const val DEFAULT_COUNTRY_ISO = "FR"
```
rokas-ambrazevicius, line 9: I use Switzerland if it does not exist first country from list.

zaneschepke: I am using FR because I think it is the most consistent country where we have gateways across all of the environments.

PR #1236, opened 2024-09-30
feat(tauri-app): add user agent

doums:
Set the `user_agent` info for the relevant grpc calls

NOTE: since proto3 `optional` label is still prohibited in the codebase (for bad reasons imo) user_agent's git hash will default to empty string if there is no available hash (very unlikely)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1236)

### DIFF
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -80,7 +80,7 @@ impl GrpcClient {
             user_agent: UserAgent {
                 application: pkg.name.clone(),
                 version: pkg.version.to_string(),
-                platform: format!("{}-{}", OS, ARCH),
+                platform: format!("{}-{}-{}", OS, tauri_plugin_os::version(), ARCH),
                 git_commit,
             },
         };
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -80,7 +80,7 @@ impl GrpcClient {
             user_agent: UserAgent {
                 application: pkg.name.clone(),
                 version: pkg.version.to_string(),
-                platform: format!("{}-{}-{}", OS, tauri_plugin_os::version(), ARCH),
+                platform: format!("{}; {}; {}", OS, tauri_plugin_os::version(), ARCH),
                 git_commit,
             },
         };
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -171,6 +171,25 @@ impl GrpcClient {
         Ok(response.into_inner())
     }
 
+    /// Update `user_agent` with the daemon info
+    // TODO this is dirty, this logic shouldn't be handled in the client side
+    #[instrument(skip_all)]
+    pub async fn update_agent(&mut self) -> Result<(), VpndError> {
+        let mut vpnd = self.vpnd().await?;
+
+        let request = Request::new(InfoRequest {});
+        let response = vpnd.info(request).await.map_err(|e| {
+            error!("grpc info: {}", e);
+            VpndError::GrpcError(e)
+        })?;
+        let d_info = response.get_ref();
+        self.user_agent.version = format!("{} ({})", self.user_agent.version, d_info.version);
+        self.user_agent.git_commit =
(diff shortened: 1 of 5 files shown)


nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -59,13 +60,31 @@ pub enum VpndError {
...
+            .as_ref()
+            .and_then(|vc| vc.git())
+            .map(|g| g.commit_short_id.clone())
+            .unwrap_or_default();
+
+        let client = GrpcClient {
+            transport: Transport::from((config, cli)),
+            user_agent: UserAgent {
+                application: pkg.name.clone(),
+                version: pkg.version.to_string(),
```
octol, line 82: It's pretty useful to have the daemon version here as well in addition to the app version. See e.g how it's done for nym-vpnc. Same for the git commit, although it's probably less important in comparison

octol (changes requested):

nym-vpn-app/src-tauri/src/grpc/client.rs
doums: Checked the code of vpnc, at first glance it looks a bit dirty to have to call grpc `Info` method before any other method needing the user_agent, just in the purpose to assemble the version value.

Imo this logic should be handled in daemon side only, e.g the daemon mutates `user_agent.version` by appending its version to the value or we add a new field for it like `user_agent.daemon_version`.
As well this would remove the possibility for grpc clients to set 'fake' daemon version

nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -59,13 +60,31 @@ pub enum VpndError {
...
+                platform: format!("{}-{}", OS, ARCH),
```
octol, line 83: Please see the ticket and the surrounding discussion regarding the format. Also I don't think `OS` will give you specific enough info, what is the resulting string?

octol (changes requested):

nym-vpn-app/src-tauri/src/grpc/client.rs
octol: The proper solution imho would be to introduce a new type on the lib side with additional fields. But then this needs to be compatible with the existing UserAgent type since this is used in crates in the monorepo so probably need to create some shared trait in `nym-http-api-client`

But if we're not doing that, then the simpler and less error prone way I think is to set it all up front in once place rather than some field getting rewritten half way down the call stack

At the end of the day I think the first priority here is to just set them consistently according to the format in the ticket. Feel free to try to refactor things if you think you have the extra time

nym-vpn-app/src-tauri/src/grpc/client.rs
doums: Ah, yes the OS version would be useful

octol: Check the PR for nym-vpnc which contains an example for what it's set to. Note the field separators for the platform. There is a bit more material in the Jira ticket about the choice of format

```rust
UserAgent {
    application: "nym-vpnc",
    version: "0.2.2-dev (0.2.2-dev)",
    platform: "Ubuntu; Linux 23.10 Ubuntu; x86_64",
    git_commit: "6dde4bd8bda00cce89bc2a0c75dac267d3451882 (ba519d1cefb65d48fe6997ec303932fbb31fe4c6)",
}
```

nym-vpn-app/src-tauri/src/grpc/client.rs
doums: I think relying on clients to set user_agent daemon info is wrong.
Also with the current solution each client would have to carry on this extra logic (calling rpc `Info` to fill user agent with daemon info before calling any other rpc methods requiring that user agent).

>  I think is to set it all up front in once place rather than some field getting rewritten half way down the call stack

So yea this "once place" would be daemon instead of a piece of logic repeated for each client implementation (vpnc, vpn-cli, vpn-app)

Anyway, for the sake of 'moving on', and not falling into a consequent refactor, I implemented the logic in the tauri app as well https://github.com/nymtech/nym-vpn-client/pull/1236/commits/62d45f6bca6e328f5c815ee972fdbc57415cb713

zaneschepke (approved):

octol (approved):

PR #1301, opened 2024-10-14
Update to latest nym

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1301)


octol:
I think that failing test indicates that will not work for testnets where you need to be able to deserialize DescribedGateway?

neacsu:
> I think that failing test indicates that will not work for testnets where you need to be able to deserialize DescribedGateway?

Until the changes in `develop`  hit mainnet, I've pointed to an earlier, forked version of monorepo

octol (approved):

PR #1374, opened 2024-10-23
Extend API of account controller

octol:
- Extract storage to separate mod
- Fix bug waiting on polling tasks on exit
- Move credential storage to storage mod
- Add command to refresh account state
- Add command to reset device identity

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1374)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/ecash_client.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/ecash_client.rs
@@ -1,12 +1,17 @@
 // Copyright 2024 Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+use nym_credentials::{
+    AggregatedCoinIndicesSignatures, AggregatedExpirationDateSignatures, EpochVerificationKey,
+};
 use nym_http_api_client::{Client, HttpClientError, NO_PARAMS};
 use nym_validator_client::ecash::models::{
     AggregatedCoinIndicesSignatureResponse, AggregatedExpirationDateSignatureResponse,
     MasterVerificationKeyResponse,
 };
 use url::Url;
+
+use crate::error::Error;
 
 pub(crate) type VpnEcashApiClientError = HttpClientError;
 
@@ -23,7 +28,7 @@ impl VpnEcashApiClient {
         })
     }
 
-    pub(crate) async fn get_master_verification_key(
+    async fn _get_master_verification_key(
         &self,
     ) -> Result<MasterVerificationKeyResponse, VpnEcashApiClientError> {
         self.inner
@@ -31,7 +36,26 @@ impl VpnEcashApiClient {
             .await
     }
 
-    pub(crate) async fn get_aggregated_coin_indices_signatures(
+    pub(crate) async fn get_master_verification_key(&self) -> Result<EpochVerificationKey, Error> {
+        let master_verification_key = self
+            ._get_master_verification_key()
+            .await
+            .map_err(Error::from)?
+            .key;
+
+        // Temporary workaround
+        let current_epoch = self
+            ._get_aggregated_coin_indices_signatures()
+            .await?
+            .epoch_id;
+
+        Ok(EpochVerificationKey {
+            epoch_id: current_epoch,
+            key: master_verification_key,
+        })
+    }
+
+    pub(crate) async fn _get_aggregated_coin_indices_signatures(
         &self,
     ) -> Result<AggregatedCoinIndicesSignatureResponse, VpnEcashApiClientError> {
         self.inner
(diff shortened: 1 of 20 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/ecash_client.rs
```diff
@@ -23,15 +28,34 @@ impl VpnEcashApiClient {
...
+    pub(crate) async fn get_master_verification_key(&self) -> Result<EpochVerificationKey, Error> {
+        let master_verification_key = self
+            ._get_master_verification_key()
+            .await
+            .map_err(Error::from)?
+            .key;
+
+        // Temporary workaround
+        let current_epoch = self
+            ._get_aggregated_coin_indices_signatures()
```
jstuczyn, line 48: that seems quite nasty as coin indices are the biggest (size wise)

jstuczyn: can't you perhaps combine queries for the verification key with coin index signatures? because you only need to get both of them once, when DKG epoch changes (i.e. at the same time)

nym-vpn-core/crates/nym-vpnc/src/main.rs
```diff
@@ -188,6 +189,14 @@ async fn store_account(client_type: ClientType, store_args: &cli::StoreAccountAr
     Ok(())
 }
 
+async fn refresh_account_state(client_type: ClientType) -> Result<()> {
+    let mut client = vpnd_client::get_client(client_type).await?;
+    let request = tonic::Request::new(RefreshAccountStateRequest {});
+    let response = client.refresh_account_state(request).await?.into_inner();
+    println!("{:#?}", response);
```
jstuczyn, line 196: I think that println! is some leftover from your debugging?

nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub(crate) async fn check_local_remaining_tickets(&self) -> Vec<(TicketType, u32)> {
+        // TODO: remove unwrap
+        let ticketbooks_info = self.storage.get_ticketbooks_info().await.unwrap();
+
+        // For each ticketbook type, iterate over and check if we have enough tickets stored
+        // locally
+        let ticketbook_types = ticketbook_types();
+
+        let mut request_zk_nym = Vec::new();
+        for ticketbook_type in ticketbook_types.iter() {
```
jstuczyn, line 94: wouldn't it make more sense to return a `HashMap` with an entry per `TicketType` in case you have multiple ticketbooks for the same underlying ticketbook type?

nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+pub(crate) struct VpnCredentialStorage {
+    pub(crate) storage: nym_credential_storage::persistent_storage::PersistentStorage,
+}
+
+impl VpnCredentialStorage {
+    pub(crate) async fn check_local_remaining_tickets(&self) -> Vec<(TicketType, u32)> {
+        // TODO: remove unwrap
+        let ticketbooks_info = self.storage.get_ticketbooks_info().await.unwrap();
```
jstuczyn, line 87: sanity check: does it filter out expired tickets?

jstuczyn (approved):

nym-vpn-core/crates/nym-vpn-account-controller/src/ecash_client.rs
octol: This stuff will change once we the credential-proxy is integrated with the nym-vpn-api. Then I wont need to reach out directly to any ecash endpoints on nym-api

But yes, the idea is to only download them once an epoch if possible

nym-vpn-core/crates/nym-vpnc/src/main.rs
octol: This is the cli client, and it's pretty rudimentary at the moment so that part is intended :)

nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs
octol: They are summed per ticketbook type. Problem with HashMap is the requirment on deriving Hash, which I encountered. But let me double check

nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs
octol: Good point, I'll look into it!

PR #1376, opened 2024-10-24
build(tauri-linux): use latest official protoc

doums:
should fix https://github.com/nymtech/nym-vpn-client/actions/runs/11492450452/job/31986537918

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1376)

### DIFF
--- a/.github/workflows/build-nym-vpn-app-linux.yml
+++ b/.github/workflows/build-nym-vpn-app-linux.yml
@@ -64,6 +64,7 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
         with:
+          version: "21.12" # version on ubuntu 24.04
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Get package version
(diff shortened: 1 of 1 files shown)


.github/workflows/build-nym-vpn-app-linux.yml
```diff
@@ -55,6 +55,11 @@ jobs:
         with:
           node-version: 20
 
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
```
octol, line 59: Imho let's stick to the version that core uses

octol (changes requested):

.github/workflows/build-nym-vpn-app-linux.yml
doums: ok please push to the pr the version you want

doums: as side note, using latest is building -> https://github.com/nymtech/nym-vpn-client/actions/runs/11495747973

octol (approved):

PR #1399, opened 2024-10-26
Gracefully shutdown task manager

pronebird:
This PR introduces the cancellation token to `tunnel_state_machine::tunnel::connect_mixnet()` and ensures that the `TaskManager` is properly shutdown before the drop.

In more details:

- `CancellationToken::run_until_cancelled()` is used for gateway selection to bail as soon as cancellation arrives. The assumption that it's safe to do so with the gateway client. I am not all that familiar with it but I assume any network requests it performs at best will be cancelled on drop, and at worst ignored. I assume that those requests are non-mutating and side-effect free.
- `TaskManager` is created right after the gateways are selected. At this point we run a combination of `CancellationToken::run_until_cancelled(tokio::time::timeout(<Future>))`. There are no early returns here to ensure that we get the opportunity to shutdown the `TaskManager` before the drop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1399)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
@@ -6,7 +6,6 @@ use std::{path::PathBuf, result::Result};
 use nym_config::defaults::NymNetworkDetails;
 use nym_sdk::mixnet::{MixnetClientBuilder, NodeIdentity, StoragePaths};
 use nym_vpn_store::mnemonic::MnemonicStorage as _;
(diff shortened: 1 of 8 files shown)


octol (approved):
> Reviewed 5 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1399)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pronebird)

PR #1414, opened 2024-10-28
feat(tauri): dynamic urls based on network env

doums:
- adds auto update of nymtech.net urls when daemon network changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1414)


rokas-ambrazevicius (approved):

octol:
The idea is that all that info will be queried from the vpnlib/daemon. Especially those URL might change

doums:
yea temp solution while we're iterating under canary/qa etc

PR #1422, opened 2024-10-29
wg: Capture IPv6 traffic

pronebird:
- Configure wireguard tun adapters with unique local addresses (ULA)
- Enable IPv6 routing
- Wireguard configuration is set to: `allowed_ips=0.0.0.0/0, ::0/0` but we do not list the ipv6 addresses set on tun interfaces which makes me wonder if wg will simply drop all ipv6 as it would not be able to route it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1422)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connecting_state.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connecting_state.rs
@@ -52,19 +52,18 @@ const DEFAULT_TUN_MTU: u16 = if cfg!(any(target_os = "ios", target_os = "android
 } else {
     1500
 };
-
-#[cfg(any(target_os = "linux", target_os = "macos"))]
-/// Entry IPv6 address (ULA) used by WireGuard, currently not routable.
-const WG_ENTRY_IPV6_ADDR: Ipv6Addr = Ipv6Addr::new(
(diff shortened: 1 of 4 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connecting_state.rs
```diff
@@ -3,6 +3,8 @@
 
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 use std::net::Ipv4Addr;
+#[cfg(unix)]
```
neacsu, line 6: Why not use the same cfg for both IP types?

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connecting_state.rs
```diff
@@ -247,7 +258,7 @@ impl ConnectingState {
             tracing::debug!("Created tun device: {}", tun_name);
 
             let routing_config = RoutingConfig::Mixnet {
-                enable_ipv6,
+                enable_ipv6: true,
```
neacsu, line 261: You can probably remove this field altogether

neacsu (approved):

PR #1432, opened 2024-10-30
Wireguard Build Script Update

jmwample:
Modify the script used to build the wrapped wireguard go library to use variables and `getopt`.

On mac (and mac-adjacent platforms) `getopt` has different behavior which doesn't handle long options - one simple answer is to use the `gnu-getopt` package which I have added to the mac CI build script

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)

### DIFF
--- a/.github/workflows/ci-nym-vpn-core-ios.yml
+++ b/.github/workflows/ci-nym-vpn-core-ios.yml
@@ -46,9 +46,6 @@ jobs:
           version: "21.12" # 3.21.12: the version on ubuntu 24.04. Don't change this!
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install script dependencies
-        run: brew install gnu-getopt
-
       - name: Build wireguard
         shell: bash
         run: |
--- a/.github/workflows/ci-nym-vpn-core-ios.yml
+++ b/.github/workflows/ci-nym-vpn-core-ios.yml
@@ -46,6 +46,9 @@ jobs:
           version: "21.12" # 3.21.12: the version on ubuntu 24.04. Don't change this!
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Install script dependencies
+        run: brew install gnu-getopt
+
       - name: Build wireguard
         shell: bash
         run: |
--- a/.github/workflows/ci-nym-vpn-core-macos.yml
+++ b/.github/workflows/ci-nym-vpn-core-macos.yml
@@ -24,7 +24,7 @@ jobs:
           rm -rf ./* || true
           rm -rf ./.??* || true
           ls -la ./
-      
+
       - name: Checkout repo
         uses: actions/checkout@v4
 
@@ -44,6 +44,9 @@ jobs:
         with:
           version: "21.12" # 3.21.12: the version on ubuntu 24.04. Don't change this!
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install script dependencies
+        run: brew install gnu-getopt
 
       - name: Build wireguard
         shell: bash
--- a/.github/workflows/ci-nym-vpn-core-macos.yml
+++ b/.github/workflows/ci-nym-vpn-core-macos.yml
@@ -45,9 +45,6 @@ jobs:
           version: "21.12" # 3.21.12: the version on ubuntu 24.04. Don't change this!
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install script dependencies
-        run: brew install gnu-getopt
-
       - name: Build wireguard
         shell: bash
         run: |
--- a/wireguard/build-wireguard-go.sh
+++ b/wireguard/build-wireguard-go.sh
@@ -2,39 +2,70 @@
 
 # This script is used to build wireguard-go libraries for all the platforms.
 
-set -eu
+function stringContain {
+    case $2 in *$1* ) return 0;; *) return 1;; esac;
+}
+
+ANDROID_BUILD=false
+IOS_BUILD=false
+DOCKER_BUILD=true
+
+function parseArgs {
+    if stringContain "Darwin" "$(uname -s)"; then
+        # Mac builds require gnu-getopt because regular macos getopt doesn't allow long args. -_-
+        # This could be avoided using something like `getopts` instead, but then we don't
+        # have the ability to use long options which pre-date this script change.
+        # > brew install gnu-getopt
+        echo "using gnu-getopt"
+        export PATH="/opt/homebrew/opt/gnu-getopt/bin:$PATH"
+    fi
+
+    which getopt
+    TEMP=$(getopt -o ai --long android,no-docker,ios \
+                  -n 'build-wireguard-go.sh' -- "$@")
+
+    if [ $? != 0 ]; then
+        echo "encountered an error parsing args"
+        exit 2
+    fi
+
+    # Note the quotes around '$TEMP': they are essential!
+    eval set -- "$TEMP"
+
+    while true; do
+      case "$1" in
+        "-a" | "--android" ) ANDROID_BUILD=true; shift ;;
+        "-i" | "--ios" ) IOS_BUILD=true; shift ;;
+        "--no-docker" ) DOCKER_BUILD=false; shift ;;
+        -- ) shift; break ;;
+        * ) break ;;
+      esac
+    done
+
+    echo "android:$ANDROID_BUILD ios:$IOS_BUILD docker:$DOCKER_BUILD"
+
+    set -eu
+}
 
 function is_android_build {
-    for arg in "$@"
-    do
-        case "$arg" in
-            "--android")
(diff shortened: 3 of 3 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jmwample)

___
*[`wireguard/build-wireguard-go.sh` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWuMCkEjKn8UIvtMJP:-OAWuMCkEjKn8UIvtMJQ:b-j3qym6) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L35)):*
> ```Shell
>     eval set -- "$TEMP"
> 
>     while true; do
> ```

Do you think we can use a regular for loop to avoid introducing a dependency? Seems simple enough. 

```sh
for arg in "$@"
do
  case "$arg" in
    "--ios")
      IOS_BUILD=true
    "--android")
     ANDROID_BUILD=true
   "--no-docker")
     DOCKER_BUILD=false
   *) echo "Unsupported argument: $arg" && exit 2 ;;
  esac
```

Also arguments like `-a` and `-i` are not very descriptive. Since this script is meant to be run by build scripts for the most part, I think we better stick to spelling out the long names for clarity of course unless you see some practical use for this.
___
*[`wireguard/build-wireguard-go.sh` line 47 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWuUiJCLbk0CX9H6i0:-OAWuUiJCLbk0CX9H6i1:bm1k2bv) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L47)):*
> ```Shell
>     echo "android:$ANDROID_BUILD ios:$IOS_BUILD docker:$DOCKER_BUILD"
> 
>     set -eu
> ```

Shouldn't this be set at the top of file?
___
*[`wireguard/build-wireguard-go.sh` line 299 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWsZSK-CnslAQZaRva:-OAWsZSK-CnslAQZaRvb:bwjk25z) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L299)):*
> ```Shell
> }
> 
> LIB_DIR="libwg"
> ```

I think we should move globals to the top of this file. This shouldn't affect the functions above right?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jmwample)

___
*[`wireguard/build-wireguard-go.sh` line 51 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWvhsABrZPefofR56J:-OAWvhsABrZPefofR56K:b-6vn80z) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L51)):*
> ```Shell
> 
> function is_android_build {
>     if [ "$ANDROID_BUILD" = true ]; then
> ```

FYI we could as well remove all these `is_xxx` functions and use variables directly where needed. They seem to be merely the remains of old way of parsing args. But you don't have to do this in this PR if you don't feel like doing so.

jmwample (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @pronebird)

___
*[`wireguard/build-wireguard-go.sh` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWuMCkEjKn8UIvtMJP:-OAYkVWF8_yObamWP3v_:bnsg606) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L35)):*

Do you think we can use a regular for loop to avoid introducing a dependency? Seems simple enough. 

```sh
for arg in "$@"
do
  case "$arg" in
    "--ios")
      IOS_BUILD=true
    "--android")
     ANDROID_BUILD=true
   "--no-docker")
     DOCKER_BUILD=false
   *) echo "Unsupported argument: $arg" && exit 2 ;;
  esac
```

Also arguments like `-a` and `-i` are not very descriptive. Since this script is meant to be run by build scripts for the most part, I think we better stick to spelling out the long names for clarity of course unless you see some practical use for this.
</blockquote></details>

Yeah, we could do that. I was having issues with adding arguments and then having the library end up is a strange triplet path because we pass `$@` to some functions to use as args as well. Using a real tool intended for option parsing made it easier to guarantee that things that were options were treated as such. However, for some reason mac hates long options so there is no default installed tool to handle them. 

Short options made it easier when I was running the script by hand a large number of times, but you are right they are not necessary. 

All this to say that this process was more thought than ever probably needed to go into this script and string matching seems fine.
___
*[`wireguard/build-wireguard-go.sh` line 47 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWuUiJCLbk0CX9H6i0:-OAYkQM1BvPFMlLKxV4o:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L47)):*

Shouldn't this be set at the top of file?
</blockquote></details>

Done.
___
*[`wireguard/build-wireguard-go.sh` line 299 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWsZSK-CnslAQZaRva:-OAYkSid33DZHZEhcGph:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L299)):*

I think we should move globals to the top of this file. This shouldn't affect the functions above right?
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: all files reviewed, 2 unresolved discussions (waiting on @jmwample)

___
*[`wireguard/build-wireguard-go.sh` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAWuMCkEjKn8UIvtMJP:-OAavgPs4rQOdCsWZyG0:bpgyn8r) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/2d2fa855cb3918560c828507730afacd975c1eae/wireguard/build-wireguard-go.sh#L35)):*
> Yeah, we could do that. I was having issues with adding arguments and then having the library end up is a strange triplet path because we pass `$@` to some functions to use as args as well. 

The whole thing with passing "$@" was just a legacy code and we kept following down the same path up until now. Good riddance and I really like your improvements.

> Using a real tool intended for option parsing made it easier to guarantee that things that were options were treated as such. However, for some reason mac hates long options so there is no default installed tool to handle them.

I am being conservative because less prerequisites we have the better.

> Short options made it easier when I was running the script by hand a large number of times, but you are right they are not necessary.

Makes sense. If all you do is rebuild the same script then arrow up should pick up the last command from history, that's what I do. But that's of course not the most optimal way if you do a lot of things in terminal in between the rebuilds.

> All this to say that this process was more thought than ever probably needed to go into this script and string matching seems fine.

I see. Let's keep it simple. :+1: 
___
*[`wireguard/build-wireguard-go.sh` line 37 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432#-OAawO5W2pFPK3z25ezE:-OAawO5W2pFPK3z25ezF:b-93rhp4) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/763ea0706d86942b6cd87b1f97ff6d25d170d88c/wireguard/build-wireguard-go.sh#L37)):*
> ```Shell
> }
> 
> function stringContain {
> ```

Nit: unused function

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: all files reviewed, 1 unresolved discussion (waiting on @jmwample)

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1432)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jmwample)

pronebird:
Approved. I took a liberty of merging this.

PR #1435, opened 2024-10-31
mobile: fix account api

zaneschepke:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1435)


nym-vpn-core/crates/nym-vpn-lib/src/platform/error.rs
```diff
@@ -21,41 +21,8 @@ pub enum VpnError {
...
-                Self::NoActiveSubscription
-            }
-            nym_vpn_account_controller::ReadyToConnect::DeviceNotRegistered => {
-                Self::DeviceNotRegistered
-            }
-            nym_vpn_account_controller::ReadyToConnect::DeviceNotActive => Self::DeviceNotActive,
-        }
-    }
+    #[error("{details}")]
+    Account { details: String },
```
octol, line 25: This is unfortunate, but oh well you gotta do what you gotta do. Maybe we can return an integer instead so that at least you can localize the response?

octol (approved):

PR #1436, opened 2024-10-31
bugfix: filter out entry-capable nodes from the described endpoint

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1436)

### DIFF
--- a/nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
+++ b/nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
@@ -277,6 +277,7 @@ impl GatewayClient {
                 .collect();
             Ok(GatewayList::new(gateways))
         } else {
+            warn!("OPERATING IN FALLBACK MODE WITHOUT NYM-VPN-API!");
             self.lookup_all_gateways_from_nym_api().await
         }
     }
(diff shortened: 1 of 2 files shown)


octol (approved):
Note though this should be unused for fully deployed networks

nym-vpn-core/crates/nym-gateway-directory/src/gateway_client.rs
```diff
@@ -202,9 +202,10 @@ impl GatewayClient {
 
     pub async fn lookup_all_gateways_from_nym_api(&self) -> Result<GatewayList> {
         let mut gateways = self
-            .lookup_described_gateways()
+            .lookup_described_nodes()
             .await?
             .into_iter()
+            .filter(|node| node.description.declared_role.entry)
```
neacsu, line 208: Exit no?

jstuczyn: you tell me what's the expected behaviour. the old endpoints returned all gateways, i.e. both entry and exit, so I tried to replicate the behaviour

neacsu: Yes, that's what I was thinking, why isn't it entry | exit, instead of just entry?

jstuczyn: because this is not the role that has been assigned to the node in the contract (this would be the next step), but rather what this node declares it can operate as. and currently exit implies entry, i.e. you must declare you can operate as an entry if you declare yourself exit-capable

octol (approved):
Works on my machine!

PR #1502, opened 2024-11-08
chore(tauri): use standard path for grpc socket

doums:
use /run instead of /var/run symlink

### DIFF
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -32,7 +32,7 @@ use crate::{events::AppHandleEventEmitter, states::SharedAppState};
 
 const VPND_SERVICE: &str = "nym.vpn.NymVpnd";
 #[cfg(not(windows))]
-const DEFAULT_SOCKET_PATH: &str = "/var/run/nym-vpn.sock";
+const DEFAULT_SOCKET_PATH: &str = "/run/nym-vpn.sock";
 #[cfg(windows)]
 const DEFAULT_SOCKET_PATH: &str = r"\\.\pipe\nym-vpn";
 const DEFAULT_HTTP_ENDPOINT: &str = "http://[::1]:53181";
--- a/nym-vpn-app/src/screens/location/NodeLocation.tsx
+++ b/nym-vpn-app/src/screens/location/NodeLocation.tsx
@@ -129,7 +129,7 @@ function NodeLocation({ node }: { node: NodeHop }) {
 
   const renderError = (e: AppError) => (
     <div className="w-4/5 h-2/3 overflow-auto break-words text-center">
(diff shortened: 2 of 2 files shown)


octol (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1502)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pronebird, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-app/src-tauri/src/grpc/client.rs` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1502#-OBAjy6h-BFoHJD3EHxb:-OBAlsRb-LOt_3En6AL1:b73zu73) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8514bae53069b9857fbb66a569aabda36a082dff/nym-vpn-app/src-tauri/src/grpc/client.rs#L35)):*

Seems like a significant change and not a "chore". Not sure how important but `/run` is absent on macos and we might not be running a tauri app on mac anymore but that also means that we can't develop it using macos. Is this what we want?
</blockquote></details>

Also IMHO the "correct" solution here is that things like this is configured in a common crate shared between the daemon and apps, to avoid hardcoding things

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1502)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @octol, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-app/src-tauri/src/grpc/client.rs` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1502#-OBAjy6h-BFoHJD3EHxb:-OBApebz-jHUKeDyF5GO:bqffalq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8514bae53069b9857fbb66a569aabda36a082dff/nym-vpn-app/src-tauri/src/grpc/client.rs#L35)):*
<details><summary><i>Previously, octol (Jon Häggblad) wrote…</i></summary><blockquote>

Also IMHO the "correct" solution here is that things like this is configured in a common crate shared between the daemon and apps, to avoid hardcoding things
</blockquote></details>

This would be out of scope for this PR but certainly moving rpc into individual crate that can be consumed by the daemon and clients is the ideal solution.

pronebird (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1502)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @octol, @rokas-ambrazevicius, and @zaneschepke)

doums:
@pronebird I reverted to `/var/run` for apple

PR #1565, opened 2024-11-14
[Product Data] Reports exit country and tickets

simonwicky:
# What it does
Add reporting of exit country and tickets used. 

This PR requires https://github.com/nymtech/nym/pull/5134 to be merged first

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1565)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mixnet/connector.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mixnet/connector.rs
@@ -6,6 +6,7 @@ use std::net::IpAddr;
 use nym_gateway_directory::{GatewayClient, IpPacketRouterAddress, Recipient};
 use nym_ip_packet_client::IprClientConnect;
 use nym_ip_packet_requests::IpPair;
+use nym_sdk::mixnet::ConnectionStatsEvent;
 use nym_task::TaskManager;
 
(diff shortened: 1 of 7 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mixnet/connector.rs
```diff
@@ -72,6 +73,13 @@ impl Connector {
             exit_mix_addresses,
             interface_addresses,
         };
+        if let Some(exit_country_code) = selected_gateways.exit.two_letter_iso_country_code() {
+            self.mixnet_client
+                .send_stats_event(
+                    ConnectionStatsEvent::MixCountry(exit_country_code.to_string()).into(),
```
neacsu, line 79: Doesn't this heavily deanonymises the mixnet mode of the vpn, by creating a link between mixnet client's identity and the exit gateway? At least in the current state, where SURBs are not implemented for mixnet mode.

simonwicky: So this info gets sent to the aggregator and the report including this is sent through the mixer, with a stats id, hashed from the identity. 
You're indeed reducing the set of gateway you might be connected to, but you're not doxing yourself by revealing your entry gateway

octol (approved):

PR #1566, opened 2024-11-14
chore(tauri): bump dev versions

doums:
- bump dev version to `1.0.0-dev`
- cherry-picked gh workflow changes from `release/2024.1`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1566)

### DIFF
--- a/nym-vpn-app/package.json
+++ b/nym-vpn-app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "nym-vpn-app",
   "private": true,
-  "version": "1.0.0-dev",
+  "version": "1.0.1-dev",
   "type": "module",
   "authors": [
     "pierre <dommerc.pierre@gmail.com>",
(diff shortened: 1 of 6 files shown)


nym-vpn-app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "nym-vpn-app",
   "private": true,
-  "version": "0.2.4-dev",
+  "version": "1.0.0-dev",
```
octol, line 4: This branch `main` is targetting the next release after the one in the release branch, so convention is that it has a newer version `>1.0.0`. So something like `1.1.0-dev` or `1.0.1-dev`

doums: make sense if `1.0.0` has been released, which is not the case

zaneschepke (approved):

PR #1576, opened 2024-11-16
Report no_serving on shutdown in vpnd

octol:
It seems the grpc listener blocks as long as someone is listening on the health service. This prevents graceful shutdown.

This PR makes sure to report `not_serving` on the health service at shutdown. This does not fix the problem, but at least the listener can choose to disconnect and unblock the shutdown

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1576)


doums (approved):

doums (commented):
I dont think it's good idea to rely on the client to close the stream connection of `Healt::watch`, when the health reports `not_serving`

PR #1581, opened 2024-11-18
Return partial set of account links when no account id available

octol:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1581)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
@@ -257,12 +257,12 @@ async fn fetch_account_links(
     network_name: &str,
     locale: &str,
 ) -> Result<AccountLinks, VpnError> {
-    let account_id = account::get_account_id(path).await?;
+    let account_id = account::get_account_id(path).await.ok();
     nym_vpn_network_config::Network::fetch(network_name)
         .and_then(|network| {
             network
                 .nym_vpn_network
-                .try_into_parsed_links(locale, &account_id)
+                .try_into_parsed_links(locale, account_id.as_ref().map(|s| s.as_str()))
         })
         .map(AccountLinks::from)
         .map_err(|err| VpnError::InternalError {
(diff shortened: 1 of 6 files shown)


dynco-nym (approved):

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
```diff
@@ -948,15 +948,15 @@ where
         &self,
         locale: String,
     ) -> Result<ParsedAccountLinks, AccountError> {
-        let account = self.load_account().await?;
-        let account_id = account.id();
+        let account = self.load_account().await.ok();
```
jstuczyn, line 951: nit: if account failed to get loaded, it might be useful to log the error in case it's something different from it missing. (maybe even match on that error to make sure its nothing critical)

nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
```diff
@@ -257,12 +257,12 @@ async fn fetch_account_links(
     network_name: &str,
     locale: &str,
 ) -> Result<AccountLinks, VpnError> {
-    let account_id = account::get_account_id(path).await?;
+    let account_id = account::get_account_id(path).await.ok();
```
jstuczyn, line 260: same comment as in `nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` regarding possible error matching

jstuczyn (approved):

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
octol: So this is an interesting one. Basically it's not an error if the account isn't there, it just means it was called when we are not logged in yet. I toyed with the idea to have separate API for the two use-cases, but that would mean yet another breaking change so I left it as is, at least for now

PR #1625, opened 2024-11-20
Change forget account functionality to only delete files explicitly

octol:


nym-vpn-core/crates/nym-vpn-account-controller/src/util.rs
```diff
@@ -0,0 +1,44 @@
...
+    // credentials_database.db
+    // *.sqlite
+    // *.pem
+
+    for file in fs::read_dir(data_dir).map_err(Error::internal)? {
+        let file = match file {
+            Ok(file) => file,
+            Err(err) => {
+                tracing::error!("failed to read file in data directory: {:?}", err);
+                break;
```
doums, line 22: `break` will abort dir file iteration
I think it's better to just ignore any no matching files (like directory) and continue to iterate through

octol: ah yes good catch, I mean `continue`

doums: also can change the `error` lvl to `warn` imo

nym-vpn-core/crates/nym-vpn-account-controller/src/util.rs
```diff
@@ -0,0 +1,44 @@
...
+        let file_name = match file.file_name().to_str() {
+            Some(file_name) => file_name.to_owned(),
+            None => {
+                tracing::error!("failed to convert file name to string: {:?}", file);
+                break;
+            }
+        };
+        if file_name == "credentials_database.db"
+            || file_name.ends_with(".sqlite")
+            || file_name.ends_with(".pem")
```
doums, line 35: suggestion: `Path::extension`

PR #1630, opened 2024-11-21
Improve usability of nym-vpnc

octol:
- Document vpnc commands
- Move internal commands not intended for end-users to subcommand
- Pretty print grpc responses
- Log number of gateways found before erroring out
- Listen until connected/disconnected

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1630)

### DIFF
--- a/nym-vpn-core/crates/nym-vpnc/src/main.rs
+++ b/nym-vpn-core/crates/nym-vpnc/src/main.rs
@@ -33,6 +33,11 @@ use crate::{
     },
 };
 
+struct CliOptions {
+    client_type: ClientType,
+    verbose: bool,
+}
+
 #[tokio::main]
 async fn main() -> Result<()> {
     let args = cli::CliArgs::parse();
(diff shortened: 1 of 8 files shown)


nym-vpn-core/crates/nym-vpnc/src/main.rs
```diff
@@ -141,31 +159,113 @@ async fn connect(client_type: ClientType, connect_args: &cli::ConnectArgs) -> Re
     });
 
     let response = client.vpn_connect(request).await?.into_inner();
-    println!("{:#?}", response);
+
+    if opts.verbose {
+        println!("{:#?}", response);
```
jstuczyn, line 164: should all of those responses be printed with `println` or rather `tracing::...`?

jstuczyn (approved):

nym-vpn-core/crates/nym-vpnc/src/main.rs
octol: Currently there is no logger setup at all. Honestly I'm not sure. Usually you don't have a logger for interactive CLI apps? Or should we?

PR #1659, opened 2024-11-26
Handle zknym ticketbooks in account controller

octol:
- Create a command handler for requesting zknym ticketbooks. This means it will not block other account controller activities, and will enable the same style command query as the other commands (e.g register device), and make it possible to integrate with apps
- Similar to how registering devices is handled, add commands to ensure that tickets are available, either by already having them or going and fetching new ones. 
- Update both daemon and uniffi API

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659)


jstuczyn (changes requested):
apart from few nitpicks, looks good to me, ship it!

> Reviewed 43 of 43 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-:-ODLKLPn-CCMjnUt-cre:bj9zd2)* status: all files reviewed, 17 unresolved discussions (waiting on @doums, @neacsu, @octol, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs` line 116 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLJIfw5HJvUOnhQaFE:-ODLJIfw5HJvUOnhQaFF:bpkuvft) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs#L116)):*
> ```XML
>             .get_available_tickets()
>             .await?
>             .is_all_ticket_types_above_threshold(0)
> ```

do we need **all** tickets above 0? 

consider the following cases:
1. somebody say is only using wireguard (so doesn't care about mixnet) - I reckon it will be quite a lot of users
2. somebody connects to a gateway where they have already claimed a lot of bandwidth. in that case they don't really need to have fresh tickets in the storage

but i guess thats an optimisation for the future
___
*[`nym-vpn-core/crates/nym-vpnc/src/main.rs` line 195 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODL88QPAG1xgWusbRA7:-ODL88QPAG1xgWusbRA8:bub0e3w) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpnc/src/main.rs#L195)):*
> ```XML
>     println!("Connect command failed: {} (id={kind:?})", error.message);
>     for zk_nym_error in error.zk_nym_error {
>         let zk_nym_error = zk_nym_error.clone();
> ```

nit: do we have to clone it just to print its details to stdout? (unless .message() consumes it)
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs` line 116 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLJsi-AAqSsJMwP_Pc:-ODLJsi-AAqSsJMwP_Pd:byngarp) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs#L116)):*
> ```XML
>             .await
>             .map_err(Error::SetupCredentialStorage)?;
>         let storage = Arc::new(tokio::sync::Mutex::new(storage));
> ```

does the storage need to be behind a mutex? sql should be able to handle concurrent uses, no?
___
*[`nym-vpn-core/crates/nym-vpn-lib/src/platform/error.rs` line 86 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODL7Ti_9eeONQBpe0xh:-ODL7Ti_9eeONQBpe0xi:b-74ff0m) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-lib/src/platform/error.rs#L86)):*
> ```XML
> 
> #[derive(uniffi::Record, Clone, Debug, PartialEq, Eq)]
> pub struct RequestZkNymSuccess {
> ```

nit: should a "Success" struct live inside an error module : )
___
*[`proto/nym/vpn.proto` line 777 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODL8wCs8lzlahR51m15:-ODL8wCs8lzlahR51m16:b-cnvs8o) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/proto/nym/vpn.proto#L777)):*
> ```proto
>   optional MnemonicState mnemonic = 1;
>   optional AccountRegistered account_registered = 2;
>   optional AccountSummary account_summary = 3;
> ```

does protobuf allow renaming fields without breaking compatibility?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/ticketbooks.rs` line 16 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLK9DZ3BZEELc_1Ncd:-ODLK9DZ3BZEELc_1Nce:bo8mi1y) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/ticketbooks.rs#L16)):*
> ```XML
> // TODO: I picked a random number, check what is should be. Or if we can express this in terms of
> // data/bandwidth.
> const TICKET_NUMBER_THRESHOLD: u64 = 30;
> ```

I thnk 30 tickets is a rather high threshold, given it's 3/5 of a ticketbook. so we haven't even used half of it
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 36 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAT_MCs2ts6XOau1I:-ODLAT_MCs2ts6XOau1J:b4tlwu9) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L36)):*
> ```XML
> // The maximum number of zk-nym requests that can fail in a row before we disable background
> // refresh
> const ZK_NYM_MAX_FAILS: u32 = 10;
> ```

sanity check: we do reset the counter on success, right?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 42 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAHjQ8ZKcHjF1q178:-ODLAHjR9Ij2j74hhr7O:b-kadtgy) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L42)):*
> ```XML
>     account_state: SharedAccountState,
>     vpn_api_client: VpnApiClient,
>     zk_nym_fails_in_a_row: Arc<tokio::sync::Mutex<u32>>,
> ```

a Mutex behind u32 is an overkill. you can use `AtomicU32` instead (and then you wouldn't need to have all the async overhead)
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 95 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAccCCpkyH7JxxUPm:-ODLAccCCpkyH7JxxUPn:brrpl08) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L95)):*
> ```XML
> impl RequestZkNymCommandHandler {
>     fn id_str(&self) -> String {
>         format!("{:.8}", self.id.to_string())
> ```

ah, that's the pattern you've told me about : )
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 155 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAySQDkzZk6HCm7cR:-ODLAySQDkzZk6HCm7cS:b-lwz2ob) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L155)):*
> ```XML
>     async fn request_zk_nym_inner(
>         &mut self,
>         ticket_types: Vec<TicketType>,
> ```

do we have to ensure those types are unique?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 176 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLB2Y3DXfyNtFqyynl:-ODLB2Y3DXfyNtFqyynm:b-hm444q) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L176)):*
> ```XML
>                 async move { request_zk_nym(request, &account, &device, &vpn_api_client).await }
>             })
>             .buffer_unordered(4)
> ```

perhaps make the buffer be the size of the number of requested tickets?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 229 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLBRxT7U_dKik07vvu:-ODLBRxT7U_dKik07vvv:b7ksjga) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L229)):*
> ```XML
>                         .map(|e| {
>                             tracing::warn!(
>                                 "nym-vpn-api reports: message={}, message_id={:?}, code_reference_id={:?}",
> ```

you can add those extra variables to tracing natively, i.e. `warn!(message = e.message, message_id=%e.message_id, "nym-vpn-api reports")`
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 287 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLC-3q6bglSaIJnV3L:-ODLC-3q6bglSaIJnV3M:b-9i6nmy) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L287)):*
> ```XML
>                             Ok(_) => {
>                                 tracing::info!("Successfully imported zk-nym: {}", id);
>                                 *self.zk_nym_fails_in_a_row.lock().await = 0;
> ```

should this be reset to zero at this point or only after the full request (i.e. all tickets being requested in the request) succeeded? say you requested 4 ticketbooks and only 1 worked. would it make sense to set the counter back to 0?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 343 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLCI-WAI8h3ANJ-Qrq:-ODLCI-WAI8h3ANJ-Qrr:b-q9mv5m) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L343)):*
> ```XML
>         let issuers = self
>             .vpn_api_client
>             .get_directory_zk_nyms_ticketbookt_partial_verification_keys()
> ```

ugh. do we query for the keys for EVERY response? could we at least maybe cache then in some internal RwLock<HashMap<...>> ? so if the entry exists, we use them, otherwise we query
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 353 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLC_Yz1QQioo8luqef:-ODLC_Yz1QQioo8luqeg:b-r43yhi) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L353)):*
> ```XML
>         tracing::info!("epoch_id: {}", shares.epoch_id);
> 
>         let master_vk_bs58 = shares
> ```

I know, I know it's "temporary", but it pains me to see this extra bloat with every response : (
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 504 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLCtxM0_L8UxrMCi3X:-ODLCtxM0_L8UxrMCi3Y:bfcg480) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L504)):*
> ```XML
>     let start_time = Instant::now();
>     loop {
>         tokio::time::sleep(std::time::Duration::from_secs(5)).await;
> ```

5s is a rather slow polling time. In best case scenario I assume each request would take around 5-6s to resolve, so you'd end up having to wait 10s until you get your zk-nym. perhaps lower it to like 3-4s?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 523 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLD6qjCE0SmBz8Uxxk:-ODLD6qk5p5jP0lnc86_:bhze9nf) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L523)):*
> ```XML
>             Ok(poll_response) => {
>                 tracing::info!("zk-nym polling not finished: {:#?}", poll_response);
>                 if start_time.elapsed() > Duration::from_secs(60) {
> ```

we should have some mechanism to save the timed out request to try to retrieve it at later date (maybe on restart?) because unless credential proxy explicitly abandoned it, it should be avaialble eventually

octol (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659)* status: all files reviewed, 15 unresolved discussions (waiting on @doums, @jstuczyn, @neacsu, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs` line 116 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLJIfw5HJvUOnhQaFE:-ODLlB529mNqDKmdKcMv:bekv64j) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs#L116)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

do we need **all** tickets above 0? 

consider the following cases:
1. somebody say is only using wireguard (so doesn't care about mixnet) - I reckon it will be quite a lot of users
2. somebody connects to a gateway where they have already claimed a lot of bandwidth. in that case they don't really need to have fresh tickets in the storage

but i guess thats an optimisation for the future
</blockquote></details>

Yes. It's also complicated by the fact that you're supposed to be able to switch between mixnet and wireguard. At least in the future. So for now we do the possibly wasteful simpler option

And also you're right, if we need to request tickets should depend on the gateway connecting to

I think both of these will be simpler to address once the account controller is more integrated into the state machine and we can do ticket requests as part of the connecting step, and not as currently, before attempting to connect
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs` line 116 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLJsi-AAqSsJMwP_Pc:-ODLPQ4SCXl-x-SRqiMv:b-iqcmuq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/storage.rs#L116)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

does the storage need to be behind a mutex? sql should be able to handle concurrent uses, no?
</blockquote></details>

Doesn't need to be, I think it was mainly about implementation ergonomics. So I can pass it around easier. It's not read/write to very frequently anyway. And I'm afraid about windows and locked files so trying to touch the sqlite files from as few places as possible
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/ticketbooks.rs` line 16 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLK9DZ3BZEELc_1Ncd:-ODLPpGrA6FWUgLrTBT-:b-vywrlk) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/ticketbooks.rs#L16)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

I thnk 30 tickets is a rather high threshold, given it's 3/5 of a ticketbook. so we haven't even used half of it
</blockquote></details>

What it's being a bit defensive on is that it might take some time to successfully download more tickets, especially if it happens over the mixnet. But also I don't know. What threshold would you suggest?
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 36 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAT_MCs2ts6XOau1I:-ODLQGPi5Nht1xnuYbAU:b-y083xz) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L36)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

sanity check: we do reset the counter on success, right?
</blockquote></details>

We do!
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 42 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAHjQ8ZKcHjF1q178:-ODLljSl040y9ptnF2Le:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L42)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

a Mutex behind u32 is an overkill. you can use `AtomicU32` instead (and then you wouldn't need to have all the async overhead)
</blockquote></details>

Done.
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 95 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAccCCpkyH7JxxUPm:-ODLllN82jCoNKz0645H:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L95)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

ah, that's the pattern you've told me about : )
</blockquote></details>

Done.
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 155 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLAySQDkzZk6HCm7cR:-ODLTLVDDG49DjNlaHz6:b-x1z0tw) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L155)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

do we have to ensure those types are unique?
</blockquote></details>

Ideally we should, but I don't think it's critical. And I think in this case it was because TicketType doesn't implement `Hash`
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 176 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLB2Y3DXfyNtFqyynl:-ODLlmpB2KIXN7pWXrYU:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L176)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

perhaps make the buffer be the size of the number of requested tickets?
</blockquote></details>

Done.
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 229 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLBRxT7U_dKik07vvu:-ODLlncj0X31oA8xiuQK:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L229)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

you can add those extra variables to tracing natively, i.e. `warn!(message = e.message, message_id=%e.message_id, "nym-vpn-api reports")`
</blockquote></details>

Done.
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 287 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLC-3q6bglSaIJnV3L:-ODLWULy5nI6CF6l1TUH:b-xdunqq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L287)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

should this be reset to zero at this point or only after the full request (i.e. all tickets being requested in the request) succeeded? say you requested 4 ticketbooks and only 1 worked. would it make sense to set the counter back to 0?
</blockquote></details>

Well, I thought about it. And I think if 1 worked that's an indication to keep trying. So that's why it's per request and not per task
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 343 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLCI-WAI8h3ANJ-Qrq:-ODLxR8r8YhtSopwZWtI:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L343)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

ugh. do we query for the keys for EVERY response? could we at least maybe cache then in some internal RwLock<HashMap<...>> ? so if the entry exists, we use them, otherwise we query
</blockquote></details>

Done.
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 353 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLC_Yz1QQioo8luqef:-ODM0yK_2oBXOOe0n9ZO:b-syk6q5) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L353)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

I know, I know it's "temporary", but it pains me to see this extra bloat with every response : (
</blockquote></details>

Yes, hopefully we can add some parameters to those calls to tell if we need to include them in the response
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 504 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLCtxM0_L8UxrMCi3X:-ODLWws67GXASEqD879U:b-hmd10t) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L504)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

5s is a rather slow polling time. In best case scenario I assume each request would take around 5-6s to resolve, so you'd end up having to wait 10s until you get your zk-nym. perhaps lower it to like 3-4s?
</blockquote></details>

The limitation is that these can take a long time to complete when run over the mixnet, like >10 sec sometimes
___
*[`nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs` line 523 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODLD6qjCE0SmBz8Uxxk:-ODLXKeODkci7OBkYni8:bdzm403) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/request_zknym.rs#L523)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

we should have some mechanism to save the timed out request to try to retrieve it at later date (maybe on restart?) because unless credential proxy explicitly abandoned it, it should be avaialble eventually
</blockquote></details>

Yep. This also requires to persist the request info needed to handle the response
___
*[`nym-vpn-core/crates/nym-vpn-lib/src/platform/error.rs` line 86 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODL7Ti_9eeONQBpe0xh:-ODLXTru0NCmK2OvEjU2:bvh5igh) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/nym-vpn-core/crates/nym-vpn-lib/src/platform/error.rs#L86)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

nit: should a "Success" struct live inside an error module : )
</blockquote></details>

Well, it's a partial failure / partial success ;)
___
*[`proto/nym/vpn.proto` line 777 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-ODL8wCs8lzlahR51m15:-ODM27WMEkpCeDwcxC8j:bksuy7a) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0ff1058331e3a33fd8564a2021df10c1c78099ac/proto/nym/vpn.proto#L777)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

does protobuf allow renaming fields without breaking compatibility?
</blockquote></details>

No it doesn't but, not sure why I only partially cared for keeping compatibility. But now I don't want to change it since people have started integrating against the API.... hmm

jstuczyn (commented):
lgtm, ship it!

> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-:-ODMhJIT7NJuoypFitmw:bkddzkv)* status: all files reviewed, 7 unresolved discussions (waiting on @doums, @neacsu, @octol, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

jstuczyn (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1659#-:-ODMhNBVBezERpYl1_9n:bnfp4nl)* status: all files reviewed, 7 unresolved discussions (waiting on @doums, @neacsu, @octol, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

PR #1663, opened 2024-11-26
Enable cover traffic for mobile

neacsu:
In order to not kill the whole tunnel on mobile platforms like Android because the mixnet channel auto-shuts down when the tcp connection is unused, we do the same thing as on mixnet mode i.e. enable cover traffic, which basically keeps the websocket active permanently. This is a temporary solution, until we have a better way to treat the machine going to sleep.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
@@ -1,7 +1,7 @@
 // Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::{path::PathBuf, result::Result};
+use std::{path::PathBuf, result::Result, time::Duration};
 
 use nym_config::defaults::NymNetworkDetails;
 use nym_sdk::mixnet::{MixnetClientBuilder, NodeIdentity, StoragePaths};
@@ -9,6 +9,8 @@ use nym_vpn_store::mnemonic::MnemonicStorage as _;
 
 use super::{MixnetError, SharedMixnetClient};
 use crate::{storage::VpnClientOnDiskStorage, MixnetClientConfig};
+
+const MOBILE_LOOP_COVER_STREAM_AVERAGE_DELAY: Duration = Duration::from_secs(10);
 
 #[allow(unused)]
 fn true_to_enabled(val: bool) -> &'static str {
@@ -76,8 +78,15 @@ pub(crate) async fn setup_mixnet_client(
     mut task_client: nym_task::TaskClient,
     mixnet_client_config: MixnetClientConfig,
     enable_credentials_mode: bool,
+    two_hop_mode: bool,
 ) -> Result<SharedMixnetClient, MixnetError> {
     let mut debug_config = nym_client_core::config::DebugConfig::default();
+    // for mobile platforms, in two hop mode, we do less frequent cover traffic,
+    // to preserve battery
+    if two_hop_mode && (cfg!(target_os = "android") || cfg!(target_os = "ios")) {
+        debug_config.cover_traffic.loop_cover_traffic_average_delay =
+            MOBILE_LOOP_COVER_STREAM_AVERAGE_DELAY;
+    }
     apply_mixnet_client_config(&mixnet_client_config, &mut debug_config);
 
     let user_agent = nym_bin_common::bin_info_owned!().into();
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs
@@ -159,11 +159,7 @@ pub async fn connect_mixnet(
             // Always disable poisson process for outbound traffic in wireguard.
             mixnet_client_config.disable_poisson_rate = true;
             // Always disable background cover traffic in wireguard, except for android
-            if cfg!(target_os = "android") {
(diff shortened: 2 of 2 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @doums, @neacsu, @octol, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs` line 161 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663#-OCdhSAMDZZZNQaNjejx:-OCdhSAMDZZZNQaNjejy:blhrunt) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/3a70b3dd57514ced1e58340ec487fe81003b2053/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs#L161)):*
> ```XML
>             // Always disable poisson process for outbound traffic in wireguard.
>             mixnet_client_config.disable_poisson_rate = true;
>             // Always disable background cover traffic in wireguard.
> ```

Can you keep it but only for desktop platforms macos/windows/linux. and set it to false for android only?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @doums, @neacsu, @octol, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs` line 161 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663#-OCdhSAMDZZZNQaNjejx:-OCdhmSs3bfqBbtEVrU5:brc1beo) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/3a70b3dd57514ced1e58340ec487fe81003b2053/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs#L161)):*

Can you keep it but only for desktop platforms macos/windows/linux. and set it to false for android only?
</blockquote></details>

oh: keep it true on ios because i don't think it matters for ios. It will sleep regardless.

octol (approved):
Consider lowering the rate from the default 20 sphinx packets/sec if it saves battery

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @doums, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs` line 161 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663#-OCdhSAMDZZZNQaNjejx:-OCh3EEaEo8OIpTOLOt0:b-gwyj2w) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/3a70b3dd57514ced1e58340ec487fe81003b2053/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/mod.rs#L161)):*

oh: keep it true on ios because i don't think it matters for ios. It will sleep regardless.
</blockquote></details>

For desktop platforms it is already disabled, when in two hop mode (https://github.com/nymtech/nym-vpn-client/blob/d3bafc5198761a12be6d7293b00a406995922e1a/nym-vpn-core/crates/nym-vpnd/src/command_interface/listener.rs#L930)
I've pushed a change to keep it disabled only for iOS, and on Android it will get enabled by default.

neacsu (commented):
Changed it to 1 packet/10 seconds

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1663#-:-OChA2IK6JXr5jYI6Mbz:bpa069h)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @doums, @pronebird, @rokas-ambrazevicius, and @zaneschepke)

nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
```diff
@@ -76,8 +78,15 @@ pub(crate) async fn setup_mixnet_client(
     mut task_client: nym_task::TaskClient,
     mixnet_client_config: MixnetClientConfig,
     enable_credentials_mode: bool,
+    two_hop_mode: bool,
 ) -> Result<SharedMixnetClient, MixnetError> {
     let mut debug_config = nym_client_core::config::DebugConfig::default();
+    // for mobile platforms, in two hop mode, we do less frequent cover traffic,
+    // to preserve battery
+    if two_hop_mode && (cfg!(target_os = "android") || cfg!(target_os = "ios")) {
```
pronebird, line 86: Nit: this will have no effect on ios as we disable cover traffic in there anyway. The thing is, ios tunnel doesn't care about cover traffic, it will suspend entirely whenever it wants to.

zaneschepke (approved):

nym-vpn-core/crates/nym-vpn-lib/src/mixnet/connect.rs
neacsu: Yep, I know, it's just redundant code, just in case at some point it gets enabled, we have rare keep-alives there as well

pronebird (approved):

PR #1676, opened 2024-11-28
Support for wireguard-go on windows

pronebird:
This PR introduces initial support for wireguard-go on Windows:

- Fixes crash in libwg related to C->Go->C->Go roundtrip
- Creates wintun adapter using Go and later configures it using syscalls

Main differences:

- Unlike other platforms where we pass tun fd, on windows wireguard-go creates a wintun interface itself and configures MTU on it. We set the IP configuration using winapi syscalls using a network adapter id (luid) returned by wireguard-go.
- Since wireguard-go creates the wintun interface, we cannot set routing before starting the tunnel so the order in which system is being configured is distinct from other platforms.
- Due to the fact that wintun device is owned by wireguard-go, since we drop the tunnel before resetting the routing table, the error is printed in console saying that the routing manager couldn't clear all of the routes. I'll address that later.
- I have ideas how to refactor tunnel monitor because it has grown substantially, but not in this PR. We're really rushing to have something working asap.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1676)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -4868,6 +4868,7 @@ dependencies = [
  "nym-vpn-store",
  "nym-wg-gateway-client",
  "nym-wg-go",
+ "nym-windows",
  "nym-wireguard-types",
  "oslog",
  "pnet_packet",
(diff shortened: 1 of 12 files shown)


octol:
Great work! :)

IMHO I think we should re-point the PR to `develop`

octol (approved):
Great work, as always!

PR #1678, opened 2024-11-28
rip windows

doums:
Unreliable windows runner -> no windows release

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1678)


rokas-ambrazevicius (approved):

octol (commented):
I don't think this will work, since the file needs to also exist on the main branch (`develop`) for github to recognize it.
Better to just edit the existing one

PR #1683, opened 2024-11-29
Fix if condition in uniffi upload step

pronebird:
There is no such step as `diff_ios` so the if condition `steps.diff_ios.outcome` is no longer valid. Let's make sure that that the upload only happens if "Generate Uniffi" step succeeds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1683)

### DIFF
--- a/.github/workflows/ci-nym-vpn-core-ios.yml
+++ b/.github/workflows/ci-nym-vpn-core-ios.yml
@@ -71,7 +71,6 @@ jobs:
             --language swift --out-dir build -n
 
       - name: Upload generated uniffi file
-        if: steps.diff_ios.outcome == 'failure'
         uses: actions/upload-artifact@v4
         with:
           name: updated-uniffi-ios


.github/workflows/ci-nym-vpn-core-ios.yml
```diff
@@ -71,7 +72,7 @@ jobs:
             --language swift --out-dir build -n
 
       - name: Upload generated uniffi file
-        if: steps.diff_ios.outcome == 'failure'
+        if: steps.generate_uniffi.outcome == 'success'
```
octol: Conditional not needed. If the above fails this wont run anyway

pronebird: Right. Thanks. Simply removed the bogus if condition

octol (approved):

PR #1685, opened 2024-11-29
Simon/bugfix wait for connected

simonwicky:
- `listen_until_connected` is now `listen_until_connected_or_failed` and additionally returns upon reaching the `ConnectionFailed` state.
- `-w` to wait until connection is established or not. (Disabled by default, do we want it the other way around?)
- `listen_until_connected_or_failed` returns with an error, such that `nym-vpnc connect` has an error code of `1` if the connection failed

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1685)


pronebird (approved):

octol (approved):

pronebird:
Thank you for disabling wait by default.

PR #1689, opened 2024-11-29
Cross Compile targetting windows

jmwample:
PR adds support for cross compiling to target windows from unix (linux) envs. 

This came about while debugging windows builds in general, as cross compiling is easier for me than compiling in windows native. I figured I would make a PR in case this is useful to anyone else.

Some details: 
* using MSVC from non-Windows environments is a big hassle, so instead we target `x86_64-pc-windows-gnu`
* `x86_64-pc-windows-gnu` tools can be installed with `sudo apt-get install binutils-mingw-w64 mingw-w64` or your platforms equivalent.
* ARM64 cannot be built this way as `mingw-w64` doesn't yet have support for `aarch64-w64-mingw32-gcc`
  * There is probably a more involved way to cross compile targetting windows arm, but I am going to leave it as "unsupported" for now

:warning: I have not tested the produced binaries yet as I don't have access to a windows machine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1689)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,14 @@ build-wireguard-ios:
 build-nym-vpn-core:
 	$(MAKE) -C nym-vpn-core build
 
+# Used for cross compiling to target windows from linux
+# requires:
+#     binutils-mingw-w64 mingw-w64
+windows-cross: build-wireguard-windows build-nym-vpn-core-windows
+
+build-wireguard-windows:
+	./wireguard/build-wireguard-go.sh --windows-cross
+
+build-nym-vpn-core-windows:
+	$(MAKE) -C nym-vpn-core build-win-cross
+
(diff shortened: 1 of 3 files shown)


octol (commented):
Oh this is pretty cool :)

pronebird (approved):
Thanks for bringing this, this should also work for macos. Let's merge! 🎉

pronebird:
For macOS:

```sh
brew install mingw-w64
rustup target add x86_64-pc-windows-gnu
```

Configure linker in `.cargo/config.toml`:

```toml
[target.x86_64-pc-windows-gnu]
linker = "x86_64-w64-mingw32-gcc"
```

Then

```sh
cargo build --target=x86_64-pc-windows-gnu -p nym-vpn-lib
```

PR #1719, opened 2024-12-04
win: bump wg socket on route change

pronebird:
- Add default route monitoring on windows and rebind tunnel socket on route change.
- Introduce `Tombstone` struct holding the remains of tunnel device or wireguard-go tunnels. 
  This struct unifies the return type of a tunnel handle at the end of its lifecycle.
  _On Windows_ where wireguard-go tunnels create and own the wintun interface, dropping the tunnel also drops the wintun adapter, which interferes with the routing table. Instead the tunnels that own the tunnel interface are kept alive, for a very brief moment, until the routing table and dns are fully reset.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1719)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5076,6 +5076,7 @@ dependencies = [
  "base64 0.22.1",
  "hex",
  "ipnetwork",
+ "nym-windows",
  "thiserror",
  "tracing",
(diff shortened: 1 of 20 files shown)


doums:
> Introduce Tombstone struct

good naming

octol (approved):

PR #1771, opened 2024-12-10
Remove dependency on ctr::cipher::zeroize

pronebird:
Use our own `zeroize` instead of the one re-exported from `ctr::cipher::zeroize`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1771)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -4799,6 +4799,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "tracing",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


octol (approved):
Can this maybe lead to possibly multiple versions of zeroize crate being used? And in some corner cases cause issuing resolving mutual dependencies?

jstuczyn (approved):

pronebird:
@octol I thought that Rust didn't allow multiple versions of the same dependency in a workspace. Same goes for features, that accumulate. Isn't that so?

octol:
> @octol I thought that Rust didn't allow multiple versions of the same dependency in a workspace. Same goes for features, that accumulate. Isn't that so?

As an example we have multiple versions of reqwest in the dependency tree (according to cargo tree)

pronebird:
@octol 

> > @octol I thought that Rust didn't allow multiple versions of the same dependency in a workspace. Same goes for features, that accumulate. Isn't that so?
> 
> As an example we have multiple versions of reqwest in the dependency tree (according to cargo tree)

Yeah that's true. I'd need to learn a bit more to understand the exact behaviour. But conflict shouldn't be a problem in this case since `Zeroizing` is used locally and not passed between crates.

PR #1811, opened 2024-12-17
Add reproducible builds for rust codebase

pronebird:
This PR ensures that our release builds are reproducible by using the following set of techniques:

- Set `VERGEN_IDEMPOTENT` env variable. The following values are set to default and remain the same across rebuilds:
  - `VERGEN_BUILD_DATE` set to default
  - `VERGEN_BUILD_TIMESTAMP` set to default
  - `VERGEN_GIT_COMMIT_DATE` set to default
  - `VERGEN_GIT_COMMIT_TIMESTAMP` set to default
  Note that the warning is emitted to console during such builds.

- Set `VERGEN_GIT_BRANCH` to default because some CIs build from detached HEAD and others from branch etc and that results into different string being embedded into the binary.

- Set rust compiler flags to replace common paths in panic traces:
  - `$HOME` dir is redacted
  - Local filesystem project root is redacted
  - Rust compiler sysroot (typically `~/.rustup/toolchains/<toolchain>`) is redacted
  After running `strings nym-vpnd` a few times I couldn't find any other identifying strings so I assume these three should be sufficient for now but we should of course ensure that it all works properly on Windows too.

- Linker/compiler flags for reproducible builds are only passed in release builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1811)

### DIFF
--- a/nym-vpn-core/Makefile
+++ b/nym-vpn-core/Makefile
@@ -1,6 +1,9 @@
 # Detect the OS and architecture
 include platform.mk
 
+# Linker/compiler flags for producing reproducible output
+include reproducible_builds.mk
+
 # Minimum deployment targets for macOS and iOS
 # These environment variables are used by clang
 export MACOSX_DEPLOYMENT_TARGET = 10.13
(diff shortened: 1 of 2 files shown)


octol (approved):
Nice :)

simonwicky (approved):

nym-vpn-core/Makefile
```diff
@@ -6,6 +6,27 @@ include platform.mk
...
 export IPHONEOS_DEPLOYMENT_TARGET = 16.0
 
+# Project root replaced with /buildroot in reproducible builds
+PROJECT_ROOT=$(shell dirname $(CURDIR))
+
+# Rust compiler sysroot that typically points to ~/.rustup/toolchains/<toolchain>
+RUST_COMPILER_SYS_ROOT=$(shell rustc --print sysroot)
+
+# Rust flags used in reproducible builds to replace common paths with defaults
+IDEMPOTENT_RUSTFLAGS=-C link-args=-Wl,--build-id=none --remap-path-prefix $(HOME)= --remap-path-prefix $(PROJECT_ROOT)= --remap-path-prefix $(RUST_COMPILER_SYS_ROOT)=
```
pronebird: `--build-id=none` does not work on macos so we need to condition it to linux only.

zaneschepke (approved):

PR #1827, opened 2024-12-20
Fix release build after adding default workspace members

octol:
Nightly build is failing https://github.com/nymtech/nym-vpn-client/actions/runs/12424425633 after adding default workspace members. This PR skips the steps in the release workflows that refers to nym-vpn-cli since that is an internal component anyway

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1827)

### DIFF
--- a/.github/workflows/build-nym-vpn-core-mac.yml
+++ b/.github/workflows/build-nym-vpn-core-mac.yml
@@ -75,7 +75,7 @@ jobs:
         env:
           RUSTFLAGS: "-L ${{ env.WG_GO_LIB_PATH }}"
         run: |
-          cargo build --${{ env.CARGO_TARGET }} --target x86_64-apple-darwin --workspace --exclude nym-gateway-probe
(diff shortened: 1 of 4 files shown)


.github/workflows/build-nym-vpn-core-mac.yml
```diff
@@ -75,6 +76,7 @@ jobs:
           RUSTFLAGS: "-L ${{ env.WG_GO_LIB_PATH }}"
         run: |
           cargo build --${{ env.CARGO_TARGET }} --target x86_64-apple-darwin --workspace --exclude nym-gateway-probe
```
pronebird: Remove `--exclude nym-gateway-probe` from here?

pronebird (changes requested):

.github/workflows/build-nym-vpn-core-mac.yml
octol: Good catch, fixed

pronebird (approved):

PR #1846, opened 2025-01-03
Switch from log to tracing

pronebird:
A large part of our codebase has already been using `tracing` instead of `log`, but bits of code brought in from here and there kept relying on `log`. This PR puts an end to this and migrates all logging to tracing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1846)


nym-vpn-core/crates/nym-dbus/src/network_manager.rs
```diff
@@ -721,14 +721,14 @@ fn eq_file_content<P: AsRef<Path>>(a: &P, b: &P) -> bool {
...
+            tracing::debug!("Failed to open file {}: {}", a.as_ref().display(), e);
             return false;
         }
     };
     let file_b = match File::open(b).map(BufReader::new) {
         Ok(file) => file,
         Err(e) => {
-            log::debug!("Failed to open file {}: {}", b.as_ref().display(), e);
+            tracing::debug!("Failed to open file {}: {}", b.as_ref().display(), e);
             return false;
```
doums, line 732: is a convention in those files to use `mod::fn()` for logs calls, instead of importing those `fn` once at the top of the files?

nym-vpn-core/crates/nym-dns/src/lib.rs
```diff
@@ -161,22 +161,22 @@ impl DnsMonitor {
 
     /// Set DNS to the given servers. And start monitoring the system for changes.
     pub fn set(&mut self, interface: &str, config: ResolvedDnsConfig) -> Result<(), Error> {
-        log::info!("Setting DNS servers: {config}",);
+        tracing::info!("Setting DNS servers: {config}",);
```
doums: remove the `,`

doums (approved):

nym-vpn-core/crates/nym-dbus/src/network_manager.rs
pronebird: Despite this being very much adopted crate which follows this style. I am the one to blame for spreading this style across the rest of the core lib.

Typically I am used to avoid importing standalone macros or functions, especially with generic or very common names to clearly hint that the function is defined elsewhere and not within the file.

My attitude towards this is shaped by [Mullvad code guidelines](https://github.com/mullvad/coding-guidelines/blob/main/rust.md#imports-and-modules) to the large extent, but also by collaboration with other developers that stuck to the same style of only importing types and not standalone functions or macros.


I don't think the rule is all that consistent however, because calling `anyhow::anyhow!()` is a [stutter](https://github.com/rust-lang/api-guidelines/discussions/66) so I'd usually import it disregarding the aforementioned guidelines. But then the guideline also states to follow the rule  "in most cases" but not all :-)

nym-vpn-core/crates/nym-dns/src/lib.rs
pronebird: Done.

PR #1926, opened 2025-01-14
Enable minimal Amnezia for all two-hop connections

jmwample:
Leverage the previous changes to enable the minimal (backwards compatible) amnezia wireguard obfuscation by default.

As the goal of this PR is to enable amnezia by default, the place where the `AmneziaConfig` is instantiated now uses a real config so that calling clients don't need to configure the wireguard node themselves (i.e  using `.with_amnezia_config(...)`).

Disabling amnezia for a build is still as easy as excluding the `"amnezia"` flag from the `nym-vpn-lib` crate.  Or if we want to add something like a runtime toggle down the line this could be set with, for example, `.with_amnezia_config(AmneziaConfig::OFF)`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1926)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/Cargo.toml
+++ b/nym-vpn-core/crates/nym-vpn-lib/Cargo.toml
@@ -116,6 +116,6 @@ vergen = { workspace = true, default-features = false, features = [
 ] }
 
 [features]
-default = []
+default = ["amnezia"]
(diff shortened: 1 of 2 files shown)


pronebird (approved):
This looks good to me. 👍

zaneschepke:
You can ignore android CI failing and merge this whenever you're ready!

PR #1950, opened 2025-01-17
[Breaking change] Add conversions from proto to core types

pronebird:
1. Add `From` conversions from proto to `TunnelState`
2. Simplify types used in `TunnelState` by converting some of the nym types into `String`
3. nym-lib-types: Add feature gate `nym-type-conversions` to remove dependendencies on nym crates when not needed.
4. Drop some of the old types and return `TunnelState` from grpc which is a *breaking change* and it will break desktop apps.
5. nym-vpnc: Leverage `From` conversions instead of working with raw grpc types when reporting vpn status.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1950)

PR #1967, opened 2025-01-21
refactor(tauri): new tunnel state api

doums:
_in short_ switch to the new tunnel state API

This introduces two new global statuses (user facing and relayed right into the UI):
- `Offline` (aka "no host internet connection")
- `Error` for tunnel state error

App is subscribing to `rpc ListenToEvents` and manages live tunnel state updates and mixnet events.
App queries the initial tunnel state from `rpc GetTunnelState`.
Any error sent from tunnel state, so `Error(reason)` will take precedence over any other errors in the UI.

**NOTE** this is a first iteration, the new tunnel state allow for new possible actions which are not part of this PR

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1967)


nym-vpn-app/src-tauri/src/grpc/tunnel.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub entry_gw_id: String,
+    pub exit_gw_id: String,
+    pub connected_at: Option<i64>,
+    pub data: TunnelData,
+}
+
+#[derive(Default, Debug, Clone, Serialize, PartialEq, TS, strum::Display)]
+#[ts(export)]
+#[serde(rename_all = "camelCase")]
+pub enum TunnelState {
```
pronebird, line 57: Looking good. A bit unfortunate that you can't reuse the `nym-vpn-lib-types` and `From` conversions provided by https://github.com/nymtech/nym-vpn-client/pull/1950.

doums: Ja that's fine. having full control over my own proto types translation is the way. Simple.
No need an extra traslation layer.
Protobuf is there to enforce the type system cross boundary anyway.

rokas-ambrazevicius (approved):

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
```diff
@@ -0,0 +1,264 @@
...
+                reconnect: o.reconnect,
+            },
+        })
+    }
+}
+
+impl From<&p::WireguardNode> for WgNode {
+    fn from(p_data: &p::WireguardNode) -> Self {
+        WgNode {
+            endpoint: p_data.endpoint.clone(),
```
pronebird: Implementing `From<p::WireguardNode>` instead would be nicer and would be more straightforward for call site that the data is not borrowed but actually cloned. With move semantics it's clear that data is moved. Why can't you consume the value?

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
pronebird: Most of these are copies of what `nym-vpn-lib-types` with `nym-vpn-proto` already offer including type conversions from/into protobuf. 

We should have discussed options for code reuse and if those TS derives could have been wrangled in some way.

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
```diff
@@ -0,0 +1,264 @@
...
+    Connecting(Option<Tunnel>),
+    Disconnecting(Option<TunnelAction>),
+    Error(TunnelError),
+    Offline {
+        reconnect: bool,
+    },
+}
+
+impl TunnelState {
+    pub fn from_proto(tunnel: &State) -> Result<TunnelState, &str> {
```
pronebird: Why not implement `TryFrom` trait instead?

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
doums: Yes each grpc client is (shoud be) responsible of its final proto types translation because each client (can) have different needs that you can't predict.
Like in this case with tauri and TypeScript types gen. But maybe in the future I would need to update some types for better fit because some specific needs in tauri side. Who knows?
Again I'm not a fanatic of creating (unnecessary) dependencies btw code for everything. Sometimes it's better to have some "code dup" in the name of better code resilience and flexibility.
Anyway the contract (which already enforce type safty cross boundary, and this should be our main and only concern right?) is protobuf itself. So that's alredy covered.

> We should have discussed options for code reuse and if those TS derives could have been wrangled in some way.

We have already discussed this point.

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
doums: good point, will switch to move

nym-vpn-app/src-tauri/src/grpc/tunnel.rs
doums: Why not using a simple method instead of trait implem?

pronebird: Well because `From` and `TryFrom` traits are somewhat standardised way to perform conversions. Also consistency, it's hard for me to say why sometimes you prefer From/TryFrom vs `from_proto()`

nym-vpn-app/src-tauri/src/grpc/events.rs
```diff
@@ -0,0 +1,67 @@
...
+        match event {
+            0 => Some(Self::EntryGwDown),
+            1 => Some(Self::ExitGwDownIpv4),
+            2 => Some(Self::ExitGwDownIpv6),
+            3 => Some(Self::ExitGwRoutingErrorIpv4),
+            4 => Some(Self::ExitGwRoutingErrorIpv6),
+            5 => Some(Self::ConnectedIpv4),
+            6 => Some(Self::ConnectedIpv6),
+            _ => unreachable!(),
+        }
```
doums, line 68: I failed to find a way to match against `Event::ConnectionEvent::*` enum variants instead of their corresponding index value.
So yea, this mapping is dirty, and we lose type safety here.

I'm not sure, but this could be due to the way the proto def has been declared:
Using an enum as a direct field value into a `oneof` message:

```proto
  oneof event {
    BandwidthEvent bandwidth_event = 1;
    ConnectionEvent connection_event = 2; // ConnectionEvent is a raw enum
    ConnectionStatisticsEvent connection_statistics_event = 3;
  }
```
In a next iteration, this declaration could be updated by wrapping the enum into a message and this could be fixed I guess.

pronebird: Proto supports `ConnectionEvent::try_from(i32)` to materialise raw value to type.

Maybe look at how it's done in https://github.com/nymtech/nym-vpn-client/blob/develop/nym-vpn-core/crates/nym-vpn-proto/src/conversions/from_proto/tunnel_event.rs#L56-L58

doums: > Proto supports `ConnectionEvent::try_from(i32)` to materialise raw value to type.

thx! That's what I was looking for

PR #1968, opened 2025-01-21
fix: unregister device raw

zaneschepke:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1968)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/platform/account.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/platform/account.rs
@@ -241,9 +241,9 @@ pub(super) async fn get_device_id() -> Result<String, VpnError> {
 pub(crate) mod raw {
     use std::path::Path;
 
-    use nym_sdk::mixnet::StoragePaths;
-
     use super::*;
+    use nym_sdk::mixnet::StoragePaths;
+    use nym_vpn_api_client::types::{Device, DeviceStatus};
 
     async fn setup_account_storage(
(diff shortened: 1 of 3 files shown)


rokas-ambrazevicius:
We should be calling this on "logout"?

zaneschepke:
> We should be calling this on "logout"?

I think it is correct to call this on forget account. This would mirror how it works on the other platforms.

nym-vpn-core/crates/nym-vpn-lib/src/platform/account.rs
```diff
@@ -326,14 +326,54 @@ pub(crate) mod raw {
...
+            .map_err(|err| VpnError::UnregisterDeviceApiClientFailure {
+                details: err.to_string(),
+            })?;
+        Ok(())
+    }
+
+    pub(crate) async fn forget_account_raw(path: &str, network: &str) -> Result<(), VpnError> {
         tracing::info!("REMOVING ALL ACCOUNT AND DEVICE DATA IN: {path}");
 
+        let network_env = nym_vpn_network_config::Network::fetch(network).map_err(|err| {
```
octol: Get the environment from `current_environment` instead, so you don't have to manually fetch it. And also risk having a network mismatch

octol (changes requested):

octol (approved):
Great work!

rokas-ambrazevicius (approved):

PR #1976, opened 2025-01-22
Replace windows-sys with windows

pronebird:
This is a backburner PR for replacing windows-sys with windows crate. 

It started as a side effort to update to `windows-sys 0.59`, which immediately caused issues with `socket2` due to peer dependency conflict with `windows-sys`. Upon further inspection, I realized that we don't really use `socket2` for anything but converting between windows native net types and Rust `std::net` types. 

Now the reasons to go with `windows` instead of lower level `windows-sys`:

- Support for most of non-fallible conversions between `SOCKADDR_INET`, `SOCKADDR_IN`, `SOCKADDR_IN6`, `IN_ADDR`, `IN6_ADDR` and `IP{v4,v6}Address`, `SocketAddr`, `SocketAddrV4`, `SocketAddrV6`
- Fallible WinAPI calls return `windows::core::Result`
- Fallible WinAPI calls returning `WIN32_ERROR` can be turned into `windows::core::Result` by calling `ok()` which removes the need for `win32_err` macro
- We can remove `socket2` from dependencies and only implement a single missing conversion from `SOCKADDR_INET` to `SocketAddr`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1976)


neacsu (approved):
> Reviewed 28 of 28 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1976)* status: all files reviewed, 1 unresolved discussion (waiting on @octol)

___
*[`nym-vpn-core/crates/nym-routing/src/windows/route_manager.rs` line 303 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1976#-OJrNyAh3cFUJo0fX_LX:-OJrNyAh3cFUJo0fX_LY:b250a5a) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/aa34e6c9b51fd2278c1124db33d2c0526758d671/nym-vpn-core/crates/nym-routing/src/windows/route_manager.rs#L303)):*
> ```XML
>                     .get_address()
>                     .map(|addr| {
>                         // Port should not matter so we set it to 0
> ```

If it doesn't matter maybe we should change the structure to not be socket address?

octol (approved):

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/1976)* status: all files reviewed, 1 unresolved discussion (waiting on @neacsu)

___
*[`nym-vpn-core/crates/nym-routing/src/windows/route_manager.rs` line 303 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/1976#-OJrNyAh3cFUJo0fX_LX:-OJrW-9WDwmKYKErh0gI:b-373n1s) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/aa34e6c9b51fd2278c1124db33d2c0526758d671/nym-vpn-core/crates/nym-routing/src/windows/route_manager.rs#L303)):*

If it doesn't matter maybe we should change the structure to not be socket address?
</blockquote></details>

That's a good question. I agree but since this was a part of original impl and `InterfaceAndGateway` is used in many places, I'd say we should look into that as a part of janitor duty at some point.

PR #2003, opened 2025-01-27
APPLE: improve debug logs

rokas-ambrazevicius:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2003)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
@@ -125,8 +125,9 @@ async fn configure_lib(data_dir: String, credential_mode: Option<bool>) -> Resul
     account::init_account_controller(PathBuf::from(data_dir), credential_mode, network).await
 }
 
-fn init_logger(path: Option<PathBuf>) {
-    let log_level = env::var("RUST_LOG").unwrap_or("info".to_string());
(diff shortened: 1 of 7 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/platform/mod.rs
```diff
@@ -125,8 +125,8 @@ async fn configure_lib(data_dir: String, credential_mode: Option<bool>) -> Resul
     account::init_account_controller(PathBuf::from(data_dir), credential_mode, network).await
 }
 
-fn init_logger(path: Option<PathBuf>) {
-    let log_level = env::var("RUST_LOG").unwrap_or("info".to_string());
+fn init_logger(path: Option<PathBuf>, debug_level: Option<String>) {
+    let log_level = debug_level.unwrap_or_else(|| "info".to_string());
```
jmwample: I don't think we should remove the ability to get the log level from the standard environment variable. I think the correct functionality is to use the debug_level if it is provided or use the env / default level if not. 

```rs
     let log_level = debug_level.unwrap_or(env::var("RUST_LOG").unwrap_or("info".to_string()));
```

or maybe more clear
```rs
     let default_log_level = env::var("RUST_LOG").unwrap_or("info".to_string());
     let log_level = debug_level.unwrap_or(default_log_level);
```

pronebird (approved):

PR #2023, opened 2025-01-28
Log to file on macos

jmwample:
When running as a service on macos logs output to service log file path ( in addition to oslog).

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2023)

### DIFF
--- a/nym-vpn-core/crates/nym-vpnd/src/logging.rs
+++ b/nym-vpn-core/crates/nym-vpnd/src/logging.rs
@@ -11,7 +11,8 @@ pub fn setup_logging(_as_service: bool) {
     #[cfg(target_os = "macos")]
     if _as_service {
         let log_level = env::var("RUST_LOG").unwrap_or("info".to_string());
-        nym_vpn_lib::swift::init_logs(log_level, "/var/log/nym-vpnd/daemon.log");
+        let log_path = PathBuf::from("/var/log/nym-vpnd/daemon.log");
+        nym_vpn_lib::swift::init_logs(log_level, Some(log_path));
         return;
     }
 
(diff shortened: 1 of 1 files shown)


nym-vpn-core/crates/nym-vpnd/src/logging.rs
```diff
@@ -11,7 +11,8 @@ pub fn setup_logging(_as_service: bool) {
     #[cfg(target_os = "macos")]
     if _as_service {
         let log_level = env::var("RUST_LOG").unwrap_or("info".to_string());
-        nym_vpn_lib::swift::init_logs(log_level, None);
+        let log_path = PathBuf::from("/var/log/nym-vpnd/daemon.log");
```
pronebird: If you look just below in setup_logging_to_file, we have constants for this defined in `service` as `DEFAULT_LOG_FILE`, `DEFAULT_LOG_DIR`. Maybe you could try reusing those? Thanks.

rokas-ambrazevicius:
Will we be able to access this file, when the owner is daemon with sudo? Maybe it would be better to pass in a path to a file created without sudo access?

pronebird:
> Will we be able to access this file, when the owner is daemon with sudo? Maybe it would be better to pass in a path to a file created without sudo access?

The file is owned by root and there is no problem making it readable by others. Otherwise good question about default permissions set on log file.

pronebird (changes requested):

pronebird (approved):

pronebird:
clippy is unhappy, otherwise it looks good. Thanks!

jmwample:
For a root user on a unix system, the default folder permissions are set to 755 (rwxr-xr-x), and file permissions to 644 (rw-r--r--).
Being a system folder `/var/log` is readable by all users. `/var/log/nym-vpnd` will either either be created by the file appender with default permissions or we could manage it as part of our (systemd? I don't know macos) service file so that it is created under the user/group that our service runs on.

PR #2024, opened 2025-01-28
wf(tauri): switch onto rust stable

doums:
Switch tauri Windows build to Rust toolchain stable

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2024)


pronebird:
Actually looks good to me. @octol what's your take on switching to stable rust?

doums:
@pronebird stable is used for Linux build, and has been pinned to 1.82 for Windows because if not mullvad libs build was failling
but since we disabled mullvad libs

octol (approved):
Looking good!

pronebird:
@doums makes sense. Thanks for explanation.

PR #2026, opened 2025-01-29
Introduce ux scores

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2026)


zaneschepke:
We need name as well

zaneschepke (approved):

rokas-ambrazevicius (approved):

proto/nym/vpn.proto
```diff
@@ -554,10 +554,23 @@ message Probe {
   ProbeOutcome outcome = 2;
 }
 
+message UxScore {
+  uint32 max_score = 1;
+  uint32 current_score = 2;
+  string color_hex = 3;
```
doums: Why a color hex in there?
Colors should be UI style related right?

doums: We should avoid including raw UI style stuff (like color values) into our business logic data.
Why not just using a classic enum?

```
  enum GatewayHealth {
    GOOD = 0;
    MEDIUM = 1;
    LOW = 2;
    VERY_LOW = 3;
    DOWN = 4;
  }
```

And deferring the style logic as usual, to the app UI layer and designs etc

pronebird: Note that knowing the min and max score you can technically do your color mapping. I assume that color_hex is basically used by backend in dashboard.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2026)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @doums, @octol, and @zaneschepke)

proto/nym/vpn.proto
zaneschepke: I tried to express this in the stand ups, but they insisted having this and no one else opposed. The api is already changed and we can't change it again so it is what it is now.

zaneschepke (approved):

PR #2060, opened 2025-02-03
Fix linux freeze

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2060)

Cherry pick Handle cancellation when underlying channel(1) is congested (#2034)


doums (approved):

pronebird (dismissed):
Still behaves weirdly on my Linux VM with 4 CPU cores. The status listener keeps posting events even after the task manager is shut down.

pronebird:
With the recent changes, the two-hop mode seems to work fine. So let's retarget that towards `develop` and merge. We can tackle mixnet in a separate PR.

PR #2143, opened 2025-02-12
Fix publish conditional in workflow

octol:
Fix the conditional that determines if the publish workflow should also enable the step for publishing to github, by making sure it's also true for release tags being pushed

We unconditionally always publish to our build archive.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2143)


pronebird:
@octol can you elaborate why this is needed? The PR description is blank.

octol:
> @octol can you elaborate why this is needed? The PR description is blank.

Updated the description

pronebird (approved):

pronebird:
@octol thanks that makes sense to me.

PR #2157, opened 2025-02-12
build: update aur packages and linux script

doums:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2157)


.pkg/linux/install
```diff
@@ -38,19 +38,19 @@ BI_MGT="$ITL$B_MGT"
...
 icon_url="https://raw.githubusercontent.com/nymtech/nym-vpn-client/fb526935/nym-vpn-app/.pkg/icon.svg"
 
 # nym-vpnd
-vpnd_tag=nym-vpn-core-v1.3.0-beta.1
+vpnd_tag=nym-vpn-core-v1.3.1
 # ⚠ there are inconsistencies in the package version naming…
-vpnd_version=1.3.0-beta.1
-vpnd_deb_ver=1.3.0~beta.1
-vpnd_deb_ver_url=1.3.0.beta.1
+vpnd_version=1.3.1
```
octol, line 51: Get the version from the tag? Or other way around?

octol (approved):

.pkg/linux/install
doums: This is an internal custom script, we're using to get quickly setup, especially for testing.
I need to swap the versions frequently, depending on what build is the current hotspot, dev nightly beta or latest stable. Vpnd and app unsync setup is also possible, eg: nightly vpnd with latest RC app.
So automation is not really possible.

doums: Imo, the issue is everything in this repo is push protected and require PR + 3d party approval.
Such changes, like this script update, does not require that process.

PR #2162, opened 2025-02-13
Migrate dns to async runtime

pronebird:
- Drop `parking_lot::Mutex`, use tokio `Mutex` instead which supports blocking/async locking.
- Update `DnsMonitorT` trait with async methods and eliminate `tokio::runtime::Handle` that used to be passed around
- Eliminate `tokio::task::block_in_place`

Further notes:
Some parts of code still use sync `std::fs` and need to be migrated to `tokio::fs`. This PR does not touch on that due to time constraints and the impact on dns initialization flow which would need to be refactored a bit to become async.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2162)


neacsu (approved):
> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2162)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @octol)

PR #2178, opened 2025-02-13
Log the RTT estimated remote time of the vpn-api

octol:
When syncing the account, also try to estimate the remote vpn-api time and log the results. If the skew is large enough, emit a warning.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2178)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
+++ b/nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
@@ -97,18 +97,21 @@ impl VpnApiTime {
 
     // Local time minus remote time. Meaning if the value is positive, the local time is ahead
     // of the remote time.
-    pub fn skew(&self) -> Duration {
+    pub fn local_time_ahead_skew(&self) -> Duration {
         self.local_time - self.estimated_remote_time
     }
 
     pub fn is_synced(&self) -> bool {
-        self.skew().abs().whole_seconds() < MAX_ACCEPTABLE_SKEW_SECONDS
+        self.local_time_ahead_skew().abs().whole_seconds() < MAX_ACCEPTABLE_SKEW_SECONDS
     }
 
     pub fn estimate_remote_now(&self) -> OffsetDateTime {
-        tracing::debug!("Estimating remote now using skew: {}", self.skew());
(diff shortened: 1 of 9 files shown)


nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
```diff
@@ -57,6 +60,72 @@ fn cosmos_derivation_path() -> DerivationPath {
...
+    }
+
+    // Local time minus remote time. Meaning if the value is positive, the local time is ahead
+    // of the remote time.
+    pub fn skew(&self) -> Duration {
+        self.local_time - self.estimated_remote_time
+    }
+
+    pub fn is_synced(&self) -> bool {
+        self.skew().abs().whole_seconds() < 30
```
neacsu: Use a constant here

nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
```diff
@@ -57,6 +60,72 @@ fn cosmos_derivation_path() -> DerivationPath {
...
+    }
+
+    pub fn estimate_remote_now(&self) -> OffsetDateTime {
+        tracing::debug!("Estimating remote now using skew: {}", self.skew());
+        let now = OffsetDateTime::now_utc();
+        now - self.skew()
```
neacsu: Short comment here: this is a bit dependent on the fact that we compute `skew()` by doing local - remote. If that subtraction would be inverted in the `skew` function, it would still respect its name definition, but the estimate here would be doubling the possible error.
In any case, I think this is good enough for the purposes we care about. Good job on playing with those signs and timestamps.

neacsu (approved):

nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
octol: Fixed

nym-vpn-core/crates/nym-vpn-api-client/src/types/account.rs
octol: I think it's a very good point. Updated the function names to try to reduce the risk of this footgun

PR #2184, opened 2025-02-14
Fix task handling in wireguard mode

octol:
- Remove the unused reconnect logic in the bandwidth controller and go back to a single unified `TaskManager`.
- Dispose the shared mixnet client after signal shutdown but before dropping the task manager

This should fix all those unexpected drop (VPN-2849) errors that happen during disconnect.

IMPORTANT: we need to solidly confirm that the reconnect logic is as unused as I think it is. And test it properly on Android

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2184)

### DIFF
--- a/nym-vpn-core/crates/nym-mixnet-client/src/lib.rs
+++ b/nym-vpn-core/crates/nym-mixnet-client/src/lib.rs
@@ -71,4 +71,9 @@ impl SharedMixnetClient {
     pub fn connection_fd_callback(&self) -> Arc<dyn Fn(RawFd) + Send + Sync> {
         self.connection_fd_callback.clone()
     }
+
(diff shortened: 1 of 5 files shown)


neacsu (approved):

pronebird (commented):
This needs to be tested.

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2184#-:-OJNd8Ak5Ll5QRQog80Q:bm9cesl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @octol)

zaneschepke (approved):

PR #2198, opened 2025-02-17
tauri: turn on zknym

doums:
- hard turns on zknyms
- prepare for dev release

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2198)


doums:
Actually we don't we don't want to merge that one

PR #2222, opened 2025-02-19
ci(tauri): update ci for dev builds

doums:
- since any non-public builds are removed from GH, pull the dev builds from the external site
- add an optional input to the workflow allowing to pass a direct URL to vpn-core artifact for windows

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2222)

### DIFF
--- a/.github/workflows/build-nym-vpn-app-windows.yml
+++ b/.github/workflows/build-nym-vpn-app-windows.yml
@@ -8,7 +8,11 @@ on:
         type: boolean
         default: false
       core_release_tag:
-        description: "nym-vpn-core release tag to use for Windows build (as it bundles the daemon)"
+        description: "nym-vpn-core release tag from GH to use for Windows, to get the daemon"
+        required: false
+        type: string
+      core_artifact_url:
+        description: "direct nym-vpn-core artifact URL for Windows, to get the daemon"
         required: false
         type: string
       sign:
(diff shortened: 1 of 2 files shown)


octol (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2222)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`.github/workflows/build-nym-vpn-app-windows.yml` line 175 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/2222#-OJVOe5K2jdaauKfnkva:-OJVOe5LCMnP7uGmTsDN:b-kpn93z) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/12de79d9b172b8e7da6f61788669228969d86399/.github/workflows/build-nym-vpn-app-windows.yml#L175)):*
> ```YAML
>         shell: bash
>         run: |
>           latest=$(curl -s ${{ env.VPN_CORE_BUILDS_URL }}/develop \
> ```

Instead of locking yourself to `develop` like this, you could for example fetch the list of the available versions from develop and the last couple of release directories. Then select the newest core that has the same major and minor version as the app

doums:
@octol the purpose of this PR is not to refactor entirely how the app release workflow is automated and switching to a new versioning semantic between the app and the daemon.

But simply to adapt this current (working) flow with the new external website, where the dev vpn-core builds are now published.
Also, I need to release asap a dev version as discussed today: can you unblock the PR?

octol (approved):
> @octol the purpose of this PR is not to refactor entirely how the app release workflow is automated and switching to a new versioning semantic between the app and the daemon.
> 
> But simply to adapt this current (working) flow with the new external website, where the dev vpn-core builds are now published. Also, I need to release asap a dev version as discussed today: can you unblock the PR?

It's a relatively straightforward change that will simplify and increase correctness. Sticking to this solution contributes to slowing us down and I wish there would be more interest in improving these workflows

doums:
what is slowing us down is, for example, this simple PR to 'patch' an already working flow to have been delayed for ~20h in review

PR #2260, opened 2025-02-24
Disable cover traffic for wireguard on Android

pronebird:
We should really stop running cover traffic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2260)


zaneschepke (approved):

zaneschepke:
Tested it and works greats.

PR #2306, opened 2025-02-27
feat(tauri): save nodes by vpn mode

doums:
Save 2 different nodes selections, one for wg, one for mixnet.
Basically, this PR allows the user to use 2 different nodes' config, one for each vpn mode, allowing to use different nodes by modes. Moreover, the nodes' topology is expected to be different depending on the VPN mode: nodes in one mode can be unavailable in the other. Or can have different score and so performances. These are reasons why a user would want to use different nodes btw the modes.
Accordingly, we need to handle 2 distinct node selections, for a good UX.

(no UI changes)

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2306)


rokas-ambrazevicius:
Doesn't it look confusing if for one mode GB GB in the other US US, when switching between modes?

doums:
I don't think so.

zaneschepke:
I disagree with these changes. I don't think our current design makes it obvious that each mode would save its own gateway selection config.

doums:
Okay, let's close it then. As I understand it, we will have future UX iterations to improve this.

PR #2313, opened 2025-02-28
core: login registration failure

zaneschepke:
* forget account on registration failures at login

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2313)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
@@ -45,7 +45,14 @@ impl AccountControllerCommander {
         self.store_account(mnemonic).await?;
         self.ensure_update_account().await?;
         self.ensure_update_device().await?;
-        self.ensure_register_device().await?;
+        //if register fails, forget the account
+        match self.ensure_register_device().await {
+            Ok(()) => Ok(()),
+            Err(register_error) => {
+                let _ = self.forget_account().await;
+                Err(AccountCommandError::RegisterDevice(register_error))
+            }
+        }?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
```diff
@@ -45,7 +45,14 @@ impl AccountControllerCommander {
         self.store_account(mnemonic).await?;
         self.ensure_update_account().await?;
         self.ensure_update_device().await?;
-        self.ensure_register_device().await?;
+        // forget account on registration failures
+        match self.ensure_register_device().await {
+            Ok(()) => Ok(()),
+            Err(register_error) => {
+                let _ = self.forget_account().await;
```
octol: Confirming that this is what we want. I'm thinking that on the app side we could handle failure to login (basically store_account) separately from failure to register device

zaneschepke: Is there a reason we would like to do this? I like to handle this in the lib because then it is the same logic for all apps. It also then behaves more like a proper "login" function.

octol: to me it seems fragile and prone to breakage and inconsistent state, given the number of files that is being touched by forget_account

zaneschepke: I think it is best if we ignore this registration error and just proceed. Will will already have a failed registration error flow on connect in the event the device was unregistered in the web.

nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
```diff
@@ -45,7 +45,8 @@ impl AccountControllerCommander {
         self.store_account(mnemonic).await?;
         self.ensure_update_account().await?;
         self.ensure_update_device().await?;
-        self.ensure_register_device().await?;
+        // ignore registration errors
+        let _ = self.ensure_register_device().await?;
```
octol: We can just delete the line altogether. Once you store the account the registration is kicked off in the background. The `ensure_register_device()` call is for waiting for the result of this, but since we ignore the result we don't need to wait for it here

octol (approved):

PR #2332, opened 2025-03-03
Derive probe keys from known values

durch:
Adds two CLI args to derive probe keys. Both are set to default values so that existing setup continues to work, however it would be good to change them to something else. Key is derived from the master key and salt specified on the command line, and an index is derived from the tested node identity, this in practice means that all runs against the same node will have the same identity, and thats probably fine for now, it will make it easier to filter them out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2332)


pronebird:
Gateway probe is no longer a part of this repository. Closing. Please re-open in nym repository if you still wish these changes to be merged.

PR #2348, opened 2025-03-04
Add entry/exit ip to mixnet connection data

pronebird:
- Add entry/exit ip to mixnet connection data
- Update tunnel state display implementation

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2348)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib-types/src/connection_data.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib-types/src/connection_data.rs
@@ -3,7 +3,7 @@
 
 use std::{
     fmt,
-    net::{Ipv4Addr, Ipv6Addr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
 };
 
 use time::OffsetDateTime;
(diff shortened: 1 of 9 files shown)


rokas-ambrazevicius (approved):

nym-vpn-core/crates/nym-vpn-lib-types/src/connection_data.rs
```diff
@@ -70,6 +70,13 @@ impl NymAddress {
     pub fn new(nym_address: String) -> Self {
         Self { nym_address }
     }
+
+    pub fn gateway_id(&self) -> &str {
+        self.nym_address
+            .rfind("@")
+            .map(|i| self.nym_address.split_at(i + 1).1)
+            .unwrap_or(self.nym_address.as_str())
+    }
```
neacsu: I think instead of re-doing this parse, of which format is external to this codebase, we should store `nym_address` as `nym_gateway_directory::Recipient` instead, and take it from there.

neacsu (changes requested):

nym-vpn-core/crates/nym-vpn-lib-types/src/connection_data.rs
pronebird: This crate specifically avoids having dependency on nym types because a lot of them are heavy and this layer does not need to have in-depth knowledge of internal types. `Recipient` itself is also 416 bytes long, almost always requires boxing etc. Let me rework that to pass gateway_id as a part of initializer.

pronebird: Fixed by parsing and passing `gateway_id` separately during initialization.

neacsu (approved):

PR #2368, opened 2025-03-05
daemon: add -v to enable debug, -vv for trace

pronebird:
This PR adds a `-v` flag to the daemon to control verbosity level without using environment variables to improve ergonomics.

- Omitting `-v` is equivalent `INFO` log level
- `-v` - `DEBUG` log level
- `-vv` (or more 'v') - `TRACE`

I believe that `RUST_LOG` overrides this behaviour as `from_env_lossy()` should default to whatever is set in the env if present. I am not sure if that should be the other way around or not. This was a quick PR and it makes it easier to toggle log levels.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2368)


octol (approved):
Nice :)
Everything that improves logging is big win

PR #2375, opened 2025-03-06
Sync JWT time with remote

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2375)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
+++ b/nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
@@ -36,6 +36,8 @@ pub(crate) const NYM_VPN_API_TIMEOUT: Duration = Duration::from_secs(60);
 #[derive(Clone, Debug)]
 pub struct VpnApiClient {
     inner: nym_http_api_client::Client,
+    // compensation, in case the skew is too big
+    vpn_api_time: Option<VpnApiTime>,
 }
 
 impl VpnApiClient {
@@ -82,7 +84,7 @@ impl VpnApiClient {
                 builder
             })
             .and_then(|builder| builder.build())
-            .map(|c| Self { inner: c })
+            .map(|c| Self { inner: c, vpn_api_time: None })
             .map_err(VpnApiClientError::FailedToCreateVpnApiClient)
     }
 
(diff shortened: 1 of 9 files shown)


nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
```diff
@@ -82,24 +85,46 @@ impl VpnApiClient {
...
+            tracing::warn!(
+                "The time skew between the local and remote time is too large, we'll use remote instead for JWT ({remote_time})."
+            );
+            true
+        }
+    }
+
+    pub async fn sync_with_remote_time(&mut self) -> Result<()> {
         let time_before = OffsetDateTime::now_utc();
         let remote_timestamp = self.get_health().await?.timestamp_utc;
```
octol, line 117: nit: what about adding a `.inspect_err` here and then removing the `if let Err() = ... tracing::error` on the callsites? Since we always continue on error anyway

nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
```diff
@@ -277,15 +306,16 @@ impl VpnApiClient {
         B: Serialize,
         E: fmt::Display + DeserializeOwned,
     {
+        let jwt = *self.vpn_api_time.read().await;
         let request = self
             .inner
             .create_patch_request(path, NO_PARAMS, json_body)
-            .bearer_auth(account.jwt(None).to_string());
+            .bearer_auth(account.jwt(jwt).to_string());
```
octol, line 313: When we construct the jwt. I've forgotten which way around it is, but just double check that it's the time and not the skew that is passed down here

octol (approved):
Great :)

nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
neacsu: Yep, it's the time

nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
neacsu: That sounds good

PR #2389, opened 2025-03-07
Adapt winfw for our requirements

pronebird:
- `build-windows-modules.ps1`: Add parameter that enables auto-copy of `winfw.dll` into `target\debug` when developing
- Add `InterfaceType` describing entry and exit tunnel interfaces and use it to set up the second set of filters using unique GUIDs. This is important because we want individual rules match IP, port, transport and each but also we cannot install the filter with the same GUID twice so we have to swap GUIDs based on input parameters.
- Add option to configure up to 4 IP addresses in vpn relay filter. We use that for mixnet + wireguard endpoints.
- Add option to configure up to 8 IP addresses for allowed endpoints. We use that for APIs.
- Allow more DNS ports which should really be revamped to be stricter, alas not now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2389)

### DIFF
--- a/build-windows-modules.ps1
+++ b/build-windows-modules.ps1
@@ -1,6 +1,16 @@
 param (
-    [Parameter(Mandatory = $true)][string]$BuildConfiguration,
-    [Parameter(Mandatory = $true)][string]$Platform,
+    [Parameter(Mandatory = $true)]
+    [ValidateSet('Release', 'Debug')]
+    # Build configuration
+    [string]$BuildConfiguration,
+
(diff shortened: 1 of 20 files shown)


neacsu (approved):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @doums)

PR #2391, opened 2025-03-07
+ Crowdin macos config and github action

windy-ux:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2391)


.github/workflows/crowdin-macos-ios.yml
```diff
@@ -0,0 +1,26 @@
+name: Crowdin Action
```
octol: The naming convention is the same as the filename, minus the suffix

octol (changes requested):

octol (approved):

PR #2419, opened 2025-03-09
Wire up system_configuration statistics_recipient

octol:
Wire up statistics_recipient in network configuration, if it is provided.

- **Parse statistics recipient from network configuration**
- **Set statistics recipient in uniffi**

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2419)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-network-config/src/system_configuration.rs
+++ b/nym-vpn-core/crates/nym-vpn-network-config/src/system_configuration.rs
@@ -17,8 +17,8 @@ impl fmt::Display for SystemConfiguration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "mixnet score thresholds: {:?}\nwireguard score thresholds: {:?}",
-            self.mix_thresholds, self.wg_thresholds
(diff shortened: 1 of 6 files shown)


nym-vpn-core/crates/nym-vpn-network-config/src/system_configuration.rs
```diff
@@ -3,12 +3,14 @@
...
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SystemConfiguration {
     pub mix_thresholds: ScoreThresholds,
     pub wg_thresholds: ScoreThresholds,
+    pub statistics_recipient: Option<Recipient>,
 }
 
 impl fmt::Display for SystemConfiguration {
```
simonwicky, line 16: Do we want to include that recipient address in the `Display` implementation?

simonwicky (approved):

nym-vpn-core/crates/nym-vpn-network-config/src/system_configuration.rs
octol: I think that makes sense. Will update

PR #2459, opened 2025-03-12
Increase the required number of zknyms before trying to connect

octol:
Increase the number of required ticketbooks in storage before allowing to proceed with attempting to connect. Also reduce the interval time when the bakground check is triggered to 1 minute

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2459)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
@@ -190,10 +190,10 @@ impl AccountControllerCommander {
             .get_available_tickets()
             .await
             .map_err(|err| RequestZkNymError::CredentialStorage(err.to_string()))?
-            .is_all_ticket_types_above_threshold(0)
+            .is_all_ticket_types_above_soft_threshold()
         {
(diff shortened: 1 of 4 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
```diff
@@ -190,10 +190,10 @@ impl AccountControllerCommander {
             .get_available_tickets()
             .await
             .map_err(|err| RequestZkNymError::CredentialStorage(err.to_string()))?
-            .is_all_ticket_types_above_threshold(0)
+            .is_all_ticket_types_above_soft_threshold()
```
neacsu, line 193: This is more aggressive than the 1 threshold we mentioned, right? I see this is above 30 tickets as threshold

neacsu (approved):

nym-vpn-core/crates/nym-vpn-account-controller/src/commander.rs
octol: Yep, if below 30 tickets then request more before trying to connect, so we have comfortably above the minimum number of tickets to operate

PR #2468, opened 2025-03-13
vpnd: remove HTTP listener

pronebird:
1. Remove HTTP listener and `--enable-http-listener` CLI flag
2. Remove `--disable-socket-listener` CLI flag since the daemon without any management interface is not manageable.
3. Make `remove_previous_socket_file()` tokio friendly
4. Only remove previous socket file on unix platforms
5. `Uds` from `nym-ipc` already removes socket file on `Drop`, so we can stop doing that in management interface.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2468)


doums (approved):

octol (approved):
I see lots of removed code, fantastic :)

PR #2481, opened 2025-03-16
Fix initial flag value for new discovery file

neacsu:
When the discovery file is initially created, the populated default value for zk-nym enable is false, and that value doesn't get updated until the next daemon restart, if that happens after 60 seconds (is considered stale).
Since moving to default zk-nym enabled, we should also make this true in the default values of the initial file

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2481)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-network-config/default/mainnet_discovery.json
+++ b/nym-vpn-core/crates/nym-vpn-network-config/default/mainnet_discovery.json
@@ -1,5 +1,10 @@
 {
   "network_name": "mainnet",
   "nym_api_url": "https://validator.nymtech.net/api/",
-  "nym_vpn_api_url": "https://nymvpn.com/api/"
+  "nym_vpn_api_url": "https://nymvpn.com/api/",
+  "feature_flags": {
+    "zkNyms": {
+      "credentialMode": "true"
+    }
+  }
 }
--- a/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
+++ b/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
@@ -126,23 +126,29 @@ impl Discovery {
 
     pub(super) fn ensure_exists(config_dir: &Path, network_name: &str) -> anyhow::Result<Self> {
         if !Self::path(config_dir, network_name).exists() && network_name == "mainnet" {
-            tracing::info!("No discovery file found, writing default discovery file");
-            Self::default()
+            tracing::info!("No discovery file found, writing creating a new discovery file");
+            Self::fetch(network_name)
+                .inspect_err(|err| {
+                    tracing::warn!(
+                        "Failed to fetch remote discovery file: {err}, creating a default one"
+                    )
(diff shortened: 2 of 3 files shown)


nym-vpn-core/crates/nym-vpn-network-config/default/mainnet_discovery.json
```diff
@@ -1,5 +1,10 @@
 {
   "network_name": "mainnet",
   "nym_api_url": "https://validator.nymtech.net/api/",
-  "nym_vpn_api_url": "https://nymvpn.com/api/"
-}
+  "nym_vpn_api_url": "https://nymvpn.com/api/",
+  "feature_flags": {
+    "zkNyms": {
+      "credentialMode": "true"
```
octol, line 7: Could this not be problematic on eg canary etc where it's disabled?

octol (commented):
I'm thinking the intended behaviour would that the bundled file is written to disk on first startup, but we also try to fetch the network one on first vpnd start? So the bundled one is only very rarely used in practice?

nym-vpn-core/crates/nym-vpn-network-config/default/mainnet_discovery.json
neacsu: I've changed it to be the default of mainnet only

neacsu:
> I'm thinking the intended behaviour would that the bundled file is written to disk on first startup, but we also try to fetch the network one on first vpnd start? So the bundled one is only very rarely used in practice?

Changed it to reflect the behaviour you described i.e. the default is only for mainnet, when there isn't an existing file in there and after the network fetch fails for whatever reason

nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
```diff
@@ -126,24 +126,30 @@ impl Discovery {
...
                 .write_to_file(config_dir)
-                .inspect_err(|err| tracing::warn!("Failed to write default discovery file: {err}"))
+                .inspect_err(|err| tracing::warn!("Failed to write discovery file: {err}"))
+                .ok();
+        } else {
+            // Download the file if it doesn't exists, or if the file is too old, refresh it.
+            // TODO: in the future, we should only refresh the discovery file when the tunnel is up.
+            // Probably in a background task.
+
+            Self::try_update_file(config_dir, network_name)
```
octol, line 145: I think it's for another PR, but probably as part of VPN-3071 we want to always try to update the file? Even on the first startup, and only fallback to the bundled one if fetching fails?

octol (approved):

nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
neacsu: But this would always update it. On the first startup, when there's nothing there, it would still first try to fetch it and fallback to bundled one if network fetch fails

PR #2498, opened 2025-03-17
Add discovery refresher

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2498)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
+++ b/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
@@ -59,7 +59,7 @@ impl Discovery {
         }
     }
 
-    pub fn fetch(network_name: &str) -> anyhow::Result<Self> {
+    pub async fn fetch(network_name: &str) -> anyhow::Result<Self> {
         // allow panic because a broken bootstrap url means everything will fail anyways.
         #[allow(clippy::expect_used)]
         let default_url = Self::DEFAULT_VPN_API_URL
@@ -68,12 +68,7 @@ impl Discovery {
         let client = BootstrapVpnApiClient::new(default_url)?;
 
         tracing::debug!("Fetching nym network discovery");
-        let rt = tokio::runtime::Runtime::new()?;
-
-        // Spawn the root task
-        let discovery: NymWellknownDiscoveryItemResponse = rt
-            .block_on(client.get_wellknown_discovery(network_name))
-            .with_context(|| "Failed to read response text")?;
+        let discovery = client.get_wellknown_discovery(network_name).await?;
 
         tracing::debug!("Discovery response: {:#?}", discovery);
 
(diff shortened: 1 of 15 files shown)


nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
```diff
@@ -1,47 +1,67 @@
...
         // Check once an hour
-        let mut interval = tokio::time::interval(Duration::from_secs(60 * 60));
+        let mut interval = tokio::time::interval(Duration::from_secs(60));
         interval.tick().await; // initial tick
 
         loop {
             tokio::select! {
                 _ = interval.tick() => {
-                    if let Err(err) = refresh_discovery_file(&config_path, &network_name) {
+                    if let Err(err) = self.refresh_discovery_file().await {
```
octol: We probably need to wrap this in a mechanism to abort if stuck, just so we don't end up in frozen state on shutdown. Maybe `self.cancel_token.run_until_cancelled(..)` but we also don't want to forcefully cancel if not necessary. In case we cancel while writing a file

nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
```diff
@@ -68,12 +68,7 @@ impl Discovery {
         let client = BootstrapVpnApiClient::new(default_url)?;
 
         tracing::debug!("Fetching nym network discovery");
-        let rt = tokio::runtime::Runtime::new()?;
-
-        // Spawn the root task
-        let discovery: NymWellknownDiscoveryItemResponse = rt
-            .block_on(client.get_wellknown_discovery(network_name))
-            .with_context(|| "Failed to read response text")?;
```
octol, line 76: Nice getting rid of this, and having the whole call stack async instead

nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
```diff
@@ -1,47 +1,67 @@
...
+        }
+        tracing::info!("Refreshing and updating file");
+        Discovery::fetch(&self.network_name)
+            .await?
+            .write_to_file(self.config_path.as_path())?;
+        Ok(())
+    }
+
+    async fn run(self) {
         // Check once an hour
```
octol, line 37: Update comment, or remove

nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
```diff
@@ -1,47 +1,67 @@
...
-        let mut interval = tokio::time::interval(Duration::from_secs(60 * 60));
+        let mut interval = tokio::time::interval(Duration::from_secs(60));
```
octol: Having this check every 60 sec is fine, but then I think the max file age should be bumped to something larger. We don't need to overwrite this file every minute imho

I think the max age used to be something pretty large, every 24h. But it was changed to 60sec since we wanted to make sure to fetch the network messages when updated.

Currently the startup logic is something like

0. write default config if nothing exists
1. check if file is older than max age
2. update if older than max age

But I'm thinking we should change it to

0. write default config if nothing exists
1. always try update file on startup
2. update in the background periodically

nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
neacsu: Or maybe even update the actual check, 60 seconds still seems quite often

nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
neacsu: Applied the feedback. Thanks

nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs
neacsu: In this case the fs won't be cancelled mid write because it is not async. The only thing that can get cancelled is the network `fetch`, which I would say in this context, is cancel safe, as the operation is just a GET.

octol (approved):
Looking good!
Ideally it would be good to make android green by gating the struct value with `cfg` or change on the kotlin side to pass in None

PR #2520, opened 2025-03-19
feat(grpc, tauri, vpnd): daemon logs

doums:
- vpnd: add a new grpc `GetLogPath` to query from daemon its log path
- vpnd: enable log file when running as systemd service, using cli flag `--run-as-service`
- tauri: update UI to open daemon logs directory in the FileManager
- when daemon is down, it's fallback to default paths
- concurrently with log file mode, when app is running from a terminal enable stdout logging

Current implementation on tauri is a bit clunky because file paths need to be whitelisted (`capabilities/app.json`) in order to allow access.
For now let's assume possible paths are `/var/log/nym-vpnd` on *unix and `C:\ProgramData\nym-vpnd\log` on Windows.
If we want to support other, we would need to update the capabilities accordingly.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2520)


rokas-ambrazevicius (approved):

nym-vpn-core/crates/nym-vpnd/src/command_interface/listener.rs
```diff
@@ -895,6 +896,18 @@ impl NymVpnd for CommandInterface {
...
     }
+
+    async fn get_log_path(
+        &self,
+        _: tonic::Request<()>,
+    ) -> Result<tonic::Response<GetLogPathResponse>, tonic::Status> {
+        let dir = service::log_dir();
+        tracing::debug!("log file path: {}", dir.display());
+        Ok(tonic::Response::new(GetLogPathResponse {
+            path: dir.to_string_lossy().to_string(),
```
octol, line 907: what about `to_display()`?

nym-vpn-core/crates/nym-vpnd/src/command_interface/listener.rs
```diff
@@ -895,6 +896,18 @@ impl NymVpnd for CommandInterface {
 
         Ok(tonic::Response::new(response))
     }
+
+    async fn get_log_path(
+        &self,
+        _: tonic::Request<()>,
+    ) -> Result<tonic::Response<GetLogPathResponse>, tonic::Status> {
+        let dir = service::log_dir();
```
octol, line 904: I can see this being fragile in the future. Ideally I think we should in the running instance track the log path similar to how we track the data path, so that we return the log dir used (output), not the input

Maybe you could try adding this to `LoggingSetup`?

doums: Daemon architecture is quite complex
I'll try to improve this in a followup PR

PR #2523, opened 2025-03-20
Do remote jwt on case by case basis

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2523)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
+++ b/nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
@@ -1,14 +1,13 @@
 // Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::{fmt, net::SocketAddr, sync::Arc, time::Duration};
+use std::{fmt, net::SocketAddr, time::Duration};
 
 use backon::Retryable;
 use nym_credential_proxy_requests::api::v1::ticketbook::models::PartialVerificationKeysResponse;
 use nym_http_api_client::{ApiClient, HttpClientError, Params, PathSegments, UserAgent, NO_PARAMS};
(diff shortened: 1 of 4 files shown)


nym-vpn-core/crates/nym-vpn-api-client/src/client.rs
```diff
@@ -150,12 +147,38 @@ impl VpnApiClient {
...
+    ) -> std::result::Result<T, HttpClientError<E>>
+    where
+        T: DeserializeOwned,
+        E: fmt::Display + DeserializeOwned,
+    {
+        match self.get_query(path, account, device, None).await {
+            Ok(response) => Ok(response),
+            Err(err) => {
+                tracing::warn!("Failed query: {err}; retrying query with remote time");
+                if let Ok(Some(jwt)) = self
```
octol: I wonder if it's possible to do this retry only if the failure was due to the the JWT error?

neacsu: Ah my bad I didn't put the error unwrapping here as in the other places.
But to respond your question, there are a few things that could make this more specific: the error status seems to be `403 Forbidden` (although I'm not sure if it's for all jwt errors) and also error message seems to contain the string `JWT`, although that can change on the server and also provides some code challenges since the function has generic `E: fmt::Display + DeserializeOwned`, and trying to tie `error` would need it a concrete type.
I figured retrying doesn't hurt anyway, even with the more generic `EndpointFailure` matching, but we could go however deep we want (maybe even introducing a new `HttpClientError` type specifically for JWT?)

neacsu: I did error message parse in the end, looking for `JWT` string as an indicative for the JWT related errors

octol (approved):

PR #2576, opened 2025-03-27
Register entry and exit wireguard gateways in parallel

octol:
Register with entry and exit gateway in parallel to speed up time-to-connect.

Create a mixnet listener that broadcasts the mixnet messages to the auth clients

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2576)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -3331,7 +3331,6 @@ dependencies = [
 [[package]]
 name = "nym-api-requests"
 version = "0.1.0"
-source = "git+https://github.com/nymtech/nym?branch=release%2F2025.5-chokito#0f6ec8610eb34177f68693795b6eb83602c09f81"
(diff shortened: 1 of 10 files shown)


jmwample (approved):
Tested with vpnd and vpnc following logs and this accomplishes the goal of parallel registration with the node wireguard authenticators :+1: . I will defer to others on the specifics.

Would it be good to add support for something like this to the `MixnetClient` itself in the core?

octol:
> Tested with vpnd and vpnc following logs and this accomplishes the goal of parallel registration with the node wireguard authenticators 👍 . I will defer to others on the specifics.
> 
> Would it be good to add support for something like this to the `MixnetClient` itself in the core?

Not just add something like it to `MixnetClient`, currently it's just taking a plain payload. Having a application layer protocol ontop with proper stream multiplexing I think is something we should work towards in the future

PR #2596, opened 2025-04-01
Temporary check for double import via error

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2596)


octol (approved):
Looking good!

PR #2627, opened 2025-04-07
APPLE: rhubarb core

rokas-ambrazevicius:
Use older version for grpc-swift and swift-protobuf, due to grpc-swift migration and minimum support target of macOS 15.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2627)


.github/workflows/build-nym-vpn-core-mac.yml
```diff
@@ -45,8 +45,11 @@ jobs:
 
       - name: Install swift-protobuf
         run: |
-          brew install swift-protobuf
-          brew install grpc-swift
+          export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+          curl -O https://raw.githubusercontent.com/Homebrew/homebrew-core/7186d42bc60247d6b5e860348db7402f12bca912/Formula/s/swift-protobuf.rb
```
pronebird, line 49: Can we not brew install older version?

pronebird (changes requested):

.github/workflows/build-nym-vpn-core-mac.yml
pronebird: After a quick googling, I guess we have no other way to handle this.

pronebird (approved):

.github/workflows/build-nym-vpn-core-mac.yml
rokas-ambrazevicius: yes, brew is not that simple when it comes to versioning :/ We have to jump through hoops.

PR #2644, opened 2025-04-09
Migrate to dispatch2, decomission nym-apple-dispatch

pronebird:
Since the improvements in https://github.com/madsmtm/objc2/pull/681 have been fully absorbed (and beyond), we can switch to using `dispatch2` and decomission `nym-apple-dispatch`. This improvement not only removes our own bindings to `libdispatch` but also improves dependency tree as `dispatch2` is used across all other `objc2-*` crates.

Unfortunately, there is no official release yet for that so this PR anchors relevant crates to the commit hash.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2644)


octol (approved):

octol (changes requested):
Just make sure to update deny.toml if that error is something we are ok with

rokas-ambrazevicius (approved):

pronebird:
@octol addressed that

octol (approved):

PR #2658, opened 2025-04-11
Add timeout to waiting for the auth client listener to stop

octol:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658)

### DIFF
--- a/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs
+++ b/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs
@@ -1,7 +1,7 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use futures::StreamExt;
 use nym_mixnet_client::SharedMixnetClient;
@@ -126,11 +126,9 @@ impl AuthClientMixnetListenerHandle {
         self.message_broadcast.subscribe()
     }
 
-    pub async fn cancel(self) -> AuthClientMixnetListener {
+    pub async fn cancel(mut self) -> AuthClientMixnetListener {
         self.cancel_token.cancel();
-        if let Err(err) = self.handle.await {
-            tracing::error!("Error while waiting for auth clients mixnet listener to stop: {err}");
-        }
+        self.wait().await;
         AuthClientMixnetListener {
             mixnet_client: self.mixnet_client,
             message_broadcast: self.message_broadcast,
@@ -139,9 +137,17 @@ impl AuthClientMixnetListenerHandle {
         }
     }
 
-    pub async fn wait(self) {
-        if let Err(err) = self.handle.await {
-            tracing::error!("Error while waiting for auth clients mixnet listener to stop: {err}");
+    pub async fn wait(&mut self) {
+        tokio::select! {
+            join_result = &mut self.handle => {
+                if let Err(err) = join_result {
+                    tracing::error!("Error waiting for auth clients mixnet listener to stop: {err}");
+                }
+            }
+            _ = tokio::time::sleep(Duration::from_secs(5)) => {
+                tracing::error!("Timeout waiting for auth clients mixnet listener to stop. Forcing stop");
+                self.handle.abort();
+            }
         }
     }
 }
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/wireguard/connected_tunnel/desktop.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/wireguard/connected_tunnel/desktop.rs
@@ -472,7 +472,7 @@ impl TunnelHandle {
     /// Wait until the tunnel finished execution.
     ///
     /// Returns a tombstone containing the no longer used tunnel devices and wireguard tunnels (on Windows).
-    pub async fn wait(self) -> Result<Tombstone, JoinError> {
+    pub async fn wait(mut self) -> Result<Tombstone, JoinError> {
         if let Err(e) = self.bandwidth_controller_handle.await {
             tracing::error!("Failed to join on bandwidth controller: {}", e);
         }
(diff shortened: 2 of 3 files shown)


neacsu (commented):
Wouldn't `tokio::time::timeout` on the handle be simpler?

octol:
> Wouldn't `tokio::time::timeout` on the handle be simpler?

The challenge then is to able to call `abort()` on the handle, since it's moved into `timeout()`

neacsu:
> > Wouldn't `tokio::time::timeout` on the handle be simpler?
> 
> The challenge then is to able to call `abort()` on the handle, since it's moved into `timeout()`

But a match on the result to figure out if it was due to timeout wouldn't be equivalent to those `select!` arms?

octol:
> > > Wouldn't `tokio::time::timeout` on the handle be simpler?
> > 
> > 
> > The challenge then is to able to call `abort()` on the handle, since it's moved into `timeout()`
> 
> But a match on the result to figure out if it was due to timeout wouldn't be equivalent to those `select!` arms?

The difference is the select takes it as a mutable ref, so you can still call abort in the other select arm if the first one hasn't fired

neacsu (approved):

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs` line 140 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZZeSD6IjGxo_Gok2G:-ONZZeSD6IjGxo_Gok2H:b-mjh2ju) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs#L140)):*
> ```XML
>     }
> 
>     pub async fn wait(&mut self) {
> ```

This doesn't look right. `wait()` is not consuming meaning that it can be called multiple times which would yield different results. `cancel()` also behaves like `wait()`

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658)* status: 0 of 3 files reviewed, 4 unresolved discussions

___
*[`nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs` line 62 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZcDHy0VMiVMlHkPFu:-ONZcDHy0VMiVMlHkPFv:b-toia5c) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs#L62)):*
> ```XML
>         loop {
>             tokio::select! {
>                 _ = self.external_cancel_token.cancelled() => {
> ```

Prefer `internal_cancel_token = external_cancel_token.child_token()` then you don't need to poll two separate tokens AFAIK
___
*[`nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs` line 92 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZbK9wBB6izyJgOyi4:-ONZbK9wBB6izyJgOyi5:b-coxqxx) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs#L92)):*
> ```XML
>         let message_broadcast = self.message_broadcast.clone();
>         let external_canel_token = self.external_cancel_token.clone();
>         let cancel_token = self.internal_cancel_token.clone();
> ```

This could be `external_cancel_token.child_token()` which would:

a. listen for cancellations from parent
b. can be cancelled but will not propagate cancellation to the parent
___
*[`nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs` line 135 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZamFDDO3NxsjquzVr:-ONZamFDDO3NxsjquzVs:bkhd3w9) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs#L135)):*
> ```XML
>             mixnet_client: self.mixnet_client,
>             message_broadcast: self.message_broadcast,
>             external_cancel_token: self.external_canel_token,
> ```

Reusing a cancelled external token may have side effects such as

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658)* status: 0 of 3 files reviewed, 4 unresolved discussions

___
*[`nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs` line 135 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZamFDDO3NxsjquzVr:-ONZcYcTEbb0zUsjdMB6:bpyipvh) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-authenticator-client/src/mixnet_listener.rs#L135)):*

Reusing a cancelled external token may have side effects such as
</blockquote></details>

such as well it can always be cancelled and the next instance will quit immediately

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658)* status: 0 of 3 files reviewed, 5 unresolved discussions

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/wireguard/connected_tunnel/desktop.rs` line 483 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2658#-ONZcqVkATOcC0LMMILg:-ONZcqVkATOcC0LMMILh:b-tsoguq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/0089ae53d41a3871dcf84eed98516e0517b039ea/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/wireguard/connected_tunnel/desktop.rs#L483)):*
> ```XML
>         // cancel_token should already be cancelled by the time we reach this point.
>         // We just need to wait for the task to finish.
>         self.auth_client_mixnet_listener_handle.wait().await;
> ```

Cancelled by whom? Is it connected to the `self.shutdown_token` somehow?

PR #2681, opened 2025-04-17
Check for offline state in background tasks

octol:
In background tasks launched by the account controller, check offline state when trying to reach to the vpn-api

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2681)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/config.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/config.rs
@@ -1,0 +1,34 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: GPL-3.0-only
+
+use std::path::PathBuf;
+
+use nym_sdk::UserAgent;
+use nym_vpn_network_config::Network;
+
+pub struct AccountControllerConfig {
+    // The data directory where we store the account and device keys.
+    pub data_dir: PathBuf,
+
+    // User agent used by api client.
+    pub user_agent: UserAgent,
+
+    // Credentials mode is a feature flag that determines if we should automatically request
+    // zk-nyms.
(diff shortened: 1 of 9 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/config.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl AccountControllerConfig {
+    // Determine if the credentials mode is enabled. This is determined by the credentials_mode
+    // field in the config, if it is set. Else the network environment feature flag is used.
+    pub fn background_zk_nym_refresh(&self) -> bool {
+        self.credentials_mode.unwrap_or_else(|| {
+            self.network_env
+                .get_feature_flag_credential_mode()
+                .unwrap_or(false)
```
neacsu, line 31: I'm wondering now, in the beginning, before credentials mode was enabled, we wanted to be conservative and set it to `false` as default, but now, wouldn't `true` as default be the conservative choice? Although this might not necessarily be the place to start it, as probably a lot of other places start with `false` as default still.

neacsu (approved):

nym-vpn-core/crates/nym-vpn-account-controller/src/config.rs
pronebird: That's a good point, for commercial product we always want this to be `true`

octol: Note that this flag primarily controls if the account controller will try to request zknyms in the background and before connecting. It's separate from setting credentials mode when connecting to the entry gateway. Defaulting to true might break canary, which currently doesn't have credentials mode enabled I think

But yes I think we should consider making credential mode for connecting to gateways true by default

PR #2683, opened 2025-04-17
refactor(windows): simplify service command handling in run()

audiano:
- **Windows-specific `run()` function**:
  - Simplified service command handling using tuple-based pattern matching
  - Reduced `if-else` nesting and improved clarity
  - Preserved all TODOs and inline comments

---

### ✅ Notes

- No behavioral changes
- All refactors are internal cleanup only
- PR is safe to review without functional regression concerns

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2683)


pronebird:
Thank you for your interest in contributing to Nym VPN. However I think that our `if-else` solution is more straightforward than a match over the tuple of four. The better solution would be to turn the boolean flags into `enum` and handle that as a subcommand in clap.

PR #2696, opened 2025-04-23
Presume online by offline monitor for desktop

neacsu:
As we do on mobile, we should presume that the state machine is online, before starting to connect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2696)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5320,6 +5320,7 @@ dependencies = [
  "nym-bin-common",
  "nym-credentials-interface",
  "nym-ipc",
+ "nym-offline-monitor",
  "nym-task",
  "nym-validator-client",
  "nym-vpn-account-controller",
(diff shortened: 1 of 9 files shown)


rokas-ambrazevicius (approved):

octol (commented):
Is this tested on reboot? I don't think that assumption holds then

octol:
Does this work?

> Is this tested on reboot? I don't think that assumption holds then

I think last time I tried locally on linux. On service startup at boot time, there is no connectivity, and the account controller is started before the offline monitor so it had to assume that there is no connectivity unless told otherwise

neacsu:
> Does this work?
> 
> > Is this tested on reboot? I don't think that assumption holds then
> 
> I think last time I tried locally on linux. On service startup at boot time, there is no connectivity, and the account controller is started before the offline monitor so it had to assume that there is no connectivity unless told otherwise

With the latest commits we won't assume things on desktop and instead just start the offline monitor earlier, like discussed with @pronebird . But these scenarios are good to be taken into account, let's see if the fix holds with them

pronebird (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, 5 of 7 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2696#-:-OOYOtjR1JeG9iLz540J:bnfp4nl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @neacsu)

PR #2702, opened 2025-04-23
Fix cargo install

pronebird:
Our workflows [fail from time to time](https://github.com/nymtech/nym-vpn-client/actions/runs/14624080640/job/41031239605) because some of binaries have already been installed during the previous run.

Our current duct-tape approach is to ignore errors from `cargo install` by using `|| true` which may ignore legit errors. Let's use a proper github action that can tell if dependency is already installed and skip installation if not needed, but then also error when legit errors happen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2702)

### DIFF
--- a/.github/workflows/build-nym-vpn-android.yml
+++ b/.github/workflows/build-nym-vpn-android.yml
@@ -116,9 +116,15 @@ jobs:
           toolchain: 1.85 # If this changes, we need a PR to Fdroid to update pipeline for reproducibility
           targets: aarch64-linux-android
 
-      - name: Install cargo deps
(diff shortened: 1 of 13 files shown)


pronebird (commented):
Self-reviewed for sanity

neacsu (approved):
Watch out for the two extra commits from other branches

pronebird:
> Watch out for the two extra commits from other branches

Fixed, repointed to release branch. Will cherry-pick to `develop` later on

PR #2703, opened 2025-04-23
APPLE: add localizations

rokas-ambrazevicius:
Add localization support

https://github.com/user-attachments/assets/915ac76d-8fe5-4763-a054-2145292148a6

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2703)

### DIFF
--- a/nym-vpn-apple/Daemon/Info.plist
+++ b/nym-vpn-apple/Daemon/Info.plist
@@ -11,7 +11,7 @@
 	<key>CFBundleShortVersionString</key>
 	<string>1.8.0</string>
 	<key>CFBundleVersion</key>
-	<string>47</string>
+	<string>49</string>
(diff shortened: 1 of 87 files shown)


pronebird:
This way of localizing the app will not apply automatically to numbers, currencies, date formats, localizable notifications etc... As I previously pointed out the proper way to do that is to use a built-in way to change the UI language as described in https://developer.apple.com/news/?id=u2cfuj88

pronebird (changes requested):
I think we should take a step back and re-evaluate whether this is the best way to go about that.

rokas-ambrazevicius:
> I think we should take a step back and re-evaluate whether this is the best way to go about that.

No worries, let's discuss this in standup tomorrow.

PR #2827, opened 2025-05-21
Fix dns resolver hang

neacsu:
JIRA-VPN-3381

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)


pronebird (changes requested):
> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @octol)

___
*[`nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs` line 72 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827#-ORBRHgF5nWP_8KRzWZw:-ORBRHgF5nWP_8KRzWZx:b-sd0ybu) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8edb8c9985e2b2650c2099427dcfc797c70f09d2/nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs#L72)):*
> ```XML
>                             Ok(false) => {
>                                 tracing::error!("Inconsistent network");
>                                 self.cancel_token.cancel();
> ```

This makes no sense. Cancelling this token is a no-op. Just return from the loop to bail.

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @octol and @pronebird)

___
*[`nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs` line 72 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827#-ORBRHgF5nWP_8KRzWZw:-ORBZVkX6r73iRE28-pV:bfjkt5n) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8edb8c9985e2b2650c2099427dcfc797c70f09d2/nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs#L72)):*

This makes no sense. Cancelling this token is a no-op. Just return from the loop to bail.
</blockquote></details>

Until this PR, we would presumably not start any work if there was inconsistency. Given that we put this in a background thread, shouldn't we keep this behaviour and signal the other threads to stop, on that inconsistent case? I've replaced the line where this token was a child token to being a clone, so I would think that makes the `cancel` call stop the other threads, not a no-op

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @neacsu and @octol)

___
*[`nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs` line 72 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827#-ORBRHgF5nWP_8KRzWZw:-ORBcSTF07fQ0_MdARvL:b-tvxlo2) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8edb8c9985e2b2650c2099427dcfc797c70f09d2/nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs#L72)):*

Until this PR, we would presumably not start any work if there was inconsistency. Given that we put this in a background thread, shouldn't we keep this behaviour and signal the other threads to stop, on that inconsistent case? I've replaced the line where this token was a child token to being a clone, so I would think that makes the `cancel` call stop the other threads, not a no-op
</blockquote></details>

I see. Well this sort of interior mutability could create complex relationships between objects. Instead we should perhaps tell the parent (supervisor) component that something went terribly wrong and let it decide how to go about it. Understandably, it will probably do the same thing but at least this decision will come from the top.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @neacsu and @octol)

___
*[`nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs` line 72 at r2](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827#-ORBRHgF5nWP_8KRzWZw:-ORBe5U277z8ErfXS49s:b2jql9z) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8edb8c9985e2b2650c2099427dcfc797c70f09d2/nym-vpn-core/crates/nym-vpn-network-config/src/refresh.rs#L72)):*

I see. Well this sort of interior mutability could create complex relationships between objects. Instead we should perhaps tell the parent (supervisor) component that something went terribly wrong and let it decide how to go about it. Understandably, it will probably do the same thing but at least this decision will come from the top.
</blockquote></details>

I have created a PR to ensure that we don't clone cancellation tokens uncontrollably and instead use child tokens:
https://github.com/nymtech/nym-vpn-client/pull/2845

pronebird (commented):
> Reviewed 3 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827)* status: all files reviewed, 1 unresolved discussion (waiting on @neacsu and @octol)

pronebird (approved):
:lgtm:

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2827#-:-ORCMeQ34yxw7FDDfBH1:bnfp4nl)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @neacsu and @octol)

PR #2853, opened 2025-05-27
feat(tauri-ui): updater flow windows

doums:
One-click update flow (Windows only)

- add UI flow
- live download and install progress
- rust updater API tweaks
- app restarts automatically after upgrading
- 0 user interaction needed (apart from the initial click on "Update")
- update NSIS template to tauri lastest (restored previous custom logic)
- update NSIS template and add required vpnd uninstall call when upgrading/reinstall
- fixed install flow without uninstalling previous install
- fixed `dev:browser` run mode, updated dev setup

### misc
thx to latest nsis tauri template, now the installer UI supports HiDPI res

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853)

### DIFF
--- a/nym-vpn-app/eslint.config.js
+++ b/nym-vpn-app/eslint.config.js
@@ -56,6 +56,7 @@ export default [
           allowSeparatedGroups: true,
         },
       ],
+      'import/no-unresolved': 0,
       'react/react-in-jsx-scope': 0,
       '@typescript-eslint/no-floating-promises': 0,
       '@typescript-eslint/prefer-nullish-coalescing': 0,
--- a/nym-vpn-app/index.html
+++ b/nym-vpn-app/index.html
@@ -45,6 +45,7 @@
     <!-- ----------- SPLASH SCREEN ----------- -->
 
     <div id="root" class="h-full font-sans"></div>
+    <script type="module" src="/src/dev.ts"></script>
     <script type="module" src="/src/main.tsx"></script>
   </body>
 </html>
--- a/nym-vpn-app/package-lock.json
+++ b/nym-vpn-app/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.0.0",
       "license": "GPL-3.0-only",
       "dependencies": {
+        "@base-ui-components/react": "^1.0.0-alpha.8",
         "@headlessui/react": "^2.1.2",
         "@radix-ui/react-accordion": "^1.2.3",
         "@radix-ui/react-slider": "^1.2.2",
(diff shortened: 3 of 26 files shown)


rokas-ambrazevicius (approved):

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853)* status: 0 of 24 files reviewed, 1 unresolved discussion

___
*[`nym-vpn-app/src-tauri/src/commands/updater.rs` line 83 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853#-OS-2ejgBlxHsVFSua4m:-OS-2ejgBlxHsVFSua4n:bg4u45f) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/9f325ea8ac27a837ca4505583e44d5509d957ec9/nym-vpn-app/src-tauri/src/commands/updater.rs#L83)):*
> ```XML
>                     );
>                     let _ = on_event.send(DownloadUpdateEvent::Started {
>                         content_length: content_length.unwrap_or(20_000_000), // default to 20MB
> ```

I assume that `content_length` can only be `None` if the CDN does not return it in the header. Seems a bit arbitrary to default to 20mb. Not sure, but maybe we can show progress as "indetermined" instead which should never happen anyway right?

doums:
@pronebird 20MB is roughly the size of the setup.exe artifacts. So it is a pretty decent fallback value in case of `content_length` is `None`.
As you said it is very unlikely to happen anyway.

Thus, this is a simple way to handle it. If the real size is more or less, it is not like that would break or hurt something.

Not sure what changes you are requesting?

pronebird (commented):
Looking at the UI, since we only display % completion at the moment, perhaps leave it as is for now.

p.s.: Something I had in mind was to show "undetermined" progress bar with number of downloaded bytes.

> Reviewed 1 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853#-:-OS-FmRnBkZNpSb03ZQq:b-jertis)* status: 1 of 24 files reviewed, all discussions resolved

pronebird (commented):
> Reviewed 11 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853)* status: 12 of 24 files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> Reviewed 12 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2853#-:-OS-IJFi3CFMxnnlykUm:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @doums)

doums:
> p.s.: Something I had in mind was to show "undetermined" progress bar with number of downloaded bytes.

let's keep it simple

PR #2695, opened 2025-04-23
Add data-testids for test automation

tommyv1987:
- we will perform automation testing using tauri driver and webdriverio for simple ui automation
- adding the data-testid will help this

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2695)

PR #2856, opened 2025-05-28
fix(tauri): data test attributes clashes

doums:
Some recently added (#2695) data attribute for tests purposes conflict with the ones used and exposed by UI component libraries (Radix, Headless UI etc). This result in broken style and wrong behaviors because those test attributes override the values.
To solve this, this PR put all test attributes behind a specialized prefix `data-test-*`.

`data-testid` remains as is, which should be safe to use.

NOTE: QA tests would need to be updated accordingly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2856)

### DIFF
--- a/nym-vpn-app/src/screens/StartupError.tsx
+++ b/nym-vpn-app/src/screens/StartupError.tsx
@@ -40,7 +40,7 @@ function StartupError({
     <div
       className={clsx([theme === 'dark' && 'dark', 'h-full'])}
       data-testid="startup-error-container"
-      data-theme={theme}
+      data-test-theme={theme}
     >
       <div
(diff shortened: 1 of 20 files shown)


tommyv1987 (commented):
This looks good

pronebird (approved):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2856)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @doums)

tommyv1987 (approved):

PR #2875, opened 2025-06-03
Adopt sqlite pool guard

pronebird:
In sympathy with https://github.com/nymtech/nym/pull/5796 this change also switches sqlite connections pools to the `SqlitePoolGuard` to ensure that database file is fully closed before proceeding.

JIRA-VPN-1471

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2875)


pronebird:
Let's make it ready for merge given that we want to test this work before merging the same change in nym platform.

neacsu (approved):

PR #2904, opened 2025-06-10
Add defaults networks for sandbox and canary

pronebird:
- Add defaults for sandbox and canary
- ~~Remove qa which I believe is no longer available~~
- Drop codegen in favor of runtime json parser which is really fast. Maybe I am exaggerating the complexity of `build.rs`? Alternatives?

JIRA-VPN-3511

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2904)


neacsu (approved):

rokas-ambrazevicius (approved):

pronebird:
I couldn't remove QA environment since backend still returns it along with all other envs. Let's ignore it for now then as I don't think we test in it.

PR #2942, opened 2025-06-15
build(tauri): fix workflow

doums:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2942)

### DIFF
--- a/.github/workflows/publish-nym-vpn-app.yml
+++ b/.github/workflows/publish-nym-vpn-app.yml
@@ -77,7 +77,6 @@ jobs:
       version: ${{ steps.cargo-get.outputs.metadata }}
       win_updater_exe: ${{ steps.win_updater.outputs.exe }}
       win_updater_sig: ${{ steps.win_updater.outputs.sig }}
-      core_link: ${{ steps.core_link.outputs.link }}
     env:
       # stable | nightly | dev
       RELEASE_TYPE: ${{ inputs.release_type || (github.event_name == 'schedule' && 'nightly' || 'dev') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-nym-vpn-app.yml
```diff
@@ -258,11 +241,22 @@ jobs:
...
           echo "TOPIC=$topic" >> "$GITHUB_ENV"
 
+      - name: Set vpn-core link
+        run: |
+          if [[ -n "${{ inputs.core_release_tag }}" ]]; then
+            url=https://github.com/nymtech/nym-vpn-client/releases/tag/${{ inputs.core_release_tag }}
+          elif [[ -n "${{ inputs.core_artifact_url }}" ]]; then
+            url="$(dirname ${{ inputs.core_artifact_url }})/"
+          else
+            url="https://${{ secrets.CI_WWW_REMOTE_HOST }}/nym-vpn-client/nym-vpn-core/develop/"
```
pronebird, line 251: I am not sure in which case the `else` branch is going to be evaluated, however should it be `develop` or current branch? (`${{ github.ref_name }}`)

doums: The workflow is scheduled to run daily for nightly builds, so without inputs. Hence, the "else" branch which set the url to the develop channel. Can serve as fallback value too.

doums: I'm not sure if it is a good idea to use the current branch here (`${{ github.ref_name }}`)? As the corresponding directory in the builds' website should exist and matches.

pronebird: Fine. I actually have no idea how it's wired up on build server.

pronebird (approved):

PR #2957, opened 2025-06-17
Update rust2go to a833a4a

pronebird:
There is a patch in rust2go that hasn't been released yet and it seems to be fixing some memory issues. 

@durch could you please test it and see if that helps with some of crashes that you've been seeing lately?

Commit: https://github.com/ihciah/rust2go/commit/a833a4a103e0ea77684cbf9cc84320d049394301

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2957)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -6442,7 +6442,7 @@ dependencies = [
 [[package]]
 name = "rust2go"
 version = "0.4.1"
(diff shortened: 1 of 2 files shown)


tommyv1987:
This didn't fix our probe issue but might, be worth changing anyway

pronebird:
@tommyv1987 let's merge it then.

PR #2972, opened 2025-06-18
Run daemon in debug to collect more useful logs

pronebird:
Let's enable debug for the daemon because info level logs are rarely all that useful. We'll need to silence some of our crates to reduce the noise but this PR is the first step.

JIRA-VPN-3560

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2972)

### DIFF
--- a/nym-vpn-apple/Daemon/net.nymtech.vpn.helper.plist
+++ b/nym-vpn-apple/Daemon/net.nymtech.vpn.helper.plist
@@ -7,6 +7,7 @@
 	<key>ProgramArguments</key>
 	<array>
 		<string>Contents/MacOS/net.nymtech.vpn.helper</string>
+		<string>-v</string>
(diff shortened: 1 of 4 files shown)


doums (approved):

doums:
What about the Windows service? It is running at INFO level if I'm not wrong.
For ref https://github.com/nymtech/nym-vpn-client/blob/e6d7024803b074cc8f8d134386433c6730d2bf09/nym-vpn-app/src-tauri/bundle/windows/installer.nsi#L717

pronebird:
> What about the Windows service? It is running at INFO level if I'm not wrong.

I have added `-v` to windows service registration, is it not enough?

```
vec![OsString::from("-v"), OsString::from("--run-as-service")],
```

PR #2999, opened 2025-06-23
gh: fix deb workflow

pronebird:
When both Linux and deb core build workflows build wireguard-go-linux they attempt to overwrite the same artifact which causes issues on CI. Let's make a workflow identical to the one used for linux.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/2999)

### DIFF
--- a/.github/workflows/build-nym-vpn-core-deb.yml
+++ b/.github/workflows/build-nym-vpn-core-deb.yml
@@ -13,6 +13,8 @@ env:
 jobs:
   build-wireguard-go-linux:
     uses: ./.github/workflows/build-wireguard-go-linux.yml
+    with:
+      artifact_name: wireguard-go_deb_x86_64
 
   build-linux:
     needs: build-wireguard-go-linux
@@ -61,7 +63,7 @@ jobs:
       - name: Download wireguard-go artifacts
         uses: actions/download-artifact@v4
         with:
-          name: wireguard-go_ubuntu-22.04_x86_64
+          name: wireguard-go_deb_x86_64
           path: build/lib/x86_64-unknown-linux-gnu
 
       - name: Build nym-vpn-core debian packages
--- a/.github/workflows/build-wireguard-go-linux.yml
+++ b/.github/workflows/build-wireguard-go-linux.yml
@@ -2,6 +2,10 @@ name: build-wireguard-go-linux
 on:
   workflow_dispatch:
   workflow_call:
+    inputs:
+      artifact_name:
+        required: false
+        type: string
   pull_request:
     paths:
       - "wireguard/**"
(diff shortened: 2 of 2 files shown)


.github/workflows/build-wireguard-go-deb.yml
doums, line 1: Cloning a whole workflow for the sake of uploading the same artifact but with a different name seems overkill.
Why not adding an input to build-wireguard-go-linux.yml to allow artifact name customization (on workflow_call)?

pronebird: Ok fine. Let's go with workflow_call customization.

.github/workflows/build-nym-vpn-core-deb.yml
```diff
@@ -11,11 +11,13 @@ env:
   UPLOAD_DIR_DEB: deb_artifacts
 
 jobs:
-  build-wireguard-go-linux:
+  build-wireguard-go-deb:
```
doums: Need to be reverted right?

.github/workflows/build-nym-vpn-core-deb.yml
```diff
@@ -11,11 +11,13 @@ env:
...
     uses: ./.github/workflows/build-wireguard-go-linux.yml
+    with:
+      artifact_name: wireguard-go_deb_x86_64
 
   build-linux:
-    needs: build-wireguard-go-linux
+    needs: build-wireguard-go-deb
```
doums: same here

.github/workflows/build-nym-vpn-core-deb.yml
pronebird: Fixed

.github/workflows/build-nym-vpn-core-deb.yml
pronebird: Fixed

doums (approved):

.github/workflows/build-wireguard-go-linux.yml
```diff
@@ -31,10 +35,19 @@ jobs:
         run: |
           ./wireguard/build-wireguard-go.sh
 
+      - name: Set artifact name
+        id: set_artifact_name
+        run: |
+          if [ -z "${{ inputs.artifact_name }}" ]; then
+            echo "artifact_name=wireguard-go_ubuntu-22.04_x86_64" >> $GITHUB_OUTPUT
+          else
+            echo "artifact_name=${{ inputs.artifact_name }}" >> $GITHUB_OUTPUT
```
doums, line 44: Heads up, if Im not wrong, using outputs requires you setup the job's top level `outputs` property for it to work

doums: In general, I find that outputs are mainly useful if you need to pass the value across jobs
If not, using `$GITHUB_ENV` is the trick, just beware of env var collision and propagation.

pronebird: > Heads up, if Im not wrong, using outputs requires you setup the job's top level `outputs` property for it to work

[Steps context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#steps-context) offers outputs too which can be accessed by step id. This does not require defining `outputs` at job level. Job outputs only need to be defined when the goal is to share the output between jobs.

> In general, I find that outputs are mainly useful if you need to pass the value across jobs If not, using `$GITHUB_ENV` is the trick, just beware of env var collision and propagation.

Both work but `steps.outputs` does not pollute the environment. A bit more verbose perhaps.

pronebird: ps: you can [inspect logs](https://github.com/nymtech/nym-vpn-client/actions/runs/15829426386/job/44618046959?pr=2999#step:7:21) from "upload artifact"  and see that the output is properly picked up. Seems to work as intended.

PR #3001, opened 2025-06-23
flatpak: use xmlstarlet to modify metainfo.xml

pronebird:
This PR migrates metadata update from `sed` to  `xmlstarlet` to modify `metainfo.xml` for flatpak.

Test it with:

```sh
cd nym-vpn-app/.pkg/flatpak

export VERSION=999.9
export MATAINFO_XML=net.nymtech.NymVPN.metainfo.xml
export RELEASE_TAG=nym-vpn-app-999.9

./update_matainfo.sh
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3001)

### DIFF
--- a/.github/workflows/publish-nym-vpn-app.yml
+++ b/.github/workflows/publish-nym-vpn-app.yml
@@ -261,6 +261,10 @@ jobs:
         with:
           ref: 'develop'
 
+      - name: Install deps
+        run: |
+          sudo apt install -y xmlstarlet
+
       - name: Bump installer script
         continue-on-error: true
         working-directory: .pkg/linux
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-nym-vpn-app.yml
```diff
@@ -261,6 +261,10 @@ jobs:
         with:
           ref: 'develop'
 
+      - name: Install deps
+        run: |
+          sudo apt-get install -y xmlstarlet
```
doums: suggestion `apt install`

doums: It requires to install that additional system package, for the same result, so it's not really worth it imo.
But up to you if you want it.

doums (approved):

.github/workflows/publish-nym-vpn-app.yml
pronebird: I agree, however using xml tool feels a bit cleaner than printing into xml. 

Replaced apt-get with apt

PR #3019, opened 2025-07-01
Change stdcall to system calling convention

pronebird:
Based on https://github.com/rust-lang/rust/issues/137018 Rust 1.88 deprecates the `stdcall` convention. `system` should be used instead which will translate to the right convention when using winapi on windows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3019)


RalfJung:
>  Rust 1.88 deprecates the stdcall convention

Not quite. But it should only be used on targets that actually have a [calling convention of that name](https://en.wikipedia.org/wiki/X86_calling_conventions#stdcall) -- which is only x86.

neacsu (approved):

PR #3023, opened 2025-07-03
fix(vpnd): fix sentry guard lifetime

doums:
During a local testing session I spotted that some errors expected to be caught by sentry were not.
After more digging, I figured out what was the culprit: the client guard. Returned by `init()` call, it was somehow not living long enough (i.e. dropped to early). But the issue with the guard lifetime was not very clear.
Thing that I'm sure is yielding it as a `static` variable solved it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023)

### DIFF
--- a/nym-vpn-app/src-tauri/Cargo.toml
+++ b/nym-vpn-app/src-tauri/Cargo.toml
@@ -70,8 +70,8 @@ windows = { version = "0.60.0", features = [
     "Win32_Foundation",
 ] }
 
-[profile.release]
-codegen-units = 1
-opt-level = "z"
-lto = true
-strip = true
+#[profile.release]
+#codegen-units = 1
+#opt-level = "z"
+#lto = true
+#strip = true
--- a/nym-vpn-app/src-tauri/Cargo.toml
+++ b/nym-vpn-app/src-tauri/Cargo.toml
@@ -70,8 +70,8 @@ windows = { version = "0.60.0", features = [
     "Win32_Foundation",
 ] }
 
-#[profile.release]
-#codegen-units = 1
-#opt-level = "z"
-#lto = true
-#strip = true
+[profile.release]
+codegen-units = 1
+opt-level = "z"
+lto = true
+strip = true
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5263,6 +5263,7 @@ dependencies = [
  "nym-vpn-store",
  "nym-vpnd-types",
  "nym-windows",
+ "once_cell",
  "sentry",
  "serde",
  "serde_json",
--- a/nym-vpn-core/crates/nym-vpnd/Cargo.toml
+++ b/nym-vpn-core/crates/nym-vpnd/Cargo.toml
@@ -60,6 +60,7 @@ nym-vpn-network-config = { workspace = true }
 nym-vpn-proto = { workspace = true, features = ["conversions"] }
 nym-vpn-store = { workspace = true }
 nym-vpnd-types = { workspace = true }
+once_cell = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
 windows-service.workspace = true
--- a/nym-vpn-core/crates/nym-vpnd/src/command_interface/connection_handler.rs
+++ b/nym-vpn-core/crates/nym-vpnd/src/command_interface/connection_handler.rs
@@ -17,15 +17,14 @@ use nym_vpn_lib_types::{
 use nym_vpn_network_config::{FeatureFlags, ParsedAccountLinks, SystemMessages};
 use nym_vpnd_types::gateway;
 
+use super::protobuf::error::VpnCommandSendError;
 use crate::{
     logging::LogPath,
     service::{
-        AccountLinksError, ConnectArgs, ConnectOptions, GATEWAY_DIRECTORY_CLIENT, SetNetworkError,
-        VpnServiceCommand, VpnServiceDeleteLogFileError,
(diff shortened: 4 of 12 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @mrkiwi)

___
*[`nym-vpn-core/crates/nym-vpnd/src/main.rs` line 35 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023#-OUJp5df1Jo03Pcv7hxT:-OUJp5df1Jo03Pcv7hxU:bh5wxzq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/bd70213ed5df2b14f67db004744ae6f4942397c7/nym-vpn-core/crates/nym-vpnd/src/main.rs#L35)):*
> ```XML
>     let args = CliArgs::parse();
>     let _sentry_guard = init_sentry();
>     let sentry_enabled = _sentry_guard.is_some_and(|client| client.is_enabled());
> ```

This line of code consumed `_sentry_guard` and dropped it right after due to use of consuming `is_some_and()`. So what you want is to rewrite it with `as_ref()`:

```rs
let sentry_enabled = _sentry_guard
        .as_ref()
        .is_some_and(|client| client.is_enabled());
```
___
*[`nym-vpn-core/crates/nym-vpnd/src/main.rs` line 53 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023#-OUJpctP3PoqjyW_Qfrs:-OUJpctP3PoqjyW_Qfrt:bdn314x) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/bd70213ed5df2b14f67db004744ae6f4942397c7/nym-vpn-core/crates/nym-vpnd/src/main.rs#L53)):*
> ```XML
>     let args = CliArgs::parse();
>     let _sentry_guard = init_sentry();
>     let sentry_enabled = _sentry_guard.is_some_and(|client| client.is_enabled());
> ```

Same here, use `as_ref()` instead and that should keep sentry alive until the end of this scope. With static it would live forever.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @doums and @mrkiwi)

___
*[`nym-vpn-core/crates/nym-vpnd/src/main.rs` line 27 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023#-OUJqwBTAhWqNU-9Tt50:-OUJqwBTAhWqNU-9Tt51:b-i6fw9) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/8286c9be3550ae2fb2164bb7b68032846b652ad7/nym-vpn-core/crates/nym-vpnd/src/main.rs#L27)):*
> ```XML
> use crate::{cli::CliArgs, config::GlobalConfigFile};
> 
> static SENTRY_CLIENT: OnceCell<ClientInitGuard> = OnceCell::new();
> ```

FYI  although I don't think we need to use statics, just so you know`OnceCell` is now a part of std. https://doc.rust-lang.org/std/cell/struct.OnceCell.html

pronebird (commented):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @doums and @mrkiwi)

doums:
Thanks for the explanations
Still I think holding the client guard into a global static variable is useful:
It is clear where and how the guard is yielded. Much better than having it stored in a local variable inside some random function (as you explained, I just got bitten by that).
We can have access to it from everywhere (e.g., in future potential work around sentry).
Using a static variable, the guard is more safely stored.

pronebird (commented):
I see the frustration. I suppose for the same reason someone decided to return `Option<WorkerGuard>` from `run()` to prevent that exact scenario.

Global variables (or statics) are really an anti-pattern and we should avoid them when possible. The lifetime of globals is static so `Drop` never happens for anything held in static, which can be pretty bad if the type held in static expects to exit cleanly.

`main()` is the highest level scope which should hold the top level objects like Sentry that will live throughout the application lifecycle.

If we ever need to access the instance of Sentry, we should consider adding a manager type around it and pass it down as a dependency using `Arc<Mutex<SentryManager>>`. This should also provide better isolation and less tightly coupled program.

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3023#-:-OUKsFE35qrLk7x5kMiy:b-k9i0ob)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @doums and @mrkiwi)

doums:
I don't agree with that. Static variable are not an anti-pattern. When used correctly. As you said, OnceCell even landed into the std.
When you want to store a variable you know that should live long like the whole program lifetime, you want it immutable, and you want to have easy access to, it is perfectly fine.
Also in case of OneCell you can `take` then drop the value whenever you want.

For stuff like global "singleton" instance I find it far more efficient compared to the classic Rust "iodomtic" way: having to store it from the most top common level (usually `main` or close to it) and then passing it by ref down the level tree down to the lowest level, dealing with all the hell this usually result with.

In this specific case, storing the sentry instance using OneCell is perfectly fine (and valid).

doums:
this branch is fucked re-open in #3027

PR #3034, opened 2025-07-08
Add RPC client

pronebird:
## High level overview of changes

This PR introduces a high level RPC client making gRPC an implementation detail. The RPC client is placed inside `nym-vpn-proto` and feature-gated with `rpc_client`.

The point of RPC client is to provide a communication channel with the daemon leveraging the type system defined by the daemon and the core lib.

### Protobuf changes

- Consolidate all protobuf interaction within `nym_vpn_proto`
- The protobuf types have been moved under `nym_vpn_proto::proto` for the purpose of moving them from the root and providing meaningful namespace for conversions to easier distinguish between proto and daemon/lib types.
- Protobuf files have been consolidated into `nym_vpn_service.proto` under `nym-vpn-core/crates/nym-vpn-proto/proto` since those are implementation detail of this crate.
- `NymVpnd` in protobuf was renamed to `NymVpnService`
- `VpnConnect` and `VpnDisconnect` have been renamed to `ConnectTunnel` and `DisconnectTunnel`
- Superficial request/response type carrying single primitive (like bool or string) have been reduced to primitive.
- Removed "unspecified", "unregistered" etc.. enum variants never used anywhere.

Note that `nym-vpn-proto` has not been renamed yet to keep this changeset slightly smaller. However it's worth to consider renaming it to either `nym-vpn-rpc` or `nym-vpn-management-interface` or alike. `CommandInterface` from `nym-vpnd` can be moved here down the road too but the changeset is already large enough, so let's leave this exercise for later.

### `CommandInterface` changes

`CommandInterface` implements `NymVpnService` taking protobuf representation and translating it to Nym type system. Protobuf basically only exist in here. Common type conversions between protobuf and Nym types are implemented under `nym-vpn-proto/conversions`

### Shared types

Currently all of the types that are passed through RPC layer reside in `nym-vpnd-types` and `nym-vpn-lib-types` to make them reusable. I think that going forward we could unite both. However they are kept separate for now, since a lot of daemon types aren't used in `nym-vpn-lib`.

### Cross-boundary types

There are still plenty of cross-boundary types coming from other crates that we share here and there. The solution I think of is to re-export them from `-types` crate to make it easier to get a hold of them. Currently `nym-vpn-lib` re-exports some of the types but perhaps we should consider moving those declarations to `nym-vpn-lib-types` or  `nym-vpn-lib-types` instead.

### Account controller types

There are a lot of types that should not be a part of RPC. For example account controller tends to dump its entire internal state over gRPC, which is not always easy to achieve in bidirectional fashion but also unnecessary. This should be reduced to remove internal implementation details especially everything around zknym.

### Application level errors

Some of commands no longer return errors or pass them using `tonic::Status`. That's because everything tunnel related should go over the tunnel state. We do still persist selected entry/exit points on disc in call to connect. However this is likely going into separate RPC call to avoid passing the entire config in one go but also to enable dynamic reconfiguration of the tunnel in the future.

JIRA-VPN-2747

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)


neacsu (changes requested):
> Reviewed 76 of 76 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: all files reviewed, 2 unresolved discussions (waiting on @doums, @GG-NYM, @rokas-ambrazevicius, and @simonwicky)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 745 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVMSmP186cwGQLKL5Wm:-OVMSmP186cwGQLKL5Wn:b-a7awl8) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L745)):*
> ```XML
>     }
> 
>     async fn handle_disconnect(&mut self) -> bool {
> ```

Not clear what `bool` returned value represents here. Also, making some handles return `Result`s and others `bool`s seems inconsistent, given both try to represent the same thing (possibility of error) 
___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 1000 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVMSRG-0tCfl1wwdmK-:-OVMSRG-0tCfl1wwdmK0:b-snphu0) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L1000)):*
> ```XML
>     }
> 
>     async fn handle_delete_log_file(&self) -> bool {
> ```

It's not clear from function name what the returned `bool` signifies

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: all files reviewed, 2 unresolved discussions (waiting on @doums, @GG-NYM, @neacsu, @rokas-ambrazevicius, and @simonwicky)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 745 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVMSmP186cwGQLKL5Wm:-OVMiDTj7xQ1KVDQnQFL:b-6nsd34) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L745)):*

Not clear what `bool` returned value represents here. Also, making some handles return `Result`s and others `bool`s seems inconsistent, given both try to represent the same thing (possibility of error)
</blockquote></details>

You're right. It's not very useful. We can return void from here. The clients are expected to use tunnel state to monitor what tunnel is doing.

nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -1,33 +1,41 @@
...
 
-use crate::cli::Cli;
-use crate::country::Country;
-use crate::error::BackendError;
-use crate::fs::config::AppConfig;
-use crate::{events::AppHandleEventEmitter, state::SharedAppState};
+use crate::{
+    cli::Cli, country::Country, error::BackendError, events::AppHandleEventEmitter,
+    fs::config::AppConfig, state::SharedAppState,
+};
```
doums, line 38: In the future can you avoid arbitrarily restyling module imports? That would be good.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: all files reviewed, 2 unresolved discussions (waiting on @doums, @GG-NYM, @neacsu, @rokas-ambrazevicius, and @simonwicky)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 1000 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVMSRG-0tCfl1wwdmK-:-OVMi_Th1agZ06aDN7JK:bf2af82) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L1000)):*

It's not clear from function name what the returned `bool` signifies
</blockquote></details>

Perhaps we don't have to report anything back.

nym-vpn-app/src-tauri/src/grpc/client.rs
pronebird: This is automatic with `rustfmt +nightly --config imports_granularity=Crate` in my case. There was no intention but we should adopt it in my opinion import sorting has been around for a very long time (5 years+?) alas still nightly feature.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: 72 of 77 files reviewed, 3 unresolved discussions (waiting on @doums, @GG-NYM, @neacsu, @rokas-ambrazevicius, and @simonwicky)

nym-vpn-app/src-tauri/src/grpc/client.rs
doums: How rustfmt can be run automatically with such args? `+nightly --config imports_granularity=Crate` including a "nighty" (= unstable?) feature

doums: I mean those have to be manually set by you right?

pronebird: So if `rustfmt.toml` is present with `imports_granularity=Crate`, then running `cargo +nightly fmt` achieves that. But in my case VScode is configured to do that automatically. Yes these have to be set manually.

neacsu (approved):
> Reviewed 2 of 2 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: all files reviewed, 2 unresolved discussions (waiting on @doums, @GG-NYM, @rokas-ambrazevicius, and @simonwicky)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 1008 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVNad_h8TDMBahSW31E:-OVNad_h8TDMBahSW31F:b-n4vfou) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L1008)):*
> ```XML
>             }
>             Err(mpsc::error::TrySendError::Closed(_)) => {
>                 tracing::error!("Failed to send command to delete log file: channel is closed");
> ```

Are we sure we don't want to keep printing this error message? This seems like an internal error if it would happen

doums (approved):

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034)* status: all files reviewed, 2 unresolved discussions (waiting on @doums, @GG-NYM, @neacsu, @rokas-ambrazevicius, and @simonwicky)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 1008 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3034#-OVNad_h8TDMBahSW31E:-OVRKNTs6O-mynAZPCeo:b-oiu7d1) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/fc8ecb8cbf31ebd240be45308b009bdcec3c002a/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L1008)):*

Are we sure we don't want to keep printing this error message? This seems like an internal error if it would happen
</blockquote></details>

So this would be internal error for sure. It means that the other end of the channel is gone. We may as well print them for all channels in here but it feels a bit excessive.  Currently it does not look like `LogFileRemover` can quit unless cancellation token is triggered in which case the daemon is going down anyway. Maybe I am wrong on this one.

PR #3051, opened 2025-07-14
Set route mtu in multihop

pronebird:
- macOS, Linux: set route MTU for multi-hop tunnel. Although I am not fully certain, the assumption is that if MTU is not set for the route, the system is free to run PMTUD to determine it, which could probably cause fragmentation issues if it turns out to be lower than tunnel interface MTU.
- Linux: ensure that only default routes `0.0.0.0/0`, `::0/0` and multi-hop exit route bypass the main table.
- Windows: set interface IP configuration specifying: the lowest interface metric (1), disable DAD, router discovery, dhcp, MTU which I assume is basically an equivalent of route MTU on other platforms.
- Set IPv6 address for mixnet tunnel interface.
- Linux, Windows: consult routing table for the MTU suitable for mixnet tunnel instead of hardcoding `1500`. Fallback to `1500` in case of error.
- Libwg: remove a workaround for Go runtime crash, which has been fixed since Go 1.23.6 (Moved to https://github.com/nymtech/nym-vpn-client/pull/3057)

JIRA-VPN-3575

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3051)

### DIFF
--- a/nym-vpn-core/CHANGELOG.md
+++ b/nym-vpn-core/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- sqlx-pool-guard public interface issue (https://github.com/nymtech/nym-vpn-client/pull/2925)
+- Wait for sqlite database to close before exiting the daemon to prevent database corruption (https://github.com/nymtech/nym-vpn-client/pull/2925)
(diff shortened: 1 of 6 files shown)


neacsu (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pronebird)

PR #3065, opened 2025-07-15
bump: tauri next dev

doums:
## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3065)

### DIFF
--- a/nym-vpn-app/src-tauri/Cargo.toml
+++ b/nym-vpn-app/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nym-vpn-app"
-version = "1.12.0-dev"
+version = "1.13.0-dev"
 description = "NymVPN desktop client"
 authors = [
     "Nym Technologies SA",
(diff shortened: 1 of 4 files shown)


rokas-ambrazevicius (approved):

nym-vpn-app/src-tauri/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "nym-vpn-app"
-version = "1.12.0-dev"
+version = "1.13.0-dev"
```
pronebird, line 3: any reason why you don't use `1.13.0-beta` just as the core does? I mean the -beta suffix. Just wondering.

doums: I'm using `-dev` because it is what it is: from the develop branch, I'm producing “development” builds of tauri. It is not specifically "beta", "alpha" or whatever the more specialized build.
If I need a more specialized build, I will update the version accordingly, on-build.

> just as the core does?

Tauri and vpnd are meant to be used together but still are 2 different apps. I'm not sure why core is always using "-beta" from develop branch, does not make much sense to me.

PR #3075, opened 2025-07-17
Add import sorting

pronebird:
## Description

Enable import sorting using crate [granularity](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=group#imports_granularity)

This feature has been around for years and makes imports quite compact. It does require nightly compiler AFAIK but that shouldn't be a problem.

For those using VSCode to format, the following needs to be added to rust-analyzer configuration:

```
"rust-analyzer.rustfmt.extraArgs": ["+nightly"],
```

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3075)

### DIFF
--- a/nym-vpn-core/README.md
+++ b/nym-vpn-core/README.md
@@ -1,5 +1,25 @@
 # Nym VPN Core
 
+## Code formatting
+
+We use some of nightly features of rustfmt to format the codebase. Please install the nightly rust with rustfmt:
+
+```sh
+rustup toolchain install nightly -c rustfmt
+```
+
+Format the code using the following command:
+
+```sh
+cargo +nightly fmt
+```
+
+If you use VSCode and automatic formatting, configure rust-analyzer to use nightly rustfmt:
+
+```json
+"rust-analyzer.rustfmt.extraArgs": ["+nightly"],
+```
+
 ## Prerequisites
 
 ### Linux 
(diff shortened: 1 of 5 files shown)


doums:
As you know, I'm not a partisan to go more and more stricter on code style and purely aesthetic stuff (tastes and colors always remain subjectives).
Don't get me wrong, code style unification is good. And we already have it thank to rustfmt. But not when it involves too much extra tooling to get there.

Here we are talking of adding custom, more stricter rules to rustfmt to enforce module import sorting and style. This will require all dev/contributors to setup locally nightly compiler to comply to those introduced rules. Not mentioning editors/IDE settings (not all devs are using VSCode btw, Im not).
As those module import rules by itself, does not bring any concrete value to the code. As it is purely aesthetic. It is not worth it in my opinion.

We run Rust compiler which already enforce a ton of stuff, Clippy with custom rules, and rustfmt. We don't need more.

pronebird:
I think that one time setup is a small price to pay for having concise and ordered imports. it's objectively easier to inspect imports when they are ordered.

IDE or rustup configuration is not really an obstacle in my opinion. It really boils down to `cargo +nightly fmt` instead of `cargo fmt`.

neacsu (approved):

PR #3134, opened 2025-07-26
Refactor daemon

pronebird:
## Description

1. cli: replace a struct with 4 flags with `enum` with 5 commands to:
   - Install windows service (Windows only)
   - Uninstall windows service (Windows only)
   - Start windows service (Windows only)
   - Run as a system service
   - Run as a standalone process (default, cannot be materialized via CLI argument)
   
   Further thoughts: it would be great to remove windows service installation from the daemon binary to keep code cleaner and improve separation of concern. But for now let it be.
1. Create tokio runtime in `main()` to avoid having mixed code using sync and async code paths.
   *Note: windows service does not instantiate new tokio runtime anymore. Instead it receives a handle to tokio runtime created in `main()` and uses it to spawn async tasks in sync `service_main()` invoked by the operating system*
2. Set up logging in `main()` and avoid passing `WorkerGuard` deep into dependency tree, keep it as close as possible to `main()`.
3. Wrap `UnboundedSender<()>` used for signaling when to remove the active log file into `RemoveLogFileSignal` struct holding a channel internally but hiding it as implementation detail. Minor imporvement that adds clarity.
4. Move `NymVpnService` configuration into `crate::setup_vpn_service()` and reuse it in both `main.rs` and `windows_service/mod.rs`. Eliminates multiple similar code paths and reduces code duplication.
5. Break down lengthy windows service initialization on a few functions
6. Fix clippy too-many-arguments warnings by moving parameters into structs.
7. Move functions used for windows service manipulation into `windows_service/installation.rs`
8. Make `uninstall_service()` async, use `tokio::time()` to sleep instead of `std::thread::sleep()`
9. Windows service: report failure when unable to setup vpn service in initialization sequence (`ServiceSpecificExitCode::SetupVpnService`)
10. Windows service: Pass all of the CLI arguments to windows service
11. Fix too large futures causing stack overflow on Windows. Can be identified by running: `cargo clippy -- -W clippy::large_futures`

This change is not user-facing for the most party so no changelog entry.

## Checklist:

- [x] Changelog

JIRA-VPN-3761

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3134)


pronebird:
I have tested this PR on macOS, Windows running `nym-vpnd` as a standalone app. Tested running `nym-vpnd` as a service on Windows

neacsu (approved):
> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3134)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @doums and @simonwicky)

PR #3211, opened 2025-08-01
Serialize IpPair to bytes

pronebird:
## Description

Serialize `IpPair` as byte array instead of JSON.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3211)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/platform/uniffi_custom_impls.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/platform/uniffi_custom_impls.rs
@@ -106,16 +106,23 @@ uniffi::custom_type!(BoxedRecepient, String, {
 
 uniffi::custom_type!(
     IpPair,
-    String, {
+    Vec<u8>, {
         remote,
         try_lift: |val| {
-            Ok(
-                serde_json::from_str(&val).map_err(|e| VpnError::InternalError {
-                    details: e.to_string(),
-                })?,
-            )
+            let bytes: [u8; 20] = val.try_into().expect("Invalid length for IpPair byte representation");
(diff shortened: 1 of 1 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/platform/uniffi_custom_impls.rs
```diff
@@ -106,16 +106,23 @@ uniffi::custom_type!(BoxedRecepient, String, {
...
         remote,
         try_lift: |val| {
-            Ok(
-                serde_json::from_str(&val).map_err(|e| VpnError::InternalError {
-                    details: e.to_string(),
-                })?,
-            )
+            let bytes: [u8; 20] = val.try_into().expect("Invalid length for IpPair byte representation");
+            let ipv4_bytes: [u8; 4] = bytes[0..4].try_into().unwrap();
+            let ipv6_bytes: [u8; 16] = bytes[4..20].try_into().unwrap();
```
doums, line 114: safe to unwrap here?

pronebird: Since `lower` produces a `Vec<u8>` that is 20 bytes long (ipv4 + ipv6 octets together), I can't imagine a situation where `try_lift` would receive any other value. I thought about that and it seems that this is only possible if uniffi has a bug.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3211)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @doums)

doums (approved):

PR #3220, opened 2025-08-05
Floriane/observability

IRClichtR:
### Debug
- Downgraded network transient errors to debug/info
- Improved Sentry error reporting accuracy

### Changes
- Downgraded network transient errors from errors to info/debug level for better log clarity

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)

### DIFF
--- a/nym-vpn-app/src-tauri/src/commands/account.rs
+++ b/nym-vpn-app/src-tauri/src/commands/account.rs
@@ -1,5 +1,5 @@
 use tauri::State;
-use tracing::{error, info, instrument, warn, debug};
+use tracing::{error, info, instrument, warn};
 
 use crate::grpc::account_links::AccountLinks;
 use crate::grpc::tunnel::TunnelState;
@@ -65,7 +65,15 @@ pub async fn is_account_stored(grpc: State<'_, GrpcClient>) -> Result<bool, Back
     grpc.is_account_stored()
         .await
         .map_err(|e| {
-            debug!("failed to check stored account: {e}");
+            if e.to_string().contains("ConnectionReset")
+                || e.to_string().contains("ConnectionRefused")
+                || e.to_string().contains("Timeout")
+                || e.to_string().contains("NetworkUnreachable")
+            {
+                warn!("account check failed due to transient network issue: {e}");
+            } else {
+                error!("failed to check stored account: {e}");
+            }
             e.into()
         })
         .inspect(|stored| {
--- a/nym-vpn-app/src-tauri/src/commands/account.rs
+++ b/nym-vpn-app/src-tauri/src/commands/account.rs
@@ -1,5 +1,5 @@
 use tauri::State;
-use tracing::{error, info, instrument, warn};
+use tracing::{error, info, instrument, warn, debug};
 
 use crate::grpc::account_links::AccountLinks;
 use crate::grpc::tunnel::TunnelState;
@@ -65,7 +65,7 @@ pub async fn is_account_stored(grpc: State<'_, GrpcClient>) -> Result<bool, Back
     grpc.is_account_stored()
         .await
         .map_err(|e| {
-            error!("failed to check stored account: {e}");
+            debug!("failed to check stored account: {e}");
             e.into()
         })
         .inspect(|stored| {
@@ -80,7 +80,7 @@ pub async fn account_links(
     locale: String,
 ) -> Result<AccountLinks, BackendError> {
     grpc.account_links(&locale).await.map_err(|e| {
-        error!("failed to get account link: {e}");
+        warn!("failed to get account link: {e}");
         e.into()
     })
 }
--- a/nym-vpn-app/src-tauri/src/commands/tunnel.rs
+++ b/nym-vpn-app/src-tauri/src/commands/tunnel.rs
@@ -11,10 +11,7 @@ use crate::{
 use tauri::State;
 use tracing::{debug, error, info, instrument, warn};
 
-#[instrument(
-    skip_all,
-    err(Debug)
-)]
+#[instrument(skip_all, err(Debug))]
 #[tauri::command]
 pub async fn get_tunnel_state(
     app: tauri::AppHandle,
--- a/nym-vpn-app/src-tauri/src/commands/tunnel.rs
+++ b/nym-vpn-app/src-tauri/src/commands/tunnel.rs
@@ -11,7 +11,10 @@ use crate::{
 use tauri::State;
 use tracing::{debug, error, info, instrument, warn};
 
-#[instrument(skip_all)]
+#[instrument(
+    skip_all,
+    err(Debug)
+)]
 #[tauri::command]
 pub async fn get_tunnel_state(
     app: tauri::AppHandle,
--- a/nym-vpn-app/src-tauri/src/commands/updater.rs
+++ b/nym-vpn-app/src-tauri/src/commands/updater.rs
@@ -34,7 +34,7 @@ pub async fn fetch_update(
     pending_update: State<'_, PendingUpdate>,
 ) -> Result<Option<UpdateMetadata>, BackendError> {
     if !*UPDATER_ENABLED {
-        error!("updater is disabled for this build");
+        warn!("updater is disabled for this build");
         return Err(BackendError::internal("updater is disabled", None));
     }
     let update = updater::check(app)
@@ -54,7 +54,6 @@ pub async fn fetch_update(
     }
 }
 
-// TODO: Downgrad log level to info
 // Based on https://v2.tauri.app/plugin/updater/#checking-for-updates
 #[tauri::command]
 #[instrument(skip_all)]
@@ -63,13 +62,13 @@ pub async fn install_update(
     on_event: Channel<DownloadUpdateEvent>,
 ) -> Result<(), BackendError> {
     if !*UPDATER_ENABLED {
-        info!("updater is disabled for this build");
+        warn!("updater is disabled for this build");
         return Err(BackendError::internal("updater is disabled", None));
     }
     let Some(update) = pending_update.0.lock().await.take() else {
         // calling this function without a pending update is an error
-        info!("no update available");
-        return Err(BackendError::internal("no update available", None));
+        info!("no pending update to install");
+        return Err(BackendError::internal("install_update called without a pending update", None));
     };
 
     let mut started = false;
@@ -95,6 +94,6 @@ pub async fn install_update(
             },
         )
         .await
-        .inspect_err(|e| warn!("Update download and install failed: {}", e))?;
+        .inspect_err(|e| error!("Update download and install failed: {}", e))?;
     Ok(())
 }
--- a/nym-vpn-app/src-tauri/src/commands/updater.rs
+++ b/nym-vpn-app/src-tauri/src/commands/updater.rs
@@ -54,6 +54,7 @@ pub async fn fetch_update(
     }
 }
 
+// TODO: Downgrad log level to info
 // Based on https://v2.tauri.app/plugin/updater/#checking-for-updates
 #[tauri::command]
 #[instrument(skip_all)]
@@ -62,12 +63,12 @@ pub async fn install_update(
     on_event: Channel<DownloadUpdateEvent>,
 ) -> Result<(), BackendError> {
     if !*UPDATER_ENABLED {
-        error!("updater is disabled for this build");
+        info!("updater is disabled for this build");
         return Err(BackendError::internal("updater is disabled", None));
     }
     let Some(update) = pending_update.0.lock().await.take() else {
         // calling this function without a pending update is an error
-        error!("no update available");
+        info!("no update available");
         return Err(BackendError::internal("no update available", None));
     };
 
@@ -94,6 +95,6 @@ pub async fn install_update(
             },
         )
         .await
-        .inspect_err(|e| error!("download and install failed: {}", e))?;
+        .inspect_err(|e| warn!("Update download and install failed: {}", e))?;
     Ok(())
 }
--- a/nym-vpn-app/src-tauri/src/commands/updater.rs
+++ b/nym-vpn-app/src-tauri/src/commands/updater.rs
@@ -68,7 +68,10 @@ pub async fn install_update(
     let Some(update) = pending_update.0.lock().await.take() else {
         // calling this function without a pending update is an error
         info!("no pending update to install");
-        return Err(BackendError::internal("install_update called without a pending update", None));
+        return Err(BackendError::internal(
+            "install_update called without a pending update",
+            None,
+        ));
     };
 
     let mut started = false;
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -179,6 +179,7 @@ impl GrpcClient {
         Ok(tunnel)
     }
 
+    // TODO change logging level to info
     /// Watch tunnel state updates and mixnet events
     #[instrument(skip_all)]
     pub async fn watch_tunnel_events(&self, app: &AppHandle) -> Result<()> {
@@ -188,7 +189,7 @@ impl GrpcClient {
             .listen_to_events(())
             .await
             .inspect_err(|e| {
-                error!("listen_to_tunnel_state_changes failed: {}", e);
+                warn!("listen_to_tunnel_state_changes failed: {}", e); // changed level from error to warning
             })?
             .into_inner();
 
@@ -401,6 +402,7 @@ impl GrpcClient {
             locale: "en".to_string(),
         });
         let response = vpnd.get_account_links(request).await.map_err(|e| {
+            // TODO change logging level to info
             error!("grpc: {}", e);
             VpndError::GrpcError(e)
         })?;
--- a/nym-vpn-app/src-tauri/src/grpc/client.rs
+++ b/nym-vpn-app/src-tauri/src/grpc/client.rs
@@ -201,7 +201,7 @@ impl GrpcClient {
                         tx.send(update).await.unwrap();
                     }
                     Ok(None) => {
-                        info!("vpnd DOWN: tunnel state stream closed");
+                        warn!("vpnd DOWN: tunnel state stream closed");
                         VPND_DOWN_LOGGED.store(true, Ordering::Relaxed);
                         return;
                     }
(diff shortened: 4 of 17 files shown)


pronebird (commented):
> Reviewed 1 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 1 of 14 files reviewed, 2 unresolved discussions

___
*[`nym-vpn-core/crates/nym-statistics/src/storage/mod.rs` line 69 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220#-OWuxJiA6xQo5siGE8oo:-OWuxJiA6xQo5siGE8op:b-odlkwp) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f7181b962a40d21788fd3da09abd3d7c23516ba0/nym-vpn-core/crates/nym-statistics/src/storage/mod.rs#L69)):*
> ```XML
>             tracing::debug!("Stats storage closed");
>         }
>         .instrument(tracing::debug_span!("close_stats_storage"))
> ```

Nit: Is there no way to achieve the same using [#[instrument]](https://docs.rs/tracing/latest/tracing/attr.instrument.html) macros?
___
*[`nym-vpn-app/src-tauri/src/commands/updater.rs` line 57 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220#-OWuwn1FC-TletukeO9i:-OWuwn1FC-TletukeO9j:b-ogtg) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f7181b962a40d21788fd3da09abd3d7c23516ba0/nym-vpn-app/src-tauri/src/commands/updater.rs#L57)):*
> ```XML
> }
> 
> // TODO: Downgrad log level to info
> ```

Nit: typo in downgrade

pronebird (changes requested):
> Reviewed 5 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 6 of 14 files reviewed, 3 unresolved discussions (waiting on @IRClichtR)

___
*[`nym-vpn-app/src-tauri/src/grpc/client.rs` line 108 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220#-OWuyNYiAxcPIAbSoG8S:-OWuyNYiAxcPIAbSoG8T:b-pn5zfd) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f7181b962a40d21788fd3da09abd3d7c23516ba0/nym-vpn-app/src-tauri/src/grpc/client.rs#L108)):*
> ```XML
>             } else if e.to_string().contains("Unimplemented") {
>                 // Unimplemented error is returned when the vpnd service is not running
>                 debug!("vpnd service is not running: {}", e);
> ```

@doums is it really returning unimplemented when the daemon is not running? Can you check please

pronebird (commented):
> Reviewed 2 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 8 of 14 files reviewed, 3 unresolved discussions (waiting on @doums and @IRClichtR)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 8 of 14 files reviewed, 4 unresolved discussions (waiting on @doums and @IRClichtR)

___
*[`nym-vpn-app/src/types/tauri.ts` line 10 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220#-OWuyloK4H-wWusg3Tny:-OWuyloK4H-wWusg3Tnz:b55atz9) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f7181b962a40d21788fd3da09abd3d7c23516ba0/nym-vpn-app/src/types/tauri.ts#L10)):*
> ```XML
> 
> // Since Javascript in the browser does not have a way to format BackendError objects
> export function formatBackendErrorData(error: BackendError): string | null {
> ```

Is it used anywhere?

pronebird (changes requested):
> Reviewed 2 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 10 of 14 files reviewed, 5 unresolved discussions (waiting on @doums and @IRClichtR)

___
*[`nym-vpn-app/src/layers/EventNotification.tsx` line 26 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220#-OWuz-rmERpCFUqOUJh-:-OWuz-rmERpCFUqOUJh0:bwei8f9) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f7181b962a40d21788fd3da09abd3d7c23516ba0/nym-vpn-app/src/layers/EventNotification.tsx#L26)):*
> ```XML
>   const registerStateListener = useCallback(() => {
>     return listen<TunnelStateEventPayload>(TunnelStateEvent, async (event) => {
>       const serializablePayload = {
> ```

Seems to be unused?

pronebird (commented):
> Reviewed 1 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3220)* status: 11 of 14 files reviewed, 5 unresolved discussions (waiting on @doums and @IRClichtR)

doums:
> ```
>             } else if e.to_string().contains("Unimplemented") {
>                 // Unimplemented error is returned when the vpnd service is not running
>                debug!("vpnd service is not running: {}", e);
> ```
> 
> @doums is it really returning unimplemented when the daemon is not running? Can you check please

I don't think so. You will get a “transport error” in such situation.

`Unimplemented` [variant](https://docs.rs/tonic/latest/tonic/enum.Code.html#variant.Unimplemented) is returned by tonic when a client tries to call non-existing rpc, not implemented in server side.
But here the underlying call is `connect_with_connector` which is related to grpc connection btw client and server, so “Unimplemented error” is not relevant in that context.

nym-vpn-app/src-tauri/src/commands/account.rs
```diff
@@ -65,7 +65,7 @@ pub async fn is_account_stored(grpc: State<'_, GrpcClient>) -> Result<bool, Back
     grpc.is_account_stored()
         .await
         .map_err(|e| {
-            error!("failed to check stored account: {e}");
+            debug!("failed to check stored account: {e}");
```
doums: Why this level downgrade? As I see it we want such critical error to be captured as error.
Here if daemon is failing to check for account stored.

nym-vpn-app/src-tauri/src/commands/tunnel.rs
```diff
@@ -11,7 +11,10 @@ use crate::{
 use tauri::State;
 use tracing::{debug, error, info, instrument, warn};
 
-#[instrument(skip_all)]
+#[instrument(
+    skip_all,
+    err(Debug)
+)]
 #[tauri::command]
```
doums, line 15: We should avoid using `instrument` macros to set the log level (here I assume the purpose of this `err(Debug)`).
Instead, let's stick with simple error!() warn! info! and cie.

nym-vpn-app/src-tauri/src/commands/updater.rs
```diff
@@ -62,12 +63,12 @@ pub async fn install_update(
     on_event: Channel<DownloadUpdateEvent>,
 ) -> Result<(), BackendError> {
     if !*UPDATER_ENABLED {
-        error!("updater is disabled for this build");
+        info!("updater is disabled for this build");
         return Err(BackendError::internal("updater is disabled", None));
```
doums, line 66: We want this to be logged as error, as it is an important one in this context, and we would want to have it reported in sentry

nym-vpn-app/src-tauri/src/commands/updater.rs
```diff
@@ -62,12 +63,12 @@ pub async fn install_update(
...
     }
     let Some(update) = pending_update.0.lock().await.take() else {
         // calling this function without a pending update is an error
-        error!("no update available");
+        info!("no update available");
```
doums: same as above here, please see the comment one line above

nym-vpn-app/src-tauri/src/commands/updater.rs
```diff
@@ -94,6 +95,6 @@ pub async fn install_update(
             },
         )
         .await
-        .inspect_err(|e| error!("download and install failed: {}", e))?;
+        .inspect_err(|e| warn!("Update download and install failed: {}", e))?;
```
doums: Same as above, it is a true error, this should not be reported as a simple warning

nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -188,7 +192,7 @@ impl GrpcClient {
             .listen_to_events(())
             .await
             .inspect_err(|e| {
-                error!("listen_to_tunnel_state_changes failed: {}", e);
+                warn!("listen_to_tunnel_state_changes failed: {}", e); // changed level from error to warning
```
doums, line 192: Error is correct, it is a critical issue if this call is failing, for whatever the reason.

nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -200,12 +204,13 @@ impl GrpcClient {
                         tx.send(update).await.unwrap();
                     }
                     Ok(None) => {
-                        warn!("vpnd DOWN: tunnel state stream closed");
+                        info!("vpnd DOWN: tunnel state stream closed");
                         VPND_DOWN_LOGGED.store(true, Ordering::Relaxed);
```
doums: Nope warn is correct here, we want it to be captured and triggering a sentry event.
Note that there is logic in place to prevent event spamming of this. see `VPND_DOWN_LOGGED`

nym-vpn-app/src-tauri/src/grpc/client.rs
```diff
@@ -401,6 +406,7 @@ impl GrpcClient {
             locale: "en".to_string(),
         });
         let response = vpnd.get_account_links(request).await.map_err(|e| {
+            // TODO change logging level to info
             error!("grpc: {}", e);
```
doums: Nope, `error` is correct

nym-vpn-app/src/contexts/gateways/provider.tsx
```diff
@@ -74,16 +75,21 @@ function GatewaysProvider({ children }: GatewaysStateProviderProps) {
...
               type: 'set-gateways-error',
               payload: {
                 type: nodeType,
-                error: e as BackendError,
+                error: {
+                  message: backendError.message,
+                  key: backendError.key,
+                  data: backendError.data || null,
+                },
               },
```
doums, line 88: What the reason for this manual assignation compared to the previous code?

nym-vpn-app/src/screens/home/Home.tsx
```diff
@@ -53,7 +53,15 @@ function Home() {
...
+          const backendError = e as BackendError;
+          dispatch({ 
+            type: 'set-error',
+            error: {
+              message: backendError.message,
+              key: backendError.key,
+              data: backendError.data,
+            }
+          });
         });
```
doums, line 65: Same question here

nym-vpn-app/src/screens/home/Home.tsx
```diff
@@ -64,10 +72,18 @@ function Home() {
...
```
doums, line 84: Same question here

doums (changes requested):
(all my feedbacks)

nym-vpn-app/src/contexts/gateways/provider.tsx
IRClichtR: The previous code was passing the entire BackendError object directly, which caused serialization issues when JavaScript attempted to stringify it resulting into a [object Object] error instead of meaning-full error in console.log and sentry reporting.

BackendError object contains structures (data) that could not be serialized by default by JS automatically (data) -> I chose manual property extraction to solve that problem

nym-vpn-app/src/screens/home/Home.tsx
IRClichtR: Same here

nym-vpn-app/src/screens/home/Home.tsx
IRClichtR: Same here

nym-vpn-app/src-tauri/src/grpc/client.rs
IRClichtR: The listening to tunnel state changes is more a monitoring than a critical functionality. Failure is likely transient (daemon restart etc.). Kipping it as an error might give the impression that something is profoundly wrong in the VPN while not.

nym-vpn-app/src-tauri/src/commands/updater.rs
IRClichtR: We could meet halfway on this one: since it is not immediately blocking the functionality of the VPN it shouldn't be an error. But since update is important in this context, it could be a warning.

nym-vpn-app/src-tauri/src/commands/updater.rs
IRClichtR: No update is not an error condition. This is the reason why I used info and not error

nym-vpn-app/src-tauri/src/commands/updater.rs
IRClichtR: You're right: since it is a failure to perform action it should remain an error

nym-vpn-app/src-tauri/src/grpc/client.rs
IRClichtR: you're right

nym-vpn-app/src-tauri/src/grpc/client.rs
IRClichtR: you're right

nym-vpn-app/src-tauri/src/commands/updater.rs
pronebird: I think the message is not clear. If there is no update available because we're using the latest version then it's not an error. If there is something else at play, like logical error, then this is different but the message should reflect that too.

nym-vpn-app/src-tauri/src/commands/updater.rs
pronebird: That's what I was afraid of that we'd go over the codebase and lower the severity level to please Sentry instead of adding such errors to ignore list on Sentry backend.

nym-vpn-app/src-tauri/src/commands/account.rs
IRClichtR: You're right, this needs more distinction. One of the recurring failing on this is transient network error, like connection instability (ConnectionReset error for example). I was looking for those. But indeed they are not the only possibility.

IRClichtR:
> `Unimplemented` [variant](https://docs.rs/tonic/latest/tonic/enum.Code.html#variant.Unimplemented) is returned by tonic when a client tries to call non-existing rpc, not implemented in server side. But here the underlying call is `connect_with_connector` which is related to grpc connection btw client and server, so “Unimplemented error” is not relevant in that context.

My bad:  this was an assumption relying on a bad comprehension of the unimplemented error cause from my part

nym-vpn-app/src-tauri/src/commands/updater.rs
IRClichtR: ok how about "no pending update to install" in the info!() and keep the backenderror with "install_update called without a pending update" message?

nym-vpn-app/src-tauri/src/commands/updater.rs
doums: Sorry but no, error is the level we want here because:
This function `fetch_update()` is a tauri command, aka a function that the JS layer can call.
This function should be called only on Windows and if the updater feature has been enabled (at build time, this is per build configuration).
That means if this function is being called when the updater feature is disabled, that is a code error in the js side. Not a runtime nor a business error. Actually the check `if !*UPDATER_ENABLED` is a sanity check, not a business routine guard.
This is why it should be reported as red as possible for the developer (me) to fix it asap.

That is, JS before calling this API, knows which platform it is running on, if the updater feature is enabled and if there is a pending update or not.

Please revert this change.

nym-vpn-app/src-tauri/src/commands/updater.rs
doums: Same as the previous one, this must be logged and captured as an `error`. If this function `install_update()` is called by the JS side when the updater feature is disabled for this build, or when there is no pending update, this is an important code logic error that must be handled as is.

That is, JS before calling this API, knows which platform it is running on, if the updater feature is enabled and if there is a pending update or not.

Please revert this change.

nym-vpn-app/src-tauri/src/commands/updater.rs
doums: Sorry but why you would want to add this error to the ignore list?
What is wrong with capturing errors as errors in Sentry backend?

If tauri call `download_and_install()` is failing, this is an important error in the updater context we would want to be highlighted clearly in the sentry issues list. And reported as error in the logs.

nym-vpn-app/src-tauri/src/grpc/client.rs
doums: > Failure is likely transient (daemon restart etc.)

It is critical, and no, it is not transient, please take a look at the calling side logic https://github.com/nymtech/nym-vpn-client/blob/fe7f254de8f463e2a22fce577147f86253ee7750/nym-vpn-app/src-tauri/src/main.rs#L230
ie: this function is called only if the connection with the daemon is established
If this call fails, we want to know, and it is an error, even a really critical one since we are talking about failure to get update of the tunnel state from the daemon!

nym-vpn-app/src/contexts/gateways/provider.tsx
doums: Ok I see the logic but, here there is no `console.log`, not sure why you're mentioning this then (the js stringify and [object Object]  log output).
Here the  `dispatch` function (from the reducer) is called, passing in the error object.
And `dispatch` does not log anything.

That is, your code does the same thing as the previous one. But it is more verbose with no gain. Please revert this and the same other changes.

nym-vpn-app/src-tauri/src/commands/account.rs
```diff
@@ -80,7 +88,7 @@ pub async fn account_links(
     locale: String,
 ) -> Result<AccountLinks, BackendError> {
     grpc.account_links(&locale).await.map_err(|e| {
-        error!("failed to get account link: {e}");
+        warn!("failed to get account link: {e}");
```
doums, line 91: Still we want this as error.

Actually, better to remove this log as the error is already logged in the grpc client method (`grpc.account_links`).

PR #3239, opened 2025-08-11
Sentry errors refact

IRClichtR:
## Description

Used Sentry  before_send_log directive to exclude errors due to user not being connected.

### tauri & vpnd
- Added before_send directive to exclude user offline or user not authenticated errors from sentry reporting. 
- Keep the errors as excluded errors in breadcrumbs to keep information while improving readability
 
### vpnd
- Moved sentry related functions to a new file for code readability

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3239)


nym-vpn-app/src-tauri/src/sentry.rs
```diff
@@ -23,6 +43,18 @@ pub fn init(os: &OsInfo) -> Option<ClientInitGuard> {
...
             shutdown_timeout: Duration::from_secs(1),
             server_name: Some(Cow::Borrowed("nym")),
+            before_send_log: Some(Arc::new(|log| {
+                if get_excluded_errors()
+                    .iter()
+                    .any(|err| log.body.contains(err.to_lowercase().as_str()))
+                {
+                    // Exclude specific errors from being sent to Sentry
+                    // The excluded log still appears in the breadcrumbs
+                    info!("Excluded log: {}", log.body);
```
pronebird: Can this cause a before_send loop? 

Can we change the severity of the logged message before giving it back to sentry instead?

pronebird (changes requested):

nym-vpn-core/crates/nym-vpnd/src/log_sentry.rs
```diff
@@ -0,0 +1,89 @@
...
+            traces_sample_rate: 1.0,
+            enable_logs: true,
+            shutdown_timeout: Duration::from_secs(2),
+            server_name: Some(Cow::Borrowed("nym")),
+            before_send_log: Some(Arc::new(|log| {
+                if get_excluded_errors()
+                    .iter()
+                    .any(|err| log.body.contains(err.to_lowercase().as_str()))
+                {
+                    tracing::info!("Excluded log: {}", log.body); // Keep excluded logs in breadcrumbs
```
pronebird: Same here. What if we mutate the event level and return it with less sever category instead?

pronebird (changes requested):

nym-vpn-app/src-tauri/src/sentry.rs
IRClichtR: as discussed I changed log level severity instead of suppressing the message

nym-vpn-core/crates/nym-vpnd/src/log_sentry.rs
IRClichtR: Same here

nym-vpn-app/src-tauri/src/sentry.rs
```diff
@@ -23,6 +44,17 @@ pub fn init(os: &OsInfo) -> Option<ClientInitGuard> {
             enable_logs: true,
             shutdown_timeout: Duration::from_secs(1),
             server_name: Some(Cow::Borrowed("nym")),
+            before_send: Some(Arc::new(|mut event| {
+                if let Some(message) = &event.message {
```
pronebird: Nit: perhaps matching against `warn` and `error` levels could save us from parsing lengthy trace/debug messages that we're not interested in. What do you think?

pronebird (approved):
Looks good except one small nitpick.

pronebird (approved):

nym-vpn-core/crates/nym-vpnd/src/log_sentry.rs
doums, line 1: this filename is a bit inaccurate since it is for sentry client initialization
Suggestion: rename it `sentry.rs`

nym-vpn-app/src-tauri/src/sentry.rs
```diff
@@ -1,11 +1,31 @@
...
+            "invalid mnemonic",
+            "no device stored",
+            "no account stored",
+            "ac is offline",
+            "account already exists",
+            "maxdevicesreached",
+            "subscriptionexpired",
+        ])
+    })
+}
```
doums: To streamline this pattern switch to `once_cell`' Lazy (this crate is already in use in tauri)

```
static EXCLUDED_ERRORS: Lazy<HashSet<&'static str>> = Lazy::new(|| {
        HashSet::from([...]);
});
```

doums: ignore that since you are moving to an array

PR #3248, opened 2025-08-12
chore: update nym.com domain

doums:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3248)


pronebird:
Let's wait a bit before merging because tests will fail if there is a difference between what API returns and what we have in the JSON

rokas-ambrazevicius (approved):

pronebird:
Closing since this should have been addressed by now

PR #3257, opened 2025-08-14
Update PR template

rokas-ambrazevicius:
## Description

Suggestion to improve the PR template.

Add ticket number and optional screenshots section,

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3257)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,7 @@
+## Ticket
+
+JIRA-VPN-XXXX
+
 ## Description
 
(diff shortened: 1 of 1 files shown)


neacsu (approved):

doums:
The repo is open public and so the PRs
If for most of them the description starts with a private organization link the average user will get a nice 401 Unauthorized or so when clicking, which is not nice
Imo, the PR link should be set on the corresponding ticket in jira side, and we should avoid linking to jira from this repo since jira is org private.

PR #3411, opened 2025-09-10
fix(vpnd-sentry): don't set extra metadata tag if no extra data available

doums:
Fix sentry "Processing error"s on event when there is no extra metadata

image
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3411)


IRClichtR (approved):
CI building error seems unrelated to code changes.

PR #3430, opened 2025-09-12
Fix type name clash

pronebird:
## Ticket

JIRA-VPN-3962

## Description

- Rename `Gateway` to `GatewayId`
- uniffi: prefix gateway types to avoid type name clash
- makefile: strip unsupported OSes from generated Package.swift

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3430)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
+++ b/nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
@@ -7,7 +7,7 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
 
 // Represents the identity of a gateway
-message Gateway {
+message GatewayId {
   string id = 1;
 }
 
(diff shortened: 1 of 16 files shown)


nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -25,7 +25,7 @@ message Location {
 
 message EntryNode {
   oneof entry_node_enum {
-    Gateway gateway = 1;
+    GatewayId gateway = 1;
```
doums, line 28: Then why not simply using `string gateway_id = 1;`?

pronebird: I kinda agree. The type is superficial. Since string can contain very much any information, I think the original idea was to leverage type system to provide a hint of what's really being held inside the string. I will look into reworking that for simplicity.

pronebird: Your feedback is noted. Let's address that later. I need this PR for Rokas' rpc client integration (draft) and I don't have time to touch that code at the moment.

rokas-ambrazevicius (approved):

PR #3514, opened 2025-09-24
tests: add unit tests for react components

jalolk:
## Ticket

JIRA-VPN-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3514)


nym-vpn-app/package.json
```diff
@@ -14,7 +14,7 @@
     "dev": "vite",
     "dev:app": "tauri dev",
     "dev:browser": "vite --mode dev-browser",
-    "check": "run-s lint tscheck fmt",
+    "check": "run-s lint tscheck fmt test",
```
doums: Please, remove test from the check script
we do not want it here as running unit tests is a heavy task
and that check command is to perform quick checks during dev iterations

doums (changes requested):
my comments

doums:
Please try to set up Typescript to use different and dedicated setup/config for tests. Do not modify the main config used by app code.
Also, `tscheck` or default `tsc` should exclude all test code. That's not relevant. Preserve the existing behavior and do not add code to be parsed  (jest/tests code) by the main existing config because this will slow down dev iterations.

nym-vpn-app/package.json
doums: this `test` script will be added to github workflow CI

doums:
To say the least, I'm quite skeptical on the real benefits to add such tests. This will add its batch of maintenance work and introduces the infamous “unit tests are broken” flow. Which will end as frictions and inevitably will slow down the development process.

For which concrete benefits? I don't remember we encountered a considerable amount of cases of dysfunctional UI component so far.
(I'm not pretending this will never happen)

But right now, the UI layer as proved to be stable and reliable. So I don't get the reasons/motivations here to add this whole layer of UI unit tests (including its batch of tradeoffs)?

nym-vpn-app/eslint.config.js
```diff
@@ -9,9 +9,17 @@ import importPlugin from 'eslint-plugin-import';
 
 export default [
   globalIgnores(['src/types/tauri.ts']),
+  {
+    ignores: [
+      '*.config.js',
+      '*.config.ts',
```
doums, line 15: What is the reason to exclude the config from lint check?

doums: Ah realized it was ignored anyway, all good

tommyv1987 (approved):

PR #3627, opened 2025-10-03
Fix for UDP network calls on iOS 

jmwample:
## Ticket

NET-343 / NET-512

## Description

In rust, on iOS if you have called `UdpSocket::connect()` it throws an error if you call `send_to` even if the remote address is the same as the one you have "connected" yourself to.

resolution options:
  - just use send / recv (in place of send_to and recv_from respectively)
  - don't use connect and provide / check address while using send_to and recv_from

## Checklist:

- [x] Use `send` /  `recv` in place of `send_to` and `recv_from` respectively.
- [x] Tested as a valid solution to the issue in a custom testflight build

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3627)


trojanfoe (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3627#-:-OatWy_h5DluRl2nvaZi:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @pronebird)

PR #3626, opened 2025-10-03
Gateway probe: add subcommand to run in local mode and add more file download details

mmsinclair:
This PR adds two changes:

1. more detailed information in the probe output for file downloads:
- size of file downloaded in bytes
- milliseconds taken to download file

2. Adds a sub-command to run the gateway probe in local mode by providing a mnemonic

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3626)

PR #3621, opened 2025-10-03
Query metadata endpoint from probe

neacsu:
## Ticket

JIRA-VPN-4158

## Description

Include a test in the probe for checking that the metadata endpoint can be queried for the version

## Checklist:

- [ ] Changelog

## Screenshots (optional, if UI related)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3621)

PR #3643, opened 2025-10-07
Gateway probe: Query metadata endpoint, subcommand to run in local mode and more probe info

mmsinclair:
## Ticket

JIRA-VPN-4158

Replaces https://github.com/nymtech/nym-vpn-client/pull/3621 and https://github.com/nymtech/nym-vpn-client/pull/3626

## Description

1. Include a test in the probe for checking that the metadata endpoint can be queried for the version
2. Various fixes in the Go code for better JSON parsing, error handling, logging and instructions to run locally
3. more detailed information in the probe output for file downloads:
- size of file downloaded in bytes
- milliseconds taken to download file
4. Adds a sub-command to run the gateway probe in local mode by providing a mnemonic

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3643)


nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -461,4 +487,76 @@ func downloadFile(url string, timeoutSecs uint64, tnet *netstack.Net) ([]byte, t
...
+			return tnet.Dial(network, addr)
+		},
+	}
+
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   time.Second * time.Duration(timeoutSecs),
+	}
+
+	bandwidthVersionUrl := fmt.Sprintf("%s/v1/bandwidth/version", url)
```
jstuczyn: I think we should rather use `JoinPath` https://pkg.go.dev/net/url#JoinPath otherwise I fear we'll run in our old nemesis: the double slash issue

nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -461,4 +487,76 @@ func downloadFile(url string, timeoutSecs uint64, tnet *netstack.Net) ([]byte, t
...
+
+	start := time.Now() // Start timing
+
+	log.Printf("Querying metadata encoding: url = %s", bandwidthVersionUrl)
+	resp, err := client.Get(bandwidthVersionUrl)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer resp.Body.Close()
```
jstuczyn, line 515: you want the call to defer to happen before checking for error, otherwise, you might return an error response without closing the reader

nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -461,4 +487,76 @@ func downloadFile(url string, timeoutSecs uint64, tnet *netstack.Net) ([]byte, t
...
+
+	if resp.StatusCode != http.StatusOK {
+		return 0, 0, fmt.Errorf("failed to query metadata endpoint: %s", resp.Status)
+	}
+
+	var contentType = resp.Header.Get("Content-Type")
+
+	log.Printf("Metadata Content-Type: %s", contentType)
```
jstuczyn, line 523: is this log relevant in the context of the probe?

jstuczyn (changes requested):

nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
mmsinclair: I think we should keep it, in case anyone assumes it is bincode in the future, or the default becomes bincode and the probe fails.

nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
mmsinclair: yes good catch

nym-vpn-core/crates/nym-gateway-probe/netstack_ping/lib.go
mmsinclair: changed

jstuczyn (approved):

PR #3725, opened 2025-10-21
Initialize COM on main thread to prevent crash on exit

pronebird:
## Ticket

JIRA-VPN-XXXX

## Description

It seems that there is a bug in COM + thread_local destructor that causes the daemon to crash on exit. As a workaround, initialize COM on main thread which magically resolves the issue.

See: https://github.com/ohadravid/wmi-rs/issues/136

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3725)


trojanfoe (approved):
Tested, working on my rig.

PR #3783, opened 2025-10-28
Add back channel for bridge forwarder

pronebird:
## Description

This PR focuses around adding a back channel into UDP forwarder for bridges along with some improvements with regards of separation of concern for `ConnectedTunnel` which is now 

- Add back channel to bridge forwarder to learn when it exits with error
- Move supporting services from `TunnelHandle` into `WgTunnelRuntime` and hold it in `TunnelMonitor`. No need to drag them into `ConnectedTunnel` and then into `TunnelHandle`
- Remove `mixnet_client_token()` from `TunnelHandle`, move it to `WgTunnelRuntime`
- Add  "quinn::connection" to warn crates since it logs every new connection in the console log which is pretty spammy

Further thoughts: we can get rid of `TunnelHandle` because it's essentially cancellation token + join handle.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3783)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
@@ -917,7 +917,7 @@ impl TunnelMonitor {
             registration_result.mixnet_client,
             assigned_addresses,
             tun_device,
-            self.shutdown_token.clone(),
+            self.shutdown_token.child_token(),
             registration_result.event_rx,
         )
         .await
@@ -971,7 +971,7 @@ impl TunnelMonitor {
                 .nym_config
                 .network_env
                 .gw_update_version(),
-            self.shutdown_token.clone(),
+            self.shutdown_token.child_token(),
         )
         .await
         .map_err(|e| Box::new(tunnel::Error::from(e)))?;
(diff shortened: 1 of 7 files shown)


jmwample (approved):
Did several manual tests to make sure this behaves predictably.

- regular use, shutdown using a disconnect command

- triggered error in UdpForwarder (upstream and downstream copy tested separately) results in clean closure and re\-connect.

- shutdown vpnd using ctrl\-c closes normally without hanging on timeout or leaving firewall rules.

- handshake error (eg crypto error) results in clean shutdown and reconnection attempt.

I don't know that this is a great long-term solution, but I think a proper runtime / error handling cancellation is a bigger project.

I am going to look into the `drive: quinn::connnection close` log lines - I still am not sure what is causing those in such volume. Especially since I had already removed the `log` feature from the quinn dependency. I want to make sure that isn't flooding logs in `nym-bridge` as well. That can be separate from this effort though.

:lgtm:

> @jmwample reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3783#-:-Ocgy3AA3YqXcy2MuoSe:byzaa47)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @simonwicky and @trojanfoe)

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
```diff
@@ -753,6 +781,14 @@ impl TunnelMonitor {
                         }
                     }
                 }
+                close_event = bridge_close_rx.recv() => {
+                    if close_event.is_some() {
+                        tracing::info!("Bridge close signal received. Exiting");
+                    } else {
+                        tracing::info!("Bridge channel is closed. Exiting");
```
simonwicky, line 808: That doesn't fly. If the bridge isn't enabled, `bridge_close_tx` is dropped, so this branch is hit, preventing all connection attempts

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
pronebird: Fixed

tommyv1987:
Let's target the FIG release branch for this please @pronebird

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
simonwicky: Looks like it is fixed now

simonwicky (approved):

PR #3789, opened 2025-10-29
VPN-3966: Add streaming icon

agentpietrucha:
## Ticket

[JIRA-VPN-3966](https://nymtech.atlassian.net/browse/VPN-3966)

## Description

- Add streaming icon to gateways with residential IPs
- Add Alert explaining that new icon
- Update exit location detail dialog text

## Checklist:

- [ ] Changelog

image
image
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3789)


nym-vpn-app/src/contexts/gateways/provider.tsx
```diff
@@ -129,7 +129,8 @@ function GatewaysProvider({ children }: GatewaysStateProviderProps) {
     countryCode: string,
     gateways: GatewaysByCountry[],
   ) => {
-    const byCountry = gateways.find((c) => c.country.code === countryCode);
+    const cc = countryCode.toUpperCase();
```
agentpietrucha: I found out that the `gateway.country.code` is uppercase and the `countryCode` is lower. So it's more efficient to make the `countryCode` upper case

doums: Hmm, looking at the code they should be both uppercase. But you're right let's play safe and to keep it consistent with the rest of the code base, can you lower case them both?

```
c.country.code.toLowerCase() === countryCode.toLowerCase()
```

nym-vpn-app/src/main.tsx
```diff
@@ -83,6 +84,8 @@ dayjs.extend(duration);
     vpnMode: (await kvGet<VpnMode>('vpn-mode')) || defaultVpnMode,
     uiTheme: await getTheme(),
     welcomeChecked: (await kvGet<boolean>('welcome-screen-seen')) || false,
+    streamingOptimizedLabelSeen:
+      (await kvGet<boolean>(STREAMING_OPTIMIZED_LABEL_FEATURE_KEY)) || false,
```
doums: you don't need to use this constant for the key name, the key values are typed anyway (from tauri rust side)

also to keep consistency
```
await kvGet<boolean>('<key-name>'))
```

nym-vpn-app/src/types/tauri.ts
```diff
@@ -156,8 +155,6 @@ export type GatewaysByCountry = {
   quic: boolean;
 };
 
-export type GpuType = 'nvidia' | 'amd' | 'intel' | { unknown: string | null };
-
```
doums, line 160: same as above

nym-vpn-app/src/types/tauri.ts
```diff
@@ -213,8 +210,6 @@ export type OsInfo = {
   version: string;
   kernel: string;
   arch: string;
-  displayServer: DisplayServer;
-  gpu: GpuType;
   hash: string;
```
doums, line 219: Same as above

nym-vpn-app/src/types/tauri.ts
```diff
@@ -77,9 +77,8 @@ export type DbKey =
   | 'cache-mx-exit-gateways'
   | 'cache-wg-gateways'
   | 'cache-account-id'
-  | 'cache-device-id';
-
-export type DisplayServer = 'x11' | 'wayland' | 'unknown';
```
doums, line 82: Please revert that removal, this is due to platform gated defined type in Rust side, and you are developing using a non Linux machine.
The source type will need a bit of rework.
For now, it's OK to keep it as is.

nym-vpn-app/src/screens/home/new-feature-alert/streaming-optimized-label/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const FEATURE_KEY = 'streaming-optimized-label-seen' as const;
+export const ACTION_TYPE = 'set-streaming-optimized-label-seen' as const;
```
doums: For code consistency and keep things simple, let's avoid creating unnecessary constants(/abstractions), TS type system is already there to keep us safe. We don't need such constants.

When needed just call the dispatch with the correct action key and that's it.
Same for kvSet.

nym-vpn-app/src/screens/home/new-feature-alert/utils/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  typeof STREAMING_OPTIMIZED_LABEL_ACTION_TYPE
+>;
+
+export const setFeatureSeen = (
+  dispatch: StateDispatch,
+  action: AllowedActionTypes,
+  featureKey: DbKey,
+) => {
+  dispatch({ type: action, seen: true });
+  kvSet(featureKey, true);
```
doums: see my comment above

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
```diff
@@ -33,21 +36,75 @@ function LocationDetailsDialog({ isOpen, onClose }: Props) {
...
+              residentialLink: (
+                <Link
+                  text={t('location-details.streaming.residentialLink')}
+                  url={ResidentialIpServersUrl}
+                  data-testid="location-details-residential-link"
+                  className="underline !text-black dark:!text-white"
+                />
+              ),
+            }}
+          />
```
doums: Let's use this way for translation with embedded links in text https://github.com/nymtech/nym-vpn-client/blob/25a0700f9915ae7783302105c67d28c511d1274f/nym-vpn-app/src/screens/node/details/NodeDetails.tsx#L282

It is both cleaner and simpler to handle by the translation team

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
```diff
@@ -33,21 +36,75 @@ function LocationDetailsDialog({ isOpen, onClose }: Props) {
...
+            i18nKey="location-details.location.description"
+            ns="nodeLocation"
+            components={{
+              locationAccuracyLink: (
+                <Link
+                  text={t('location-details.location.locationAccuracyLink')}
+                  url={LocationDetailsArticle}
+                  data-testid="location-details-location-accuracy-link"
+                  className="underline !text-black dark:!text-white"
+                />
```
doums: same as above

nym-vpn-app/src/screens/home/new-feature-alert/utils/index.ts
doums: Tip: you can make this function specialized for the stream-opti-label-seen and not passing around those unnecessary action types and db keys then simply call it from where you need, eg.

```
setStreamOptimizedLabelSeen(true);
```

nym-vpn-app/src/types/app-state.ts
```diff
@@ -32,6 +32,7 @@ export type InitState = {
   welcomeChecked: boolean;
   vpnMode: VpnMode;
   vpnd: VpndStatus;
+  streamingOptimizedLabelSeen: boolean;
```
doums: This early init state type, is good for important state bits (only) that need to be prepared as soon as possible and before the React app is spawned.

But that is not the case for `streamingOptimizedLabelSeen`.

To initialize it (using kvGet), you should do it in `initFirstBatch()` https://github.com/nymtech/nym-vpn-client/blob/d4159a0234f76ff18bb5d88bce3e5a2ed85829f7/nym-vpn-app/src/state/init.ts#L40
The advantage of `initFirstBatch()` is that it parallelizes all async calls, including db gets.

nym-vpn-app/src/main.tsx
doums: Please see my comment on the `InitState` and why you should not init this boolean here.

nym-vpn-app/src/types/tauri.ts
doums: (*the above message, idk why but gh seems mixing comments ordering) Please revert that removal, this is due to platform gated defined type in Rust side, and you are developing using a non Linux machine.
The source type will need a bit of rework.
For now, it's OK to keep it as is.

doums (changes requested):
see my comments

nym-vpn-app/src/constants.ts
```diff
@@ -49,3 +49,7 @@ export const SupportServerLocationUrl =
   'https://support.nym.com/hc/en-us/articles/26448676449297-How-is-server-location-determined-by-NymVPN';
 export const NetworkExplorerNodeUrl = 'https://nym.com/explorer/nym-node';
 export const countriesWithRegions = ['US', 'CA', 'AU', 'MX', 'BR', 'IN', 'CN'];
+export const ResidentialIpServersUrl =
+  'https://support.nym.com/hc/en-us/articles/35279486714641-Why-can-t-I-access-streaming-services-while-using-NymVPN';
+export const LocationAccuracyLink =
+  'https://support.nymvpn.com/hc/en-us/articles/26448676449297-How-is-server-location-determined-by-NymVPN';
```
doums: is it `support.nym.com` or `support.nymvpn.com`?

nym-vpn-app/src/contexts/gateways/provider.tsx
```diff
@@ -130,7 +130,8 @@ function GatewaysProvider({ children }: GatewaysStateProviderProps) {
     countryCode: string,
     gateways: GatewaysByCountry[],
   ) => {
-    const byCountry = gateways.find((c) => c.country.code === countryCode);
+    const cc = countryCode.toLowerCase();
+    const byCountry = gateways.find((c) => c.country.code.toLowerCase() === cc);
```
doums: `cc` is redundant `c.country.code.toLowerCase() === countryCode.toLowerCase()`

nym-vpn-app/src/constants.ts
agentpietrucha: `support.nym.com` does have the related content. So I guess it's correct. I also took those links directly from [Figma](https://www.figma.com/design/SMHVIU2QPxz9LVDqWXo0Ed/%F0%9F%9A%A9NymVPN?node-id=26901-110064&m=dev)

nym-vpn-app/src/contexts/gateways/provider.tsx
agentpietrucha: this way the `countryCode.toLowerCase()` is computed only once. This is the reason I introduced additional variable. However, if you wish I am totally happy to remove it

nym-vpn-app/src/screens/home/new-feature-alert/streaming-optimized-label/StreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,55 @@
...
+      setStreamOptimizedLabelSeen(dispatch);
+    }
+  };
+
+  if (streamingOptimizedLabelSeen) {
+    return null;
+  }
+
+  return (
+    <Toast
```
doums: You should not need to use `<Toast />` component directly:  check `const { push } = useInAppNotify();` for examples on how to send such notification.
(Maybe you will need some refactoring in order to pass in a component instead of raw text)

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
```diff
@@ -33,21 +36,75 @@ function LocationDetailsDialog({ isOpen, onClose }: Props) {
...
+          data-testid="location-details-streaming-description"
+        >
+          <Trans
+            i18nKey="location-details.streaming.description"
+            ns="nodeLocation"
+          >
+            <Link
+              url={ResidentialIpServersUrl}
+              data-testid="location-details-residential-link"
+              className="underline text-black! dark:text-white!"
```
doums: Are those `!important` modifier really necessary? In general, it is best to avoid them and find better solution for style override issue.
Overusing them is a known antipattern.

Note that `clsx` is extensively used in the code base, and does a pretty good job for handling “dynamic” classes and merge.

doums: Also double-check the `<Link />` component, it takes 3 levels of className props allowing you to override style as needed.

(In case this not covers your need, the right solution is then, to refactor the Link component and extend it's className/style API)

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
```diff
@@ -33,21 +36,75 @@ function LocationDetailsDialog({ isOpen, onClose }: Props) {
...
+        >
+          <Trans
+            i18nKey="location-details.location.description"
+            ns="nodeLocation"
+          >
+            Displayed locations are
+            <Link
+              url={LocationDetailsArticle}
+              data-testid="location-details-location-accuracy-link"
+              className="underline text-black! dark:text-white!"
```
doums: same here

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
agentpietrucha: my overlook there with classnames, sorry!

nym-vpn-app/src/screens/node/LocationDetailsDialog.tsx
```diff
@@ -33,21 +36,75 @@ function LocationDetailsDialog({ isOpen, onClose }: Props) {
...
+            icon="location_on"
+            data-testid="location-details-location-icon"
+          />
+          <h4 className="text-lg" data-testid="location-details-location-title">
+            {t('location-details.location.title')}
+          </h4>
+        </div>
+        <p
+          className="text-iron dark:text-bombay md:text-nowrap"
+          data-testid="location-details-location-description"
```
doums: Please in general: do not add `data-testid` (everywhere), remove them

1. it is not a dev role, this is QA stuff
2. do not do that, stop the bad practice (there has been internal extensive discussion on this topic) 
3. soon we will clean the existing ones

nym-vpn-app/src/constants.ts
doums: then fix it 
image

nym-vpn-app/src/contexts/gateways/provider.tsx
doums: But this var is only used one time in that check, so what's the point?

nym-vpn-app/src/constants.ts
agentpietrucha: sorry, I wasn't clear enough. The links I used are correct (both display related content). I got them from figma. Screenshot for proof:
image

doums: OK, but it is obv using 2 different domains is not right
Please ask internally for clarification

nym-vpn-app/src/contexts/gateways/provider.tsx
agentpietrucha: _correct me if I'm wrong_
the `.filter` goes over the `gateways` array. So the check `c.country.code.toLowerCase() === countryCode.toLowerCase()` is performed for every `gateway` element. Meaning the `countryCode.toLowerCase()` is calculated `gateways.length + 1` times. So having it calculated once, before the iteration, should be beneficial? (besides that such small improvement will not make a huge difference)

agentpietrucha: I removed the additional variable

nym-vpn-app/src/screens/home/new-feature-alert/streaming-optimized-label/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  useEffect(() => {
+    if (
+      !streamingOptimizedLabelSeen &&
+      current?.id !== 'streaming-optimized-label'
+    ) {
+      push({
+        close: true,
+        clickAway: false,
+        duration: Infinity,
```
doums: Let's not use an infinity duration! As you know the current in-app notification implementation is limited and only display one notification at a time. So let's wait 10sec let's say. So if user does not close it, it will auto close and allow other notifications in the stack to flow in.

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,63 @@
+import { useEffect } from 'react';
```
doums, line 1: Let's avoid too deep directory levels in general, where it is not necessary, let's flatten this one
You can move useStream...tsx right under the `home` directory

nym-vpn-app/src/screens/home/new-feature-alert/streaming-optimized-label/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { StateDispatch } from '../../../../types';
+import { kvSet } from '../../../../kvStore';
+
+export const setStreamOptimizedLabelSeen = (dispatch: StateDispatch) => {
+  dispatch({ type: 'set-streaming-optimized-label-seen', seen: true });
+  kvSet('streaming-optimized-label-seen', true);
+};
```
doums: let's move it into `home/util.ts` (flatten directory arch)

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
agentpietrucha: and what about the `utils.ts`? Move the function inside the `useStreamingOptimizedLabel.tsx`?

agentpietrucha: nevermind, saw next comment

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { useNavigate } from 'react-router';
+import { useTranslation } from 'react-i18next';
+import { ButtonText } from '../../ui/index';
+import { routes } from '../../router';
+import { useInAppNotify, useMainDispatch, useMainState } from '../../contexts';
+import { StateDispatch } from '../../types/index';
```
doums: `from '../../types';`

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,59 @@
+import { useEffect } from 'react';
+import { useNavigate } from 'react-router';
+import { useTranslation } from 'react-i18next';
+import { ButtonText } from '../../ui/index';
```
doums: `from '../../ui';`

nym-vpn-app/src/ui/Toast.tsx
```diff
@@ -113,8 +118,13 @@ function Toast({
...
             >
-              {title && <div data-testid={`${testId}-title`}>{title}</div>}
-              <div data-testid={`${testId}-message`}>{message}</div>
+              {!content && title && (
+                <div data-testid={`${testId}-title`}>{title}</div>
+              )}
+              {!content && message && (
+                <div data-testid={`${testId}-message`}>{message}</div>
+              )}
+              {content && content}
```
doums: suggestion (more readable)

```
              {content ? (
                content
              ) : (
                <>
                  {title && <div>{title}</div>}
                  <div>{message || ''}</div>
                </>
              )}
```

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  return (
+    <div className="flex flex-row justify-between gap-5 items-center">
+      <span className="dark:text-white text-baltic-sea">
+        {t('streaming-optimized-label')}
+      </span>
+      <ButtonText
+        color="transparent"
+        className="text-malachite"
```
doums: Instead of passing arbitrary className to override text color, you should extend the `ButtonText` prop `color` API, add a new one eg. "malachite", see `nym-vpn-app/src/ui/ButtonText.tsx`

nym-vpn-app/src/screens/home/useStreamingOptimizedLabel.tsx
```diff
@@ -0,0 +1,56 @@
...
+    navigate(routes.exitNodeLocation);
+  };
+
+  return (
+    <div className="flex flex-row justify-between gap-5 items-center">
+      <span className="dark:text-white text-baltic-sea">
+        {t('streaming-optimized-label')}
+      </span>
+      <ButtonText color="malachite" onClick={handleClick}>
+        <span className="">{t('streaming-optimized-label-button')}</span>
```
doums: Useless span wrapper? If yes just remove it

nym-vpn-app/src/ui/Toast.tsx
```diff
@@ -102,6 +104,7 @@ function Toast({
                   'border-2 text-aphrodisiac! dark:text-aphrodisiac! border-aphrodisiac',
                 type === 'warn' && 'border-2 border-king-nacho',
                 type === 'info' && 'border-2 border-iron dark:border-bombay',
+                type === 'ghost' && 'border-2 border-transparent',
```
doums, line 107: no border? `type === 'ghost' && 'border-0'`

agentpietrucha: in figma there's no border:
image

doums: yup then you can use `'border-0'` instead of `'border-2 border-transparent'`

agentpietrucha: it looks better with the transparent border. More like on the figma. And is more consistent with other types. So I'd leave the transparent border. See below for difference:
**border-0**
image

**border-2 border-transparent**
image

doums: ok

doums (approved):

PR #3801, opened 2025-10-29
fix(tauri): various fixes and cleanup

doums:
- fix a race condition that could result to reset (wrongly) the selected nodes
- filter out country and region with empty gateway list from the server list
- minor changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3801)

### DIFF
--- a/nym-vpn-app/CHANGELOG.md
+++ b/nym-vpn-app/CHANGELOG.md
@@ -20,6 +20,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search by city, region and server ID in server list
 - Optimize server list filtering, searching and rendering performance
 
+### Fixed
+
+- Fix a race condition at app startup that could lead to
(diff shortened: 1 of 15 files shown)


tommyv1987 (approved):
Looks good - let's ensure the QUIC/domain-fronting toggle is sorted before the next RC candiate, maybe restore safe optional chaining for backendFlags? And another one remove the unused Button import?

agentpietrucha (approved):

doums:
> maybe restore safe optional chaining for backendFlags

It is superfluous, as feature flags are set to a default value anyway, https://github.com/nymtech/nym-vpn-client/blob/720bb3bd7655e6945b11600790a32153e950cc94/nym-vpn-app/src/contexts/main/reducer.ts#L107

doums:
> remove the unused Button import?

Which one?
In any case that would be caught by ts or eslint -> failing the CI

PR #3849, opened 2025-10-31
DNS min TTL in default resolver

jmwample:
## Description

This PR adds a default minimum TTL for positive lookups in the default DNS resolver. 

If a DNS record lookup succeeds it will now be cached for a minimum of 30 minutes. Records with longer TTLs will keep the value configured in the record.

Considerations:
- The addresisng for Nym APIs are relatively static, but have short timeouts currently resulting in constant use of DNS.
  - `nymvpn.com` and `validator.nymtech.net` use 30s TTL for `A` records and 10800s (3d) for `AAAA`
- Should addresses change in the future, clients should be able to adapt in a reasonable amount of time.
- I believe the cache is per instance of the resolver, so if the daemon is shutdown and relaunched the cache should start from a clean slate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3849)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/resolver.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/resolver.rs
@@ -497,8 +497,12 @@ impl LocalResolver {
             NameServerConfigGroup::from_ips_clear(&dns_servers, DNS_LISTEN_PORT, true);
 
         let forward_config = ResolverConfig::from_parts(None, vec![], forward_server_config);
+        let mut opts = ResolverOpts::default();
+        // Set a 30 min minimum TTL for caching DNS responses.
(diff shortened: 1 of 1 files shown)


pronebird (commented):
The thing is we use `HickoryResolver::default()` from nym monorepo for API dns resolution. I don't think our resolver is involved at that point as it's running on localhost and we talk directly to quad9, cloudfront etc.

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3849#-:-Ocv_tC567ErbCMInd_2:bitsulh)* status: 0 of 1 files reviewed, all discussions resolved

jmwample:
I am closing this as the DNS resolver that is actually responsible is elsewhere. 
https://github.com/nymtech/nym/pull/6175

PR #3898, opened 2025-11-06
refactor(tauri): simplify os info type

doums:
Do not conditionally gate some properties of `OsInfo` and expose them as common type.
As `OsInfo` is TS exported, this remove the previous and annoying differences of the produced TS when generating from Linux, macOS, or Windows.

JIRA-VPN-4418

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/3898)


agentpietrucha (approved):
Code looks nice. I also tested in locally on macos and finally `displayServer` & `gpu` are not disappearing :)

PR #4007, opened 2025-11-20
VPN-4521: Handle `get_remote_time` error.

trojanfoe:
## Ticket

JIRA-VPN-4521

## Description

When syncing the account, handle errors from `get_remote_time()` in the same way we handle errors from `get_account_summary_with_device()`.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4007)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
@@ -1,13 +1,12 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use crate::state_machine::DecentralisedState;
 use crate::{
     SharedAccountState,
     commands::{AccountCommand, UpgradeModeCommand, common_handler, handler},
     state_machine::{
-        AccountControllerStateHandler, ErrorState, LoggedOutState, NextAccountControllerState,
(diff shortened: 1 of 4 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
```diff
@@ -91,13 +90,34 @@ impl SyncingState {
...
-            return Err(SyncError::DeviceTimeDesynced);
+        match vpn_api_client.get_remote_time().await {
+            Ok(remote_time) => {
+                if !remote_time.is_acceptable_synced() {
+                    return Err(SyncError::DeviceTimeDesynced);
+                }
+            }
+            Err(e) => {
+                return handle_vpn_api_error(e);
+            }
```
agentpietrucha, line 120: should the usage of `get_remote_time` in `crates/nym-vpn-api-client/src/client.rs` be handling the errors similarly?

trojanfoe: Yeah I reckon so.   There are two calls to `vpn_api_client` which both return the same error type, so they should be handled the same.

Not that this change fixes the root-cause, as I was unable to reproduce the issue, but it should provide better diagnostics.

agentpietrucha (approved):

trojanfoe:
@agentpietrucha Thanks for the approval.

PR #4013, opened 2025-11-21
VPN-4087: Window resize fix

agentpietrucha:
## Ticket

[JIRA-VPN-4087](https://nymtech.atlassian.net/browse/VPN-4087)

## Description

- Remove `max-w-lg` class from containers. The app max size is limited in `src-tauri/src/window.rs:create_main_window`, so the double size guard is not necessary
- Increase `min_inner_size` to match the initial size. This will prevent the app window width to "squish" it's content

## Checklist:

- [ ] Changelog

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4013)

### DIFF
--- a/nym-vpn-app/src-tauri/src/window.rs
+++ b/nym-vpn-app/src-tauri/src/window.rs
@@ -82,7 +82,7 @@ impl AppWindow {
         .center()
         .focused(true)
         .inner_size(328.0, 710.0)
-        .min_inner_size(328.0, 200.0)
+        .min_inner_size(160.0, 200.0)
         .max_inner_size(800.0, 1400.0)
         .initialization_script(format!("window._APP = {win_env};"))
         .build()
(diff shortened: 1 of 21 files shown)


nym-vpn-app/src-tauri/src/window.rs
```diff
@@ -82,7 +82,7 @@ impl AppWindow {
         .center()
         .focused(true)
         .inner_size(328.0, 710.0)
-        .min_inner_size(160.0, 200.0)
+        .min_inner_size(328.0, 200.0)
```
doums: This is a bad idea:
The user's monitor DPI is not predictable and user can set whatever in-app UI scaling.
In some situations (Linux and high or very low DPI monitors) users would want to resize down the window width as needed.
And the UI is responsive enough to be squished.

doums (approved):

PR #4015, opened 2025-11-21
VPN-4508: Allow custom DNS servers to be used.

trojanfoe:
## Ticket

JIRA-VPN-4508

## Description

Allow user to define the DNS servers used by the VPN. 

It's currently only available to set/get/clear via the CLI:

```
nym-vpnc custom-dns get
nym-vpnc custom-dns set 8.8.8.8 2a00:1450:4009:815::200e
nym-vpnc custom-dns clear
```

## Checklist:

- [x ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4015)


tommyv1987:
All primary features tested and confirmed working:

**DNS Management**
- `nym-vpnc dns get/set/clear` commands work correctly
- Settings persist across daemon restarts
- IPv4 and IPv6 addresses supported

**Tunnel Modes**
- 5-hop (mixnet) mode with custom DNS: Works
- 2-hop (wireguard) mode with custom DNS: Works
- Switching between modes with custom DNS set: Works correctly
- Brief internet drop during mode switching is expected behavior

**DNS Application**
- Custom DNS correctly applied once connected
- DNS forwarding works as expected
- Default DNS restored when custom DNS is cleared

Caveats and known limitations

Custom DNS servers are used for pre-connection API resolution. If the DNS server is not reachable before the VPN tunnel is established, connection will fail with "No route to host" errors.

**Tested DNS Servers:**
- Quad9 (9.9.9.9): Works reliably
- OpenDNS (208.67.222.222): (hit and miss) 

_More servers were tested, however, the above is highlighting two dns used in the test case scenarios_

During connection setup, the firewall policy is restrictive and only allows queries to the custom DNS server. If the server is unreachable from the user's network, API hostname resolution fails.

A user must select a custom DNS server that is reachable from their network during the connection phase. This is a network-dependent limitation.

tommyv1987 (approved):

PR #4031, opened 2025-11-25
feat(tauri): remove domain fronting ff check

doums:
Remove feature flag check for domain fronting (aka stealth API), always show the switch

JIRA-VPN-4548

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4031)

### DIFF
--- a/nym-vpn-app/src/screens/settings/Settings.tsx
+++ b/nym-vpn-app/src/screens/settings/Settings.tsx
@@ -13,7 +13,7 @@ import SettingsGroup from './SettingsGroup';
 import Logout from './Logout';
 
 function Settings() {
-  const { desktopNotifications, ipv6Support, backendFlags } = useMainState();
+  const { desktopNotifications, ipv6Support } = useMainState();
 
   const navigate = useNavigate();
   const dispatch = useMainDispatch() as StateDispatch;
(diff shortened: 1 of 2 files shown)


nym-vpn-app/src/screens/settings/Settings.tsx
```diff
@@ -13,15 +13,14 @@ import SettingsGroup from './SettingsGroup';
...
-  const { desktopNotifications, ipv6Support, backendFlags } = useMainState();
+  const { desktopNotifications, ipv6Support } = useMainState();
 
   const navigate = useNavigate();
   const dispatch = useMainDispatch() as StateDispatch;
   const { t } = useTranslation('settings');
   const { exit } = useExit();
   const { enabled: autostartEnabled, toggle: toggleAutostart } = useAutostart();
   const toggleDNotifications = useDesktopNotifications();
-  const showAntiCensorship = backendFlags.quic || backendFlags.domainFronting;
```
agentpietrucha, line 24: if those flags won't be used anymore, maybe they can be cleaned up from `feature_flags.rs`?

agentpietrucha (changes requested):

nym-vpn-app/src/screens/settings/Settings.tsx
doums: No, `feature_flags.rs` exposes the whole FF flags set. We still use it for QUIC. That's what we want. In the future we could want to access new FF as they are added etc.

agentpietrucha (approved):

PR #4039, opened 2025-11-26
Nicolas/introduced new parameters

nick1231321:
## Ticket

JIRA-VPN-4558

## Description

This update introduces four new configurable parameters to the `MixnetClient`.  
These parameters allow more manual control over timing and rate behaviors that affect packet delivery and cover-traffic generation.

> **Note:** This PR requires the following NYM SDK branch for compatibility:  
> `nicolas/sdk-param-support-debug`

## Checklist

- [ ] Verified no breaking changes to existing `MixnetClient` behavior
- [ ] Verified debugging output to ensure parameters are applied correctly.

  - Details are documented in **Configurable Mixnet Params – Design** (Confluence)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039)


nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -25,6 +25,12 @@ pub struct VpnServiceConfig {
     pub min_gateway_mixnet_performance: Option<u8>,
     pub min_gateway_vpn_performance: Option<u8>,
     pub residential_exit: bool,
+    pub poisson_parameter: Option<f32>,
+    /// Average delay for packet sending (in milliseconds)
+    pub average_packet_delay: Option<f32>,
```
trojanfoe: I don't know why `f32` would be used to represent milliseconds.  `u32` seems more natural.

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -83,6 +101,10 @@ impl Default for VpnServiceConfig {
             min_gateway_mixnet_performance: None,
             min_gateway_vpn_performance: None,
             residential_exit: false,
+            poisson_parameter: None,
+            // new defaults
```
trojanfoe: Please remove this comment.

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -25,6 +25,12 @@ pub struct VpnServiceConfig {
     pub min_gateway_mixnet_performance: Option<u8>,
     pub min_gateway_vpn_performance: Option<u8>,
     pub residential_exit: bool,
+    pub poisson_parameter: Option<f32>,
+    /// Average delay for packet sending (in milliseconds)
```
trojanfoe: Please remove these comments.

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -25,6 +25,12 @@ pub struct VpnServiceConfig {
     pub min_gateway_mixnet_performance: Option<u8>,
     pub min_gateway_vpn_performance: Option<u8>,
     pub residential_exit: bool,
+    pub poisson_parameter: Option<f32>,
```
trojanfoe: Should this be `f32`?

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -239,6 +239,12 @@ message ConnectRequest {
   UserAgent user_agent = 12;
 }
 
+message SetPoissonParameterRequest {
+  float poisson_parameter = 1;
```
trojanfoe: Should this be `float`?

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -254,6 +260,9 @@ message VpnServiceConfig {
   optional uint32 min_mixnode_performance = 8;
   optional uint32 min_gateway_mixnet_performance = 9;
   optional uint32 min_gateway_vpn_performance = 10;
+  optional float poisson_parameter = 16;
```
trojanfoe: Consider `uint32`, as per my comments above about millisecond representation.

nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs
```diff
@@ -75,7 +75,55 @@ impl RpcClient {
...
+    };
+
+    self.0
+        .set_poisson_parameter(request)
+        .await
+        .map_err(Error::Rpc)?;
+    Ok(())
+}
+    /// Sets the average per-mixnode packet delay (in milliseconds)
+    pub async fn set_average_packet_delay(&mut self, delay_ms: f32) -> Result<()> {
```
trojanfoe: There are many formatting issues; please run `cargo +nigthly fmt --all`.

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
```diff
@@ -827,6 +848,34 @@ impl NymVpnService {
         self.update_tunnel_settings_with_throttle();
     }
 
+async fn handle_set_poisson_parameter(&mut self, value: f32) {
+    tracing::debug!(
```
trojanfoe: Please remove these debug logs.

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
```diff
@@ -956,7 +1036,7 @@ impl NymVpnService {
             exit,
             options,
         } = connect_args;
-
+        let temporar_config  = self.config_manager.config();
```
trojanfoe: This all looks bad.  Why are you messing with the config here?

trojanfoe (changes requested):

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
nick1231321: How can I read the  current options for poisson, average delay, message sending average delay and propagate them to the newly constructed config?

trojanfoe: You add setters to `VpnServiceConfigManager`.

But this function is setting the network; so why does it have to get involved with the config at all?

nick1231321: In the current develop branch, the VpnServiceConfig structure is instantiated here
image
Given this, I have a question regarding the new fields I introduced:

If I remove the temporary config lookup, should these new fields be initialized as None when constructing VpnServiceConfig?
Are there any scenarios where the configuration might already exist in the config_manager and should be propagated into this newly created config after a connect command is triggered?

My concern is that if I simply pass None, the existing configuration values may never be saved or carried over. Could you advise on the correct approach here?

trojanfoe: The default values for the vpn service configuration is up to you really, however the vpn service configuration does not need to be changed when the network is changed as that occurs in the global config.

So it looks to me like there is no need to change this function at all.

nick1231321: My question is the following:
The handle_connect function is triggered when the connect command is issued to the Nym VPN client. There is a scenario where the VPN service daemon is already running and receives these options without being connected. Given that, if I set default values for these options before being connected, will they still be saved when a connect command is executed, or will they be omitted if i do not use this way?

aniampio: @simonwicky Can you maybe help here Nick on how this should be done properly? 

Side note: In general, any changes to the params should be only applied after reconnecting imo

x64x2 (commented):
50% vibecoded

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -38,6 +38,9 @@ pub struct VpnServiceConfig {
     pub enable_custom_dns: bool,
     pub custom_dns: Vec<IpAddr>,
     pub network_stats: NetworkStatisticsConfig,
+    pub poisson_parameter: Option<u32>,
+    pub average_packet_delay: Option<u32>,
+    pub message_sending_average_delay: Option<u32>,
```
simonwicky: We're starting to have a lot of mixnet client config options there. Can you isolate in a struct like we did for the `NetworkStatisticsConfig`? 
Those 3 and `disable_poisson_rate`, `disable_background_cover_traffic`, `min_mixnode_performance` and `min_gateway_mixnet_performance` please?

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -82,6 +85,18 @@ impl fmt::Display for VpnServiceConfig {
...
+        writeln!(f, "poisson_parameter: {:?}", self.poisson_parameter)?;
+        writeln!(
+            f,
+            "average_packet_delay: {} ms, message_sending_average_delay: {} ms",
+            self.average_packet_delay
+                .map(|v| format!("{v}"))
+                .unwrap_or_else(|| "<None>".to_string()),
+            self.message_sending_average_delay
+                .map(|v| format!("{v}"))
+                .unwrap_or_else(|| "<None>".to_string())
```
simonwicky: nit: `None` is a bit misleading here imo, because there still is a delay even if the options are not set. Could we write `default` or something in that spirit instead?

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/mod.rs
```diff
@@ -123,7 +124,7 @@ impl Default for TunnelConstants {
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
```
simonwicky: Unnecessary changes in that file, you can revert the whole file

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -245,6 +245,26 @@ message Threshold {
   uint32 min_performance = 1;
 }
 
+message ConnectRequest {
```
simonwicky: I don't think this is needed anymore, rebase artifact?

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -262,6 +282,9 @@ message VpnServiceConfig {
   bool enable_custom_dns = 17;
   IpAddrList custom_dns = 16;
   NetworkStatsConfig network_stats = 18;
+  optional uint32 poisson_parameter = 24;
```
simonwicky: Group the options like I've said in a comment above

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -245,6 +245,26 @@ message Threshold {
...
+  bool disable_background_cover_traffic = 7;
+  bool enable_credentials_mode = 8;
+  UserAgent user_agent = 12;
+}
+
+message SetPoissonParameterRequest {
+  uint32 poisson_parameter = 1;
+}
+
+message SetPoissonParameterResponse {}
```
simonwicky: This doesn't look used. Remove it if it isn't

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -245,6 +245,26 @@ message Threshold {
...
+  bool enable_two_hop = 5;
+  bool enable_bridges = 21;
+  bool netstack = 13;
+  bool disable_poisson_rate = 6;
+  bool disable_background_cover_traffic = 7;
+  bool enable_credentials_mode = 8;
+  UserAgent user_agent = 12;
+}
+
+message SetPoissonParameterRequest {
```
simonwicky: Can you group the message you added together please? With some comments maybe?

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -843,10 +869,17 @@ message NetworkStatisticsIdentity {
...
+    uint32 delay_ms = 1;
+}
+message SetDisablePoissonRateRequest {
+  bool disable = 1;
+}
 service NymVpnService {
   // Get info regarding the nym-vpnd in general, like version etc.
   rpc Info (google.protobuf.Empty) returns (InfoResponse) {}
-
+  rpc SetPoissonParameter (SetPoissonParameterRequest) returns (google.protobuf.Empty);
```
simonwicky: Same comment as above, group what you're adding, add some comments and put them somewhere where they make sense

nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs
```diff
@@ -75,7 +77,51 @@ impl RpcClient {
...
+        };
+
+        self.0
+            .set_poisson_parameter(request)
+            .await
+            .map_err(Error::Rpc)?;
+        Ok(())
+    }
+    /// Sets the average per-mixnode packet delay (in milliseconds)
+    pub async fn set_average_packet_delay(&mut self, delay_ms: u32) -> Result<()> {
```
simonwicky: Please leave an empty line between each `fn`

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
```diff
@@ -38,6 +38,9 @@ pub struct VpnServiceConfig {
     pub enable_custom_dns: bool,
     pub custom_dns: Vec<IpAddr>,
     pub network_stats: NetworkStatisticsConfig,
+    pub poisson_parameter: Option<u32>,
```
simonwicky: Thinking about it, we need another name for that option, looking at it, I have zero clue of what it is

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
```diff
@@ -35,6 +36,56 @@ pub struct SetParams {
     /// Enable Circumvention Transport (CT) wrapping for the connection to the entry gateway in two hop wireguard mode.
     #[arg(long, alias = "ct", value_parser = clap::value_parser!(BooleanOption))]
     circumvention_transports: Option<BooleanOption>,
+    /// Set the average delay for a loop cover packet (milliseconds)
+    #[arg(
+        long,
+        value_name = "MILLISECONDS",
+        value_parser = ValueParser::from(|s: &str| -> Result<u32, String> {
```
simonwicky: @pronebird @trojanfoe How do feel about the parameters validation on that layer? 
I reckon it should happen in the daemon and not here

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
```diff
@@ -35,6 +36,56 @@ pub struct SetParams {
...
+            let val: u32 = s.parse().map_err(|_| format!("Invalid integer: {}", s))?;
+            if !(0..=200).contains(&val) {
+                return Err(format!("Value must be between 0 and 200 (got {val})"));
+            }
+            Ok(val)
+        })
+    )]
+    pub loop_cover_stream_average_delay: Option<u32>,
```
simonwicky: Looking at the rest, these don't need to be `pub`

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
```diff
@@ -35,6 +36,56 @@ pub struct SetParams {
...
+            }
+            Ok(val)
+        })
+    )]
+    pub message_sending_delay: Option<u32>,
+
+    #[arg(
+        long,
+        help = "Disable Poisson process rate limiting for real traffic",
+        value_parser = BooleanOption::custom_parser("on","off")
```
simonwicky: Use `clap::value_parser!(BooleanOption))` instead, like the other options

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
```diff
@@ -150,6 +150,10 @@ pub enum VpnServiceCommand {
         oneshot::Sender<Result<NetworkStatisticsIdentity, StatisticsControllerError>>,
         (),
     ),
+    SetPoissonParameter(oneshot::Sender<()>, u32),
```
simonwicky: I mentioned it somewhere else, but this needs another name

nym-vpn-core/crates/nym-vpnd/src/service/config/v4.rs
```diff
@@ -68,6 +68,9 @@ impl TryFrom<VpnServiceConfig> for nym_vpn_lib_types::VpnServiceConfig {
             enable_custom_dns: value.enable_custom_dns,
             custom_dns,
             network_stats: nym_vpn_lib_types::NetworkStatisticsConfig::from(value.network_stats),
+            average_packet_delay: None,
+            message_sending_average_delay: None,
+            poisson_parameter: None,
```
simonwicky: At some point you need to add your options to a version of `VpnServiceConfig`. 
Current latest is v4, but it's getting code frozen for the release on Thursday Dec 11th so you will probably need a v5

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
```diff
@@ -187,7 +175,21 @@ impl VpnServiceConfigManager {
             self.save_config_and_send_event().await;
         }
     }
-
+    #[allow(unused)]
```
simonwicky: Empty line between the `fn`s please

nym-vpn-core/crates/nym-vpnd/src/main.rs
```diff
@@ -109,7 +109,6 @@ async fn run_vpn_service(args: CliArgs) -> anyhow::Result<()> {
 
     #[cfg(not(windows))]
     run_standalone(run_parameters, remove_log_file_signal, shutdown_token).await?;
-
     let _worker_guard = if let Some(setup) = logging_setup {
```
simonwicky, line 113: Unnecessary changes here, you can revert the whole file

nym-vpn-core/crates/nym-vpnd/src/command_interface.rs
```diff
@@ -129,7 +129,69 @@ impl NymVpnService for CommandInterface {
 
         Ok(tonic::Response::new(()))
     }
+    async fn set_poisson_parameter(
```
simonwicky: Empty line between `fn`s please

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
```diff
@@ -973,6 +995,38 @@ impl NymVpnService {
         self.config_manager.set_netstack(netstack).await;
         self.update_tunnel_settings_with_throttle();
     }
+    async fn handle_set_poisson_parameter(&mut self, value: u32) {
+        // Update the Poisson parameter inside the configuration manager
+        self.config_manager.set_poisson_parameter(value).await;
+
+        // Update tunnel settings with throttle
+        self.update_tunnel_settings_with_throttle();
```
simonwicky: @trojanfoe How are we handling changes that are triggering reconnection and changes that aren't? `TunnelSettings::diff`? 
@aniampio Do we want to always reconnect on the new settings? What's the plan there?

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
trojanfoe: @simonwicky Yes, with helper functions added to `TunnelSettingsDiff` to make it easier to use.  In Connecting and Connected state we general always reconnect on a change to tunnel settings, unless there are explicit exceptions (Allow LAN, for example).  I think these settings will require a reconnection in which case no more logic is required when handling `TunnelCommand::SetTunnelSettings`.

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: fixed

nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
trojanfoe: We currently have a `disable_poisson_rate` field in the `VpnServiceConfig` but it doesn't appear to be used for anything useful.

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
trojanfoe: The type should be `u32`.  Is there a reason why it isn't?

nym-vpn-core/crates/nym-vpn-lib-types/src/service.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: fixed

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
nick1231321: fixed

nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
nick1231321: Fixed

nym-vpn-core/crates/nym-vpnd/src/command_interface.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
nick1231321: Fixed

pronebird (commented):
> @pronebird reviewed 1 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039)* status: 1 of 21 files reviewed, 29 unresolved discussions (waiting on @aniampio, @nick1231321, @simonwicky, and @trojanfoe)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 153 at r3](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039#-OgXwz5n7Uko4gZwCRPe:-OgXwz5n7Uko4gZwCRPf:btdyg1r) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/b5643dc2901de19cd416844fba8845e738d6faae/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L153)):*
> ```XML
>         (),
>     ),
>     SetPoissonParameterForLoopCoverStream(oneshot::Sender<()>, u32),
> ```

Would it not be simpler to pass a struct instead instead of having four calls?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039)* status: 1 of 21 files reviewed, 29 unresolved discussions (waiting on @aniampio, @nick1231321, @simonwicky, and @trojanfoe)

___
*[`nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs` line 153 at r3](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039#-OgXwz5n7Uko4gZwCRPe:-OgXx8ObD2uD7e452d59:bbb6k5s) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/b5643dc2901de19cd416844fba8845e738d6faae/nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs#L153)):*

Would it not be simpler to pass a struct instead instead of having four calls?
</blockquote></details>

For example:

```rs
struct MixnetParametersUpdate {
  poisson: Option<u32>,
  avg_packet_delay: Option<u32>,
  avg_msg_send_delay: Option<u32>,
  disable_poisson_rate: Option<bool>
}
```

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
pronebird: I think if we were to pass a struct and update in bulk, we could check previous and new settings, then reconnect. Otherwise it's really tricky.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4039)* status: 1 of 21 files reviewed, 29 unresolved discussions (waiting on @aniampio, @nick1231321, @simonwicky, and @trojanfoe)

nym-vpn-core/crates/nym-vpnd/src/service/vpn_service.rs
nick1231321: Could you explain it in more detail please @pronebird ?

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
```diff
@@ -266,20 +266,17 @@ impl TunnelStateHandler for ConnectedState {
...
                         }
 
-                        #[cfg(any(target_os = "android", target_os = "ios"))]
-                        let _ = diff;
-
                         shared_state.tunnel_settings = tunnel_settings;
-                        self.disconnect(PrivateActionAfterDisconnect::Reconnect, shared_state).await
+
+                        // Not all changes require the tunnel to be reconnected
+                        if diff.only_allow_lan_changed() || (diff.only_mixnet_performance_options_changed() && shared_state.tunnel_settings.tunnel_type == TunnelType::Wireguard) {
```
simonwicky, line 275: This doesn't look correct. Any change in mixnet traffic config in two hop mode shouldn't trigger a reconnect, not just the performance option no?

Given the throttling we have on the config update, can we end up in a case where we update `allow_lan` AND the `mixnet_config` in which case both part will return false and we will reconnect anyway?

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connecting_state.rs
```diff
@@ -676,19 +678,18 @@ impl TunnelStateHandler for ConnectingState {
...
-                                    shared_state.tunnel_settings.allow_lan = tunnel_settings.allow_lan;
-                                    return NextTunnelState::SameState(self);
-                                }
                             }
                         }
 
                         shared_state.tunnel_settings = tunnel_settings;
 
+                        // Not all changes require the tunnel to be reconnected
+                        if diff.only_allow_lan_changed() || (diff.only_mixnet_performance_options_changed() && shared_state.tunnel_settings.tunnel_type == TunnelType::Wireguard) {
```
simonwicky, line 687: Same comment as connected state

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/mod.rs
```diff
@@ -225,9 +225,15 @@ impl TunnelSettings {
...
+            // We care about just the mixnet performance setting changing.
+            if self.gateway_performance_options.mixnet_min_performance
+                != other.gateway_performance_options.mixnet_min_performance
+            {
+                diff.add(TunnelSettingsDiffFields::MixnetPerformanceOptions);
+            }
         }
         if self.mixnet_client_config != other.mixnet_client_config {
-            diff.add(TunnelSettingsDiffFields::MixnetClientConfig);
+            diff.add(TunnelSettingsDiffFields::MixnetPerformanceOptions);
```
simonwicky, line 236: This is very confusing, why MixnetPerformanceOptions?

nym-vpn-core/crates/nym-vpnc/src/commands/tunnel.rs
simonwicky: IMO bound check have to live in the daemon, otherwise other user of the RpcClient will be able to set illegal value

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
```diff
@@ -333,23 +308,39 @@ impl VpnServiceConfigManager {
         tracing::info!("Using config: {:?}", self.config);
 
         let gateway_options = GatewayPerformanceOptions {
-            mixnet_min_performance: self.config.min_gateway_mixnet_performance,
+            mixnet_min_performance: self.config.mixnet_traffic.min_gateway_mixnet_performance,
             vpn_min_performance: self.config.min_gateway_vpn_performance,
         };
 
+        let to_duration_ms = |v: Option<u32>| v.map(|ms| Duration::from_millis(ms as u64));
```
simonwicky: use `.map(Duration::from_millis)` directly on the `Option`s instead. And if you need `u64` set them as `u64` in the config directly

nym-vpn-core/crates/nym-vpnd/src/service/config/mixnet_traffic.rs
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+
+pub(crate) mod v5 {
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "snake_case")]
+    pub struct MixnetTrafficConfig {
+        pub poisson_parameter_for_loop_cover_stream: Option<u32>,
+        pub average_packet_delay: Option<u32>,
```
simonwicky, line 11: Here. `u64` if you need them in a duration later

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
trojanfoe: > Any change in mixnet traffic config in two hop mode shouldn't trigger a reconnect

That's what this check is for;  If the mixnet performance options have changed and we are using (or going to use) 2-hop mode then we don't reconnect.

> Given the throttling we have on the config update, can we end up in a case where we update allow_lan AND the mixnet_config in which case both part will return false and we will reconnect anyway?

Yes; we could probably apply more logic to check if `allow_lan` was changed and mixnet performance options change in 2-hop mode, but I think it's getting too much then.  If we reconnect more often than required then it's not the end of the world.

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/mod.rs
trojanfoe: Because of how the mixnet performance options are spread across two fields.

nym-vpn-core/crates/nym-vpnd/src/service/config/mixnet_traffic.rs
trojanfoe: That's not strictly true; there is no harm in casting them at conversion time if you are happy 32-bits of milliseconds covers your use-case.

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
```diff
@@ -173,14 +177,56 @@ impl VpnServiceConfigManager {
         }
     }
 
+    fn validate_mixnet_traffic_config(
```
trojanfoe: I would move this into `impl MixnetTrafficConfig`.  There is already enough code in this module.

nym-vpn-core/crates/nym-vpnd/src/command_interface.rs
```diff
@@ -235,15 +235,28 @@ impl NymVpnService for CommandInterface {
...
             .send_and_wait(
                 VpnServiceCommand::SetMixnetTrafficConfig,
                 mixnet_traffic_config,
             )
             .await
-            .map_err(|e| Status::internal(format!("Failed to set mixnet traffic config: {e}")))?;
+            .map_err(|e| {
+                tracing::error!("[set_mixnet_traffic_config] transport error: {e}");
+                Status::internal(format!("[set_mixnet_traffic_config] transport error: {e}"))
+            })?;
```
trojanfoe: This is all very messy.  I would just use `.map_err()` twice and stop logging to simplify it.

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
trojanfoe: Also the validation needs to be performed when loading the config from file, as the user may have tampered with it.

nym-vpn-core/crates/nym-vpnd/src/command_interface.rs
nick1231321: fixed

nym-vpn-core/crates/nym-vpnd/src/service/config/config_manager.rs
nick1231321: > I would move this into `impl MixnetTrafficConfig`. There is already enough code in this module.

fixed

trojanfoe: Sorry, that hasn't been fixed.

nick1231321: I will push the changes in a few minutes

nym-vpn-core/Cargo.toml
```diff
@@ -205,39 +205,39 @@ nym-wg-metadata-client = { path = "crates/nym-wg-metadata-client" }
...
-nym-sdk = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-statistics-common = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-topology = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-upgrade-mode-check = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-validator-client = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-wireguard-private-metadata-client = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-wireguard-private-metadata-shared = { git = "https://github.com/nymtech/nym", branch = "develop" }
-nym-wireguard-types = { git = "https://github.com/nymtech/nym", branch = "develop" }
-sqlx-pool-guard = { git = "https://github.com/nymtech/nym", branch = "develop" }
+nym-authenticator-client = { git = "https://github.com/nymtech/nym", branch = "nicolas/sdk-param-support-debug" }
```
trojanfoe: @nick1231321 You can just revert this file altogether and do the `cargo update` to get the latest `nym/develop` changes.

trojanfoe (approved):

simonwicky (approved):

PR #4043, opened 2025-11-26
Use ndk for build variants as defined in the client

nighthawk24:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4043)


nighthawk24:
Not a fix.

PR #4047, opened 2025-11-26
VPN-4546: Blacklist entry gateway after repeated connection failure.

trojanfoe:
## Ticket

JIRA-VPN-4546

## Description

If we fail to connect then blacklist the entry node for 20 minutes so we don't connect to it when choosing a random/country/region gateway.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4047)


tommyv1987 (approved):
LGTM

PR #4061, opened 2025-11-27
feat(tauri): dynamic vpn tunnel config

doums:
Notable changes along this refactoring:
- switch to vpnd tunnel config, listen and handle live config updates from daemon
- handle node selection through vpnd
- enable node selection during tunnel connection or whatever is the current tunnel state (vpnd manages it now)
- add allow LAN option in the settings
- enable "Random" aka Fastest (supposedly) node selection for both entry and exit

### Compromises

Due to the daemon quite limited API for the selected nodes (entry/exit points), I had to make some compromises.
In daemon terminology, selected and configured node include the following data:
- gateway: the ID only
- country: the 2 letter code only
- region: name only

This is far to cover the need we have at UI level, eg. showing the country, region and or city X gateway belong to when a gateway is selected.
When a region is selected, the corresponding country.
In order to achieve this, it requires making lookup in the gateways data list as additional steps and during runtime, at UI level. Also for region to country lookup, the need to introduce and rely on 3rd party [db](https://github.com/country-regions/country-region-data).
This is far to be optimal and require quirks.
The daemon API would need to be improved on this.

### TODO

Daemon API seems to be more ahead of what we currently support in client App.
For the exit node, vpnd includes this variant `ExitPoint::Address` but this is not supported (never been scoped afaik) in client app side.
For now if you select such address as exit point (via vpnc I presume) and then you open the app, an error will be thrown, and the app will discard the vpn config from daemon -> not great but no other choice (for this PR scope).
Ofc this will need to be handled properly, but in a next iteration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4061)


trojanfoe (approved):
LGTM

agentpietrucha (approved):
code looks good to me

PR #4137, opened 2025-12-05
Avoid cloning the RPC client.

trojanfoe:
Avoid cloning the RPC client and instead use interior mutability and pass the rpc client around by reference.

Only tested on macOS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4137)

### DIFF
--- a/nym-vpn-app/src-tauri/Cargo.lock
+++ b/nym-vpn-app/src-tauri/Cargo.lock
@@ -3641,6 +3641,7 @@ dependencies = [
  "prost-types",
  "thiserror 2.0.17",
  "time",
+ "tokio",
  "tokio-stream",
  "tonic",
  "tonic-prost",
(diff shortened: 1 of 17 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4137)* status: 0 of 17 files reviewed, 1 unresolved discussion (waiting on @trojanfoe)

___
*[`nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs` line 22 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/4137#-OfxezYWBKUSuWuSJNLs:-OfxezYWBKUSuWuSJNLt:b-jt043a) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/f5fca7100d4c0ce5571ced551224e8e12aa497d3/nym-vpn-core/crates/nym-vpn-proto/src/rpc_client.rs#L22)):*
> ```XML
> 
> #[derive(Debug)]
> pub struct RpcClient(Arc<Mutex<ServiceClient>>);
> ```

`NymVpnServiceClient` is clonable. Which is cheaper considering that it holds `Arc<Channel>` inside. There is no need to synchronize access to it.

PR #4140, opened 2025-12-07
Support BUILD_TOP environment for sandboxed builds

poelzi:
## Description

The build.rs assumes the source directory is writable, which breaks sandboxed builds like nix. Support BUILD_TOP environment variable, so build systems can simply set it to their build directory.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4140)

### DIFF
--- a/nym-vpn-core/crates/nym-wg-go/build.rs
+++ b/nym-vpn-core/crates/nym-wg-go/build.rs
@@ -4,22 +4,29 @@
 use std::{env, fs, path::PathBuf};
 
 fn main() {
-    let manifest_path = env::var_os("CARGO_MANIFEST_DIR").expect("manifest dir is not set");
     let target = env::var("TARGET").expect("target is not set");
     let target_os = env::var("CARGO_CFG_TARGET_OS").expect("target os is not set");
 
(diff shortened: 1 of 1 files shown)


pronebird (commented):
Should it not be `NIX_BUILD_TOP` according to https://nixos.wiki/wiki/Environment_variables?

poelzi:
@pronebird yes, for nix specific builds yes. I thought it might be better to make this generic. I can change this, which would simplify the nix pkgs

pronebird:
> @pronebird yes, for nix specific builds yes. I thought it might be better to make this generic. I can change this, which would simplify the nix pkgs

This can be made generic for sure. I had impression that you've tested this with nix.

However note that `build-wireguard-go.sh` outputs artifacts in the very same directory, so this patch alone is not sufficient to make it all work.

Further thoughts: @trojanfoe we could probably stop touching file system from here and remove the call to `canonicalize` which is the reason for build failure when build dir does not exist.

pronebird (approved):

pronebird:
Let's merge it as is. Nice first step 👍 Thanks for contribution! 🚀

PR #4148, opened 2025-12-08
VPN-4633: SOCKS5 proxy is not shutting-down correctly.

trojanfoe:
## Ticket

JIRA-VPN-4633

## Description

SOCKS5 proxy is not shutting-down correctly.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4148)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
+++ b/nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
@@ -807,6 +807,7 @@ message DeleteLogFileResponse {
 }
 
 // Enable SOCKS5 proxy request
+// TODO: Change arguments to EnableSocks5Request (at the time of writing I didn't want to disturb the client code)
 message EnableSocks5Request {
   Socks5Settings socks5_settings = 1;
   HttpRpcSettings http_rpc_settings = 2;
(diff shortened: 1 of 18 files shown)


trojanfoe:
@pronebird I am still looking into the root-cause of this issue, which is the obvious _file in use_.  So I don't think we are closing down properly.

trojanfoe:
@pronebird I cannot reproduce the root cause on Windows or macOS, so I would like to merge this change until we have more reports from actual users.

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -807,6 +807,7 @@ message DeleteLogFileResponse {
 }
 
 // Enable SOCKS5 proxy request
+// TODO: Change arguments to EnableSocks5Request (at the time of writing I didn't want to disturb the client code)
```
pronebird, line 810: What needs to be changed?

nym-vpn-core/crates/nym-vpn-proto/src/conversions/socks5.rs
```diff
@@ -13,19 +16,19 @@ impl TryFrom<proto::Socks5Status> for Socks5Status {
...
 
-        let socks5_settings = value
+        let socks5_settings: Socks5Settings = value
             .socks5_settings
-            .map(Socks5Settings::from)
-            .unwrap_or_else(|| Socks5Settings {
-                listen_address: String::new(),
-            });
+            .ok_or(ConversionError::NoValueSet(
+                "EnableSocks5Request.socks5_settings",
```
pronebird, line 22: Wow, good spot

pronebird (approved):

PR #4167, opened 2025-12-10
VPN-4640: Select SOCKS5 exit node by socks5 score.

trojanfoe:
## Ticket

JIRA-VPN-4640

## Description

Select the SOCKS5 exit node to use by the new socks5 score.

This change reorganises the gateway selection logic by moving the selection of entry/exit gateways into `GatewayList` and remove the dangling `min_score`, `blacklisted_gateways` and `residential_exit` parameters.  Now everything is passed using `GatewayFilters`.

It provides methods to select the "best" gateways, based on performance, reducing complexity in the calling gateway selection code.

It also provides functionality to populate the new SOCKS5 score, making use of the mixnet score if it's not available.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4167)


nym-vpn-core/crates/nym-gateway-directory/src/entries/gateway.rs
```diff
@@ -652,6 +654,266 @@ impl GatewayList {
...
+            EntryPoint::Country {
+                two_letter_iso_country_code,
+            } => {
+                tracing::debug!(
+                    "Selecting entry gateway by country: {two_letter_iso_country_code}"
+                );
+
+                let filters = base_filters
+                    .iter()
+                    .chain(&vec![GatewayFilter::Country(
```
trojanfoe: `GatewayFilters` should only have one of each type, so I think we should pass them around as `&HashSet<GatewayFilter>` to enforce?

pronebird (approved):

PR #4226, opened 2025-12-16
Fix bug of two paths for discovery file

neacsu:
## Ticket

JIRA-VPN-4563

## Description

A bug where discovery file is never properly updated because the fetched updates are written to one path (e.g. `/etc/nym/canary/canary_discovery.json`) and read from other (`/etc/nym/canary/networks/canary_discovery.json`). Fix the bug so that we only use `/etc/nym/{network}` directory for those files.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4226)


trojanfoe (approved):

pronebird:
Make sure it works on iOS and Android if it's related

PR #4283, opened 2025-12-20
APPLE: Show 'Get started' text on macOS when no credentials

tw0po1nt:
## Description

macOS's behavior on no credentials currently diverges from iOS. Update the macOS app to properly update the connect button to show 'Get started' when no credential has been imported.

Also removes an unused function

## Checklist:

- [ ] Changelog

Screenshot 2025-12-20 at 10 29 06 AM

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4283)

### DIFF
--- a/nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
+++ b/nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
@@ -333,7 +333,11 @@ extension HomeViewModel {
                 hasInternet: networkMonitor.isAvailable,
                 subscriptionDidExpire: isLastErrorSubscriptionExpired()
             )
+            #if os(iOS)
             connectButtonState = ConnectButtonState(tunnelStatus: newStatus)
+            #elseif os(macOS)
+            updateConnectButtonStateIfMnemonicImported()
+            #endif
 
             if let lastError {
                 statusInfoState = .error(message: lastError.localizedDescription)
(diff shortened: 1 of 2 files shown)


nym-vpn-apple/Home/Sources/Home/HomeViewModel.swift
```diff
@@ -333,7 +333,11 @@ extension HomeViewModel {
                 hasInternet: networkMonitor.isAvailable,
                 subscriptionDidExpire: isLastErrorSubscriptionExpired()
             )
+            #if os(iOS)
             connectButtonState = ConnectButtonState(tunnelStatus: newStatus)
+            #elseif os(macOS)
+            updateConnectButtonStateIfMnemonicImported()
```
tw0po1nt: @rokas-ambrazevicius this gets the job done, but I'm curious on your thoughts if there is a better way to accomplish this. I did it this way because I noticed this method, but it was only being called in iOS, not macOS.

rokas-ambrazevicius: I think there is a better solution.

The `func setupIsMnemonicImportedObserver()` seems to be in `HomeViewModel+iOS.swift`, moving it out of iOS only file into `HomeViewModel` + moving it out of the iOS only part in `setup()`, should fix the issue.

What do you think about this approach?

tw0po1nt: Good idea. This is done!

rokas-ambrazevicius (approved):

PR #4463, opened 2026-01-23
[Android] Social login support

mrkiwi:
## Ticket

JIRA-VPN-XXXX

## Description

Update Login screen UI & error states.
Add Social button to login screen.
Expose deeplinks from core.
Add deeplink support for Social button.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4463)

### DIFF
--- a/nym-vpn-android/core/src/main/java/net/nymtech/vpn/backend/NymBackend.kt
+++ b/nym-vpn-android/core/src/main/java/net/nymtech/vpn/backend/NymBackend.kt
@@ -49,7 +49,10 @@ import nym_vpn_lib.login
 import nym_vpn_lib.startVpn
 import nym_vpn_lib.stopVpn
 import nym_vpn_lib_types.AccountControllerState
+import nym_vpn_lib_types.DeeplinkClient
+import nym_vpn_lib_types.DeeplinkKind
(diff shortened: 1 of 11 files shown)


nym-vpn-android/core/src/main/java/net/nymtech/vpn/backend/NymBackend.kt
```diff
@@ -301,6 +304,26 @@ class NymBackend private constructor(private val context: Context) :
 		}
 	}
 
+	override suspend fun getDeeplink(): String {
```
trojanfoe, line 307: We are going to start reusing this function for the new "Privy for existing users" functionality, so you might have to lift some of these parameters into the caller in order to re-use this function.  It's OK for now though.

trojanfoe (approved):

nym-vpn-android/core/src/main/java/net/nymtech/vpn/backend/NymBackend.kt
mrkiwi: It's mostly the UI update and changes PR. Will update the whole links later, since the Privy will not be in the next RC

PR #4497, opened 2026-01-28
NYM-356: Privy For Existing Users.

trojanfoe:
## Ticket

JIRA-NYM-356

## Description

Summary of changes:

- We will use the existing Privy landing pages for the Privy-for-existing-users functionality.
- The URL will contain an additional `existing_user=1` parameter.
- Apps need to use the new `DeeplinkKind::PrivyExisting` parameter when calling `get_deeplink()`.  Handling the callback URL from the website will be the same.

I have made the assumption that we can continue using a single mnemonic for this new mode of operation 🤞

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497)


pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @trojanfoe).

___
*[`nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs` line 54 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497#-Ok3m5UR5xcbYAnIIQgx:-Ok3m5UR5xcbYAnIIQgy:b-pz0vb4) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/9eb77bddd4841c717eee5415b394e12f1ef6dfca/nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs#L54)):*
> ```XML
> pub enum DeeplinkKind {
>     Privy,
>     PrivyExisting,
> ```

This seems fragile. Why would apps need to worry about that? How would apps know which to choose and what happens if they use the wrong one?

pronebird (commented):
> @pronebird reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497)* status: all files reviewed, 1 unresolved discussion (waiting on @trojanfoe).

trojanfoe (commented):
> @trojanfoe made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird).

___
*[`nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs` line 54 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497#-Ok3m5UR5xcbYAnIIQgx:-Ok3mgIXBjed9ol1sMqP:bonjgl8) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/9eb77bddd4841c717eee5415b394e12f1ef6dfca/nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs#L54)):*

This seems fragile. Why would apps need to worry about that? How would apps know which to choose and what happens if they use the wrong one?
</blockquote></details>

The apps are driving this process.  They are allowing the user to either login using Socials, or to connect Socials to their existing account.

An alternative might be an additional parameter, but that is potentially more confusing, as it would need to relate to the "kind" parameter.  I think once we start supporting more "kinds" of deeplink this interface will settle down.

If apps choose the wrong type then it will break, just like if code does anything wrong -- something will break.

pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497)* status: all files reviewed, 1 unresolved discussion (waiting on @trojanfoe).

___
*[`nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs` line 54 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/4497#-Ok3m5UR5xcbYAnIIQgx:-Ok3tjQEEcvutjDKp5Ei:bcke4dy) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/9eb77bddd4841c717eee5415b394e12f1ef6dfca/nym-vpn-core/crates/nym-vpn-lib-types/src/account/deeplink.rs#L54)):*
<details><summary><i>Previously, trojanfoe (Andy Duplain) wrote…</i></summary><blockquote>

The apps are driving this process.  They are allowing the user to either login using Socials, or to connect Socials to their existing account.

An alternative might be an additional parameter, but that is potentially more confusing, as it would need to relate to the "kind" parameter.  I think once we start supporting more "kinds" of deeplink this interface will settle down.

If apps choose the wrong type then it will break, just like if code does anything wrong -- something will break.
</blockquote></details>

Thanks for explanation. Let's put breaks on this and discuss the whole concept of using only one mnemonic. I sense that we shouldn't be replacing nym mnemonic with privy because this would create a lot of confusion for users.

trojanfoe:
@pronebird Converted to draft as I need to do much more work on account storage unit tests.

trojanfoe:
This work has been simplified and will be submitted via a new PR.

PR #4498, opened 2026-01-28
Refactor nym-vpn-network-config

pronebird:
## Description

This PR refactors nym-vpn-network-config and decouples massive god types onto smaller:

- Persistent stores for envs, discovery, network details responsible for initializing and persisting data on disk
- `Fetcher` type for network interaction with API
- Introduce `NetworkCache` for managing a collection of persistent stores and refreshing them using `Fetcher`.
- Ensure that `Fetcher` uses new API endpoints when discovery changes.
- Allow `VpnApiClient` to accept `None` as user agent which should fallback to default user agent (useful in tests)
- Strip `async` from some of HTTP client methods that had no actual async calls in them.
- Remove `NymNetwork` type which is basically a superficial wrapper around `NymNetworkDetails`
- Remove unused `nym-vpn-network-config/src/response.rs`
- Switch persistent stores from using filetime to embedding UTC timestamp into the cached JSON. This seems to be better approach as it doesn't involve changing filesystem metadata. It also avoids using "date time in the past" when storing defaults in order to mark data stale.
- Move persistent stores from `/etc/nym/<network>` back to `/etc/nym/networks/<network>` sub-directory as it used to be. This keeps all network data together under `/etc/nym/networks`. (Reverts https://github.com/nymtech/nym-vpn-client/pull/4226, there is a clean up routine built into `NetworkCache`)
- Add test coverage

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4498)


nym-vpn-core/crates/nym-vpn-network-config/Cargo.toml
```diff
@@ -27,6 +26,7 @@ serde.workspace = true
 serde_json.workspace = true
 tempfile.workspace = true
 time = { workspace = true, features = ["serde-human-readable"] }
+chrono = { workspace = true, features = ["serde"] }
```
simonwicky: Could we do everything with `time` instead of chrono? 
Even if it's not used there, I remember that sqlx doesn't like having both time and chrono, so using only of them would be preferable I'd say

nym-vpn-core/crates/nym-vpn-network-config/src/discovery_refresher.rs
```diff
@@ -11,205 +11,107 @@ use tokio_util::sync::CancellationToken;
...
-                            trace_err_chain!(err, "Failed to refresh discovery file");
-                            self.events_tx
-                                .send(DiscoveryRefresherEvent::Error(Error::RefreshDiscoveryFile))
-                                .ok();
+                            trace_err_chain!(err, "failed to refresh network cache");
                         }
-                        _ => {}
                     }
                 }
+                _ = self.cancel_token.cancelled() => {
```
simonwicky, line 102: Don't we want a biased `select` towards that cancellation?

nym-vpn-core/crates/nym-vpn-network-config/src/lib.rs
```diff
@@ -425,6 +456,106 @@ impl Error {
...
+        let envs = ["mainnet", "sandbox", "canary", "evil"];
+
+        for env in envs {
+            let base_dir = cache_dir.path().join(env);
+            tokio::fs::create_dir(&base_dir).await.unwrap();
+            let _ = tokio::fs::File::create(base_dir.join(format!("{env}.json"))).await;
+            let _ = tokio::fs::File::create(base_dir.join(format!("{env}_discovery.json"))).await;
+        }
+
+        let _ = tokio::fs::File::create(cache_dir.path().join(format!("test.txt"))).await;
```
simonwicky: nit : useless `format!`

simonwicky (approved):
Just a few nits that I think would be good to change, but nothing blocking

nym-vpn-core/crates/nym-vpn-network-config/Cargo.toml
pronebird: I've heard about that but that does not seem to be a problem despite presence of both. I find `chrono` more ergonomic and not quite certain how to even use `time` for what I need it for.

nym-vpn-core/crates/nym-vpn-network-config/src/discovery_refresher.rs
pronebird: Reodering branches in tokio select does not set the bias, the use of `biased;` does. Tokio selects futures randomly otherwise. The ordering was done to move less interesting stuff to the bottom.

In general we don't have to be biased towards cancellation unless we have a really busy loop. I don't think this is the case here.

nym-vpn-core/crates/nym-vpn-network-config/src/lib.rs
pronebird: Fixed

nym-vpn-core/crates/nym-vpn-network-config/Cargo.toml
pronebird: After talking offline, I realized the use case is so simple that we can stick to time to reduce number of dependencies. We still have `chrono` around though but only used at one occurrence in `nym-vpn-api-client`.

PR #4669, opened 2026-02-18
E2E test workflow

dynco-nym:
## Ticket

JIRA-VPN-XXXX

## Description

I'm developing a workflow for `dyn/test-harness-latest` which will have `workflow_dispatch` trigger (manually triggered). 

For that, the [workflow file needs to already be in the main branch](https://github.com/orgs/community/discussions/25746#discussioncomment-3249049)for me to be able to trigger it (and thus test it).

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4669)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,0 +1,248 @@
+name: e2e-test
+
+on:
+  workflow_dispatch:
+    inputs:
+      test_filter:
+        # No `default:` here — the inline fallback `inputs.test_filter || 'basic_functionality'`
+        # in the run-tests step handles both manual runs (empty field) and scheduled runs
+        # (no inputs at all). A `default:` would only cover workflow_dispatch, creating a
+        # second source of truth that can silently diverge.
+        description: "Test filter (passed to test-manager). Defaults to basic_functionality."
+        required: false
+      instance_type:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,248 @@
...
+        uses: actions/checkout@v4
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y \
+            gcc make pkg-config \
+            libdbus-1-dev libmnl-dev libnftnl-dev \
+            protobuf-compiler libssl-dev libpcap-dev
+
+      - name: Install Rust toolchain
```
rokas-ambrazevicius: https://github.com/nymtech/nym-vpn-client/blob/056f69a4a838756982484606bb1ab1a8df47fba4/.github/workflows/build-nym-vpn-core-ios.yml#L37

I think you should be using the rust version core uses? Or does it not matter?

rokas-ambrazevicius (approved):

.github/workflows/e2e-test.yml
dynco-nym: done

PR #4694, opened 2026-02-20
Add split tunneling on macOS

pronebird:
## Ticket

JIRA-NYM-552

## Description

- Introduce split tunneling (ST) using BPF. In principle it works the following way:
   - ST tun device is created. It's used for collecting outgoing traffic.
   - PF is configured to redirect all outgoing traffic from exit tun (which is collecting all tunneled traffic) into ST tun
   - PKTAP is created and attached on ST tun for capture of outbound packets including additional metadata such as process ID (PID) that emitted the packet
   - BPF is attached on primary interface and collects incoming traffic, injecting it back into ST tun. There is some NAT going on to ensure that inbound traffic makes it back to VPN tunnel interface.
   - BPF is attached on VPN tunnel interface (exit tun) for the purpose of packet injection (write-only)
   - `/usr/bin/eslogger` is spawned in background, streaming JSON with events providing information on spawned and terminated processes system wide. This is used for PID collection for applications excluded from VPN tunnel. Matching is based on path to executable.
   - ST tun maintains the same MTU as exit tunnel using route manager's interface change events.
   - Once new outgoing packet arrives on ST tun, it appears in the packet stream from the corresponding PKTAP stream. Then based on PID exclusions, the packet is either injected into VPN tunnel interface or directly on primary interface. Injection involves patching up source/destination IPs and checksum update. My understanding that injection via BPF circumvents firewall thus the packets appear directly on the target interface.
   - Inbound packets are collected on default interface and then re-injected into ST tun after updating the destination IP to VPN interface IP. This way the packets that bypassed the tunnel make their way back to the VPN interface.
- Add ST settings. `SplitApp` stores executable path only for the time being. Down the road we'll store the routing method too when we start serving traffic over both mixnet and wg at the same time.
- Add `nym-ifconfig` crate, which implements some of low-level methods for configuring network interfaces on macOS and Linux. This PR does not migrate the remainder of project to it but it should be used going forward to avoid shelling out when configuring network interfaces. I thought it's worth bringing it on board now since it has a good test coverage and it's been laying around for a long time and I could use some help on shaping it up into something better.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4694)


trojanfoe (approved):
Amazing work @pronebird

PR #4800, opened 2026-03-05
feat - byon adjustments

tommyv1987:
- Use aliases to keep old version working
- Update naming:
`nym-vpnc account set "<mnemonic>" --location blockchain` 
`nym-vpnc account obtain-ticketbooks --amount 1 --source smartcontract` 
`nym-vpnc account obtain-ticketbooks --amount 1`

## Ticket

[JIRA-VPN-732](https://nymtech.atlassian.net/browse/NYM-732)

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4800)


nym-vpn-core/crates/nym-vpnc/src/commands/account.rs
```diff
@@ -30,11 +30,18 @@ pub enum Command {
     },
     /// Get account balance
     Balance,
-    /// Attempt to obtain additional accounts for a 'decentralised' account
-    DecentralisedObtainTicketbooks {
+    /// Attempt to obtain additional accounts for a 'blockchain' account
+    #[clap(
+        name = "obtain-ticketbooks",
```
jstuczyn, line 35: I think this is reduntant because it will be just obtained from the command name itself

nym-vpn-core/crates/nym-vpnc/src/commands/account.rs
```diff
@@ -30,11 +30,18 @@ pub enum Command {
...
+        alias = "decentralised-obtain-ticketbooks"
+    )]
+    ObtainTicketbooks {
         /// Amount of ticketbooks (per type) to attempt to obtain
         #[arg(long, default_value_t = 1)]
         amount: u64,
+        /// Ticketbook source (parsed for future compatibility?, currently ignored)
+        #[arg(long, value_enum, default_value_t = TicketbookSource::Smartcontract)]
```
jstuczyn, line 43: if it's ignored, why are we adding it here? 
if it's for the optics, maybe the better variant name would be `blockchain` rather than `smart-contract`, since it might be more universally understadable, idk.

jstuczyn (approved):

PR #4849, opened 2026-03-10
Sign windows RCs

neacsu:
## Ticket

JIRA-NYM-735

## Description

Sign windows release candidates and sign the daemon as well (not only the client application).
Also modify the skip of signature check so that on Windows it depends on whether the daemon was itself signed or not.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4849)

### DIFF
--- a/.github/workflows/build-nym-vpn-app-windows.yml
+++ b/.github/workflows/build-nym-vpn-app-windows.yml
@@ -269,22 +269,22 @@ jobs:
         working-directory: nym-vpn-app/src-tauri
         if: ${{ inputs.sign }}
         shell: bash
+        env:
+          SSL_COM_USERNAME: ${{ secrets.SSL_COM_USERNAME }}
+          SSL_COM_PASSWORD: ${{ secrets.SSL_COM_PASSWORD }}
+          SSL_COM_TOTP_SECRET: ${{ secrets.SSL_COM_TOTP_SECRET }}
         run: |
           yq eval --inplace '.bundle.windows +=
           {
             "signCommand": {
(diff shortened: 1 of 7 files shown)


.github/workflows/build-nym-vpn-app-windows.yml
```diff
@@ -279,16 +275,12 @@ jobs:
...
               "args": [
-                "/c/actions-runner/_work/nym-vpn-client/nym-vpn-client/nym-vpn-app/src-tauri/CodeSignTool.sh",
-                "sign",
-                "-username ${{ secrets.SSL_COM_USERNAME }}",
-                "-password ${{ secrets.SSL_COM_PASSWORD }}",
-                "-credential_id ${{ steps.get_credential_ids.outputs.SSL_COM_CREDENTIAL_ID }}",
-                "-totp_secret ${{ secrets.SSL_COM_TOTP_SECRET }}",
-                "-program_name NymVPN",
-                "-input_file_path",
+                "/c/actions-runner/_work/nym-vpn-client/nym-vpn-client/nym-vpn-app/src-tauri/scripts/sign.sh",
```
pronebird, line 282: Does it have to be full path or can we use `scripts/sign.sh`?

neacsu: According to the comment above this step, I would assume it's safer this way
```
Absolute path to CodeSignTool is required since the tauri signcommand does not execute in the same context as the shell
```

neacsu:
Close in favor of https://github.com/nymtech/nym-vpn-client/pull/4856

PR #4898, opened 2026-03-17
AccountController Syncing retry delay

jmwample:
## Description

Add delay in accont controller retry to prevent rapid error spam. Currently the account controller has a finite number of attempts, but no delay between them. So if an error occurs it will immediately retry. I saw this during the disconnecting state where the account controller attempts to sync. It spams network requests that fail because the network is unreachable. 

This PR adds a bounded exponential backoff delay to the retry attempts. 

```
// before 
2026-03-13T19:50:59.932644Z DEBUG nym_vpn_account_controller::state_machine::syncing_state: Error trying to get account summary, retrying : ApiRequestError

// after
2026-03-13T19:50:59.932644Z DEBUG nym_vpn_account_controller::state_machine::syncing_state: Error trying to get account summary attempt 10, retrying after 8s : ApiRequestError
```

The timeout for the mocked integration tests is also increased to 60 seconds as there are (currently) two tests that require exhausting the retries in order to assert a failure:

```rust
async fn desynced_device_test() -> anyhow::Result<()> {}

async fn account_with_inactive_sub_test() -> anyhow::Result<()> {}
```

## Checklist:

- [x] add in reasonable delay parameters to the account controller syncing state retry logic

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4898)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
@@ -33,7 +33,7 @@ pub(super) mod requesting_zknym_state;
 const MAX_SYNCING_ATTEMPTS: u32 = 10;
 const SYNCING_STATE_CONTEXT: &str = "SYNCING_STATE";
 
-// bounded exponential backoff for retries [0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0]
+// bounded exponential backoff for retries [0.25, 0.5, 1.0, 2.0, 4.0, 8.0, 8.0, 8.0, 8.0, 8.0, 8.0] = 55.75s max wait
 const RETRY_BACKOFF: Duration = Duration::from_millis(250);
 const MAX_BACKOFF_EXPONENT: u32 = 5;
 const BACKOFF_BASE: u32 = 2;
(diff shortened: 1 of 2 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
```diff
@@ -254,7 +261,9 @@ impl<C: ConnectivityMonitor> AccountControllerStateHandler<C> for SyncingState {
                                 tracing::debug!("Error trying to get account summary, exhausted retries : {}", err.to_string());
                                 NextAccountControllerState::NewState(ErrorState::enter(err.into()))
                             } else {
-                                tracing::debug!("Error trying to get account summary, retrying : {}", err.to_string());
+                                let delay = RETRY_BACKOFF * BACKOFF_BASE.pow(min(self.attempts, MAX_BACKOFF_EXPONENT));
+                                tracing::debug!("Error trying to get account summary attempt {}, retrying after {:?} : {}", self.attempts, delay, err.to_string());
+                                let _ = shutdown_token.run_until_cancelled(tokio::time::sleep(delay)).await;
```
simonwicky: This will prevent the AC to be responsive, that backoff should be in `sync_account` directly

nym-vpn-core/crates/nym-vpn-account-controller/tests/integrations/common/mod.rs
```diff
@@ -270,7 +270,7 @@ impl TestBench {
 
         let wait_for_state_fut = state_watcher.wait_for(|state| *state == expected_state);
 
-        let _ = tokio::time::timeout(Duration::from_secs(15), wait_for_state_fut).await;
+        let _ = tokio::time::timeout(Duration::from_secs(60), wait_for_state_fut).await;
```
simonwicky, line 278: nit: We're not testing retries in integration tests, so that bump isn't necessary. 
If you still choose to bump it, please update the comment of the `fn` (it's mentioning the timeout value)

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/mod.rs
pronebird: I think that http client might be sending 3 requests with 1 sec interval. If I remember that right.

jmwample: If configured there are conditions where the http client itself can retry operations, but the account controller sync retry is not limited to HTTP requests. For example HTTP retries are meant to help if an alternative equivalent domain is available. This retry is meant to handle (for example) the situation where account subscription processing has a slight delay on remote and retrying might help resolve the state. 

Delay has been shifted to the `sync_account`  thread.

simonwicky (approved):

PR #4973, opened 2026-03-25
Signal shutdown on connection invalidation

neacsu:
## Ticket

JIRA-NYM-960

## Description

When the client connection is started over XPC, if the daemon is not running, the client still keeps the connection with no-op for reads and writes.
The invalidation handler offers that information immediately, so it needs to be consumed to figure out that the connection wasn't actually established, so it can be instantly dropped and the initial connection test resolves with no response, signifying that the daemon is not up.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4973)

### DIFF
--- a/nym-vpn-core/CHANGELOG.md
+++ b/nym-vpn-core/CHANGELOG.md
@@ -5,15 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [1.27.0] - TBD
+
+- [macOS] XPC as transport layer between clients and daemon (https://github.com/nymtech/nym-vpn-client/pull/4695)
+- [macOS] Authentication layer for windows, feature gated (https://github.com/nymtech/nym-vpn-client/pull/4802)
+- Activate authentication layer on all desktop platforms (https://github.com/nymtech/nym-vpn-client/pull/4856)
+
+### Fixed
+
+- [macOS] XPC client stall when daemon is not running (https://github.com/nymtech/nym-vpn-client/pull/4973)
+
+
 ## [1.26.0] - 2026-03-17
 
 ### Added
 
-- [macOS] XPC as transport layer between clients and daemon (https://github.com/nymtech/nym-vpn-client/pull/4695)
 - [CLI] `nym-vpnc account set` now uses `--location blockchain`; aliases keep legacy `--mode decentralised` and `--mode decentralized` working.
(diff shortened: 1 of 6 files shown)


nym-vpn-core/crates/nym-ipc/src/xpc/client.rs
```diff
@@ -55,8 +55,10 @@ async fn xpc_connect(
 
     let xpc_conn_invalidated = Arc::new(AtomicBool::new(false));
     let xpc_conn_invalidated_cloned = xpc_conn_invalidated.clone();
+    let shutdown_token_cloned = shutdown_token.clone();
     let invalidation_handler = block2::RcBlock::new(move || {
         xpc_conn_invalidated_cloned.store(true, std::sync::atomic::Ordering::SeqCst);
```
pronebird: Since `shutdown_token_cloned.cancel();` will break the loop and cause XPCConnection deinitialization, you may as well then go ahead and remove `xpc_conn_invalidated` entirely. What do you think?

nym-vpn-core/CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [macOS] Authentication layer for windows, feature gated (https://github.com/nymtech/nym-vpn-client/pull/4802)
 - Activate authentication layer on all desktop platforms (https://github.com/nymtech/nym-vpn-client/pull/4856)
 
+### Fixed
+
+- [macOS] XPC client stall when daemon is not running (https://github.com/nymtech/nym-vpn-client/pull/4973)
```
pronebird, line 17: This should be under unreleased section which for some reason is not in here... Pilatus is v1.27

pronebird (changes requested):

nym-vpn-core/CHANGELOG.md
neacsu: Right, good catch. That means more of the entries above are also misrepresented, at least the `Added` macOS ones which are first included in this release... I'll fix them

nym-vpn-core/crates/nym-ipc/src/xpc/client.rs
neacsu: That sounds good, I've removed the atomic bool.

pronebird (approved):
:lgtm:

> @pronebird reviewed 6 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/4973#-:-Oo_9MoV-elPFGgODHIJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on trojanfoe).

PR #5060, opened 2026-04-08
Fix signing of `nym-vpnd.exe`.

trojanfoe:
During Split Tunnel signing work, I broke the signing of `nym-vpnd.exe`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5060)

### DIFF
--- a/nym-vpn-app/src-tauri/scripts/sign.sh
+++ b/nym-vpn-app/src-tauri/scripts/sign.sh
@@ -34,6 +34,16 @@ function sign {
 	fi
 }
 
-sign "$1"
+exe=$1
+
+# If we are signing the Tauri app, then also sign the daemon.
+if [[ "$exe" == *NymVPN.exe ]]; then
+	# shellcheck disable=SC2043
+	for additonal in "nym-vpnd.exe"; do
+		sign "$additonal"
+	done
+fi
+
+sign "$exe"
 
 exit 0


neacsu (changes requested):
> @neacsu reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5060)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on agentpietrucha, mrkiwi, pronebird, rokas-ambrazevicius, and trojanfoe).

___
*[`nym-vpn-app/src-tauri/scripts/sign.sh` line 40 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/5060#-OpgS_a56qbzV668q23r:-OpgS_a56qbzV668q23s:b-mspmg5) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/29bffe5a47a0c6176718bd69be866f4195bb79a1/nym-vpn-app/src-tauri/scripts/sign.sh#L40)):*
> ```Shell
> 
> # If we are signing the Tauri app, then also sign the daemon.
> if [[ "$exe" == *NymVPN.exe ]]; then
> ```

I don't think `NymVPN.exe` is the name that's used by Tauri app, I think it's something like `nym-setup.exe`, but that needs to be double checked

nym-vpn-app/src-tauri/scripts/sign.sh
```diff
@@ -34,6 +34,16 @@ function sign {
 	fi
 }
 
-sign "$1"
+exe=$1
+
+# If we are signing the Tauri app, then also sign the daemon.
+if [[ "$exe" == *NymVPN.exe ]]; then
```
agentpietrucha, line 40: This is correct. This is the name of the app @neacsu

agentpietrucha (approved):

trojanfoe:
@neacsu No that app name is correct.

nym-vpn-app/src-tauri/scripts/sign.sh
neacsu: This is the name of the app indeed, but previously on windows we were signing the setup binary (the installer), from what I've seen last time I looked at this. And that somehow magically signed this app binary too. We probably want to maintain that flow, since it's the installer that gets flagged at install time.

trojanfoe: OK so this script is used to sign many files; for example the installer and tauri plugins.  What that `if` statement does is also sign `nym-vpnd.exe` while it's signing the main Tauri app.  Previously with Split Tunneling I also used this to sign the driver files as well, but that's not required any longer.  You can see what's signed in [this run](https://github.com/nymtech/nym-vpn-client/actions/runs/24088803069/job/70269105091) (expand the "Build Tauri app" step).

Previously you had made it so it always signed `nym-vpnd.exe` along with any file that was passed in as an argument, but this was wasteful as it meant the `nym-vpnd.exe` file was signed about 6 times, which costs time and money.

neacsu (approved):

PR #5066, opened 2026-04-09
fix: reduce fresh-device login latency

rachyandco:
## Ticket

JIRA-VPN-1101

### Summary                                                                                     

Reduces fresh-device login latency on Android by addressing three client-side bottlenecks in the zk-nym ticketbook issuance flow.

### Changes:                                                  

- Raise account-ready timeout (LoginViewModel.kt): Increased ACCOUNT_READY_TIMEOUT_MS from 20s to 90s. On a fresh device, ticketbook issuance for all three ticket types can take longer than 20s, causing the UI to time out and fall through to a subscription check before  the account controller reaches ReadyToConnect.
- Skip redundant sync cycle on foreground resume (requesting_zknym_state.rs): When a RefreshAccountState event arrives while already in RequestingZkNyms, the state machine previously aborted and restarted the full sync cycle. It now checks credential storage first  — if tickets are already above the minimum threshold it transitions directly to ReadyState, making foreground-resume re-login near-instant. 
- Parallelize ticketbook crypto (request.rs): The unblind_and_aggregate function performs CPU-intensive EC cryptographic operations (key decoding, issue_verify per share, aggregate_wallets). Running this synchronously on tokio worker threads caused the three concurrently-spawned ticket-type tasks to serialize on mobile's limited thread pool. Moving the work into tokio::task::spawn_blocking lets all three run truly in parallel.

### Motivation

Fresh-device login was taking ~37s, making the onboarding experience feel broken. The timeout caused premature navigation, the redundant sync cycle added unnecessary round trips on resume, and the crypto serialization negated the existing JoinSet-based concurrency for  ticket types.                                             

### Result

Login time reduced from ~37s to ~29s on a fresh device. The remaining time is server-side zk-nym generation latency, which is not client-controllable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5066)

### DIFF
--- a/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/ui/screens/settings/login/LoginViewModel.kt
+++ b/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/ui/screens/settings/login/LoginViewModel.kt
@@ -32,7 +32,7 @@ class LoginViewModel @Inject constructor(private val settingsRepository: Setting
 
 	companion object {
 		private const val TAG = "ui-login-vm"
-		private const val ACCOUNT_READY_TIMEOUT_MS = 20_000L
+		private const val ACCOUNT_READY_TIMEOUT_MS = 90_000L
 	}
 
 	private val _uiState = MutableStateFlow(LoginUiState())
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/requesting_zknym_state.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/requesting_zknym_state.rs
@@ -359,7 +359,9 @@ impl RequestingZkNymsState {
                     .is_all_ticket_types_above_minimal_threshold()
                     .await
                 {
-                    info!("RefreshAccountState: tickets already sufficient, transitioning to ReadyState");
+                    info!(
+                        "RefreshAccountState: tickets already sufficient, transitioning to ReadyState"
+                    );
                     self.zk_nym_fetching_handle.abort();
                     return NextAccountControllerState::NewState(ReadyState::enter());
                 }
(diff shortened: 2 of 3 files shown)


nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/ui/screens/settings/login/LoginViewModel.kt
```diff
@@ -32,7 +32,7 @@ class LoginViewModel @Inject constructor(private val settingsRepository: Setting
 
 	companion object {
 		private const val TAG = "ui-login-vm"
-		private const val ACCOUNT_READY_TIMEOUT_MS = 20_000L
+		private const val ACCOUNT_READY_TIMEOUT_MS = 90_000L
```
neacsu, line 35: Not sure about the effects on this change, it's @mrkiwi call

nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/syncing_state/requesting_zknym_state.rs
```diff
@@ -348,12 +348,25 @@ impl RequestingZkNymsState {
...
-                } else {
+                if shared_state.firewall_active {
+                    return NextAccountControllerState::SameState(self);
+                }
+                // If tickets are already sufficient (written by the in-progress fetch),
+                // skip the full SyncingState round-trip and go directly to ReadyState.
+                // This makes foreground-resume refreshes instant once tickets are available.
+                if let Ok(true) = shared_state
+                    .credential_storage
+                    .is_all_ticket_types_above_minimal_threshold()
```
neacsu, line 359: The fact that we have the possibility of receiving AC commands in this state which might increase the time for exiting from state into `Ready` is indeed problematic, but I don't think it's happening at this moment. Technically this check could be made at every command, but in practice we should go very quickly into `handle_retrieved_zk_nym` which does the same check, so I think this just increases the code complexity without adding a clear benefit.

neacsu (changes requested):
On the core side, I think the initial ideas are good and should be followed up, more precisely:
1. Parallelism crypto intensive operations and using a blocking thread.
2. Making sure we don't delay the exiting of requesting zknym state because of some AC command that might take too long to handle

From some Linux CLI testing, I couldn't see a clear decrease in times, but that's mostly because of the network latency and server side computing, which make most of the delay.

PR #5069, opened 2026-04-09
(core) Skip redundant sync cycle on foreground resume

rachyandco:
## Ticket

JIRA-VPN-XXXX

## Description

When a RefreshAccountState event arrives while already in RequestingZkNyms, the state machine previously aborted and restarted the full sync cycle. It now checks credential storage first — if tickets are already above the minimum threshold it transitions directly to ReadyState, making foreground-resume re-login near-instant

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5069)


pronebird:
Unfortunately, this change is unacceptable. `spawn_blocking` is not how this should be done.

PR #5071, opened 2026-04-10
Add `inf2cat` option `/uselocaltime` option.

trojanfoe:
In order to avoid summer time issues between `stampinf` and `inf2cat`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5071)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
+++ b/nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
@@ -180,9 +180,6 @@ impl TryFrom<NymWellknownDiscoveryItemResponse> for Discovery {
 mod tests {
     use std::collections::HashMap;
 
-    // These tests need to run serially in order to avoid issues with the hyper crate.
-    use serial_test::serial;
-
     use time::{OffsetDateTime, format_description::well_known::Rfc3339};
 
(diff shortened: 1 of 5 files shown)


agentpietrucha (approved):

nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
```diff
@@ -180,6 +180,9 @@ impl TryFrom<NymWellknownDiscoveryItemResponse> for Discovery {
 mod tests {
     use std::collections::HashMap;
 
+    // These tests need to run serially in order to avoid issues with the hyper crate.
```
pronebird: Seems like tests still fail despite that change

pronebird (changes requested):

nym-vpn-core/crates/nym-vpn-network-config/src/discovery.rs
trojanfoe: Damn.  The issue seems to centre around not using `.non_shared()` when building the `reqwest` client.  I didn't want to make that change as it would have universal impact.

pronebird (approved):

PR #5161, opened 2026-04-25
fix(daemon): ship polkit policy via packaging, fall back to /etc on read-only /usr

ItsumoSeito:
## Ticket

N/A (community contribution)

## Description

**Bug:** on Fedora Silverblue and other immutable-`/usr` distros (MicroOS, NixOS, Bluefin),
`nym-vpnd` silently fails to register its polkit action because it writes
`com.nymvpn.vpnd.unix-access.policy` to `/usr/share/polkit-1/actions/` at runtime, which is
read-only (`EROFS`). Polkit never sees the action, every auth attempt is denied, and the modal
loops with no diagnostic output at the default log level.

**Commit 1 — ship the policy file via packaging**

The standard Linux convention: daemons ship polkit policies through the package manager, not at
runtime. Added `com.nymvpn.vpnd.unix-access.policy` as a packaging asset to the deb
(`Cargo.toml` assets), both AUR PKGBUILDs, and the raw install script. The file is also now the
single source of truth — the daemon embeds it via `include_str!` instead of an inline duplicate
string.

**Commit 2 — runtime fallback + better logging**

Defence-in-depth for raw binary installs (no package manager step). The daemon now tries
`POLICY_DIRS = ["/usr/share/polkit-1/actions", "/etc/polkit-1/actions"]` in order. Notable:
Fedora Silverblue does not ship `/etc/polkit-1/actions/` by default, so `create_dir_all` is
called before opening the file. Per-attempt `debug!` logging, `info!` on success, `error!` when
all candidates fail — all visible at the default log level in `/var/log/nym-vpnd/nym-vpnd.log`.
Auth-denied line in `mod.rs` promoted from `debug` to `warn` for the same reason. Four unit
tests added covering the fallback ordering.

Also fixes a pre-existing bug: `strum` was missing `features = ["derive"]` in the workspace
`Cargo.toml`, breaking `cargo test -p nym-ipc`.

**Commit 3 — client modal UX**

The wire protocol is a single byte (`Accepted`/`Denied`/`Closed`) with no room for an error
reason without a breaking change. As a best-effort: the auth-denied modal now shows a
troubleshooting hint pointing to `/var/log/nym-vpnd/nym-vpnd.log` and a docs link
(`LinuxAuthTroubleshootingUrl` which currently redirects to the issues, happy to update to a dedicated
article). Stray `console.log` removed.

**Reviewer notes**

- `cargo deny check` was already failing on `develop` (pre-existing advisories + unused `ts-rs`
  dep) — nothing introduced by this PR.
- `/etc/polkit-1/actions/` being scanned by polkit is common on most distros but not in upstream
  polkit's documented search path — flagging for reviewer confirmation.
- The runtime install path could be removed entirely in a follow-up once packaging has
  propagated; left in as defence-in-depth for now.
- Wire protocol redesign to pass a structured error reason to the client is intentionally out of
  scope.

## Checklist

- [x] Changelog

## Screenshots

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5161)

### DIFF
--- a/.pkg/linux/install
+++ b/.pkg/linux/install
@@ -88,6 +88,7 @@ vpnd_deb_ver_url=$vpnd_version
 vpnd_raw_url="https://github.com/nymtech/nym-vpn-client/releases/download/$vpnd_tag/nym-vpn-core-v${vpnd_version}_linux_${arch}.tar.gz"
 vpnd_deb_url="https://github.com/nymtech/nym-vpn-client/releases/download/$vpnd_tag/nym-vpnd_${vpnd_deb_ver_url}_${deb_arch}.deb"
 unit_url="https://raw.githubusercontent.com/nymtech/nym-vpn-client/fb526935/nym-vpn-core/crates/nym-vpnd/.pkg/aur/nym-vpnd.service"
+polkit_policy_url="https://raw.githubusercontent.com/nymtech/nym-vpn-client/$vpnd_tag/nym-vpn-core/crates/nym-ipc/.pkg/com.nymvpn.vpnd.unix-access.policy"
 ##################################################################
 
 # function called when an error occurs, will print the exit
@@ -178,6 +179,9 @@ core_archive="nym-vpn-core-v${vpnd_version}_linux_${arch}.tar.gz"
 vpnd_bin="nym-vpnd"
 socks5_proxy_bin="nym-socks5-proxy"
 vpnd_service="nym-vpnd.service"
+polkit_policy_file="com.nymvpn.vpnd.unix-access.policy"
+polkit_policy_dir="/usr/share/polkit-1/actions"
+polkit_policy_fallback_dir="/etc/polkit-1/actions"
 vpnd_deb="nym-vpnd_${vpnd_deb_ver}_${deb_arch}.deb"
 units_dir="/usr/lib/systemd/system"
 desktop_exec='env RUST_LOG=info,nym_vpn_app=debug NymVPN.AppImage -l %U'
@@ -350,6 +354,8 @@ download_vpnd_raw() {
   log "   ${BI_MGT}ok$RS"
   log "  ${B_GRN}Downloading$RS unit file"
   curl -fL -# "$unit_url" -o nym-vpnd.service
+  log "  ${B_GRN}Downloading$RS polkit policy"
+  curl -fL -# "$polkit_policy_url" -o "$polkit_policy_file"
   log "  ${B_GRN}Unarchiving$RS nym-vpnd"
   tar -xzf "$core_archive"
   mv "${core_archive%.tar.gz}/$vpnd_bin" $vpnd_bin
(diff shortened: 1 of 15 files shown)


trevarj:
~This is an issue on Guix System as well. Would be nice to have it merged!~

Actually, I was able to solve this on Guix System by simply registering the action in polkit with the provided .xml policy file. Works on the latest release v1.29.2 of core.

nym-vpn-core/crates/nym-ipc/src/authentication/linux.rs
```diff
@@ -98,6 +96,95 @@ impl AuthorizationChecker for AuthProxy<'_> {
...
+             this typically means /usr/share is read-only \
+             AND /etc/polkit-1/actions/ is not writable.",
+        );
+        AuthenticationError::CreateActionPolicy(err)
+    };
+
+    Err(final_err)
+}
+
+enum PolicyWriteError {
```
neacsu, line 162: I think we can reuse the `AuthenticationError`, no need for a secondary error type.

nym-vpn-core/crates/nym-ipc/src/authentication/linux.rs
```diff
@@ -98,6 +96,95 @@ impl AuthorizationChecker for AuthProxy<'_> {
...
+        tracing::error!(
+            error = %err,
+            candidates = ?dirs,
+            "failed to install polkit policy in any candidate directory; \
+             see the daemon log file at /var/log/nym-vpnd/nym-vpnd.log \
+             and the troubleshooting docs",
+        );
+        AuthenticationError::WriteActionPolicy(err)
+    } else {
+        let err = create_err.expect("at least one path attempted");
```
neacsu, line 148: This `expect` would trigger on empty `dirs`. [I've pushed some refactoring of this function](https://github.com/nymtech/nym-vpn-client/tree/ItsumoSeito/fix-polkit-policy-rofs) so that is not possible anymore, using two arguments for the function, a primary and fallback options. Please take a look.

nym-vpn-core/CHANGELOG.md
neacsu, line 1: We usually do just one entry per PR in core changelog, specifying the PR link. I've made the modifications in [the pushed branch.](e4f6f7e7b240150866e1a0a582f7d2e0ea577c45)

.pkg/linux/install
```diff
@@ -178,6 +179,9 @@ core_archive="nym-vpn-core-v${vpnd_version}_linux_${arch}.tar.gz"
 vpnd_bin="nym-vpnd"
 socks5_proxy_bin="nym-socks5-proxy"
 vpnd_service="nym-vpnd.service"
+polkit_policy_file="com.nymvpn.vpnd.unix-access.policy"
+polkit_policy_dir="/usr/share/polkit-1/actions"
+polkit_policy_fallback_dir="/etc/polkit-1/actions"
```
neacsu, line 184: I see this is only used on the uninstall function, in the package install script. There is no fallback mechanism in the package installation

nym-vpn-app/src/i18n/en/system-authentication.json
```diff
@@ -6,6 +6,8 @@
   "modal": {
     "title": "Authentication required",
     "description": "You need to authenticate with your password to secure your connection.",
+    "troubleshooting": "If clicking Authenticate has no effect, check the GitHub issues for troubleshooting.",
```
neacsu, line 9: I'm not sure if we want to include UI changes with this too, but I defer to @agentpietrucha on this one

neacsu (changes requested):
Thank you for your active role in developing the NymVPN application.
This is a useful PR, I think there are things here we should be merging.
I've left some comments and pushed some commits to https://github.com/nymtech/nym-vpn-client/tree/ItsumoSeito/fix-polkit-policy-rofs .

nym-vpn-app/src/i18n/en/system-authentication.json
agentpietrucha: I am good with additional explanation info

neacsu:
Closing in favor of https://github.com/nymtech/nym-vpn-client/pull/5528

PR #5112, opened 2026-04-17
Quick connect algorithm

neacsu:
## Ticket

JIRA-NYM-989

## Description

Implement two more options for quickly connecting when there's little information from the user (no entry gateway selected or neither entry nor exit gateway selected).

## Checklist:

- [x] Changelog

## Screenshots (optional, if UI related)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5112)

PR #5185, opened 2026-04-29
feat(desktop): expose Quick connect (#5112) in Home UI

vvmmaann:
Brings parity for #5112 (Quick connect algorithm) to the Linux/Windows Tauri desktop app. PR #5112 added GatewaySelectionAlgorithm::Auto to core and `nym-vpnc` CLI but did not expose it in the desktop UI; this PR fills that gap.

## Commits

1. **fix(lib-types)**: 1-line fix for a missing conditional import in `nym-vpn-lib-types/src/gateway_selection_algorithm.rs`. The file uses `derive(TS)` under `cfg_attr(feature = "typescript-bindings", ...)` without the corresponding `use ts_rs::TS;` (which `service.rs` does have).
2. **chore(desktop)**: `src-tauri/Cargo.toml` was on `ts-rs = "11.0"` while the core workspace uses 12.0.1, causing two `ts_rs::TS` trait impls when pulling lib-types bindings. Bumping aligns them. Existing TS-derived types in src-tauri compile without source changes.
3. **feat(desktop)**: the actual feature.

## Backend (src-tauri)

- New Tauri command `set_gateway_selection_algorithm`, mirrors CLI pattern in `nym-vpnc/src/commands/tunnel.rs`
- Companion `VpndClient::set_gateway_selection_algorithm` method
- `gateway_selection_algorithm_config: GatewaySelectionAlgorithmConfig` added to `VpndConfig` so daemon state is observable in the UI
- Command registered in `main.rs` handler list

## Frontend (src)

- Quick connect toggle on Home, visible only in Fast (WireGuard) mode
- In Anonymous (mixnet) mode the toggle is hidden and the algorithm auto-resets to Explicit with a toast
- HopSelect cards show "Auto" placeholder when toggle is on and tunnel is down; once connecting/connected, the actually-selected gateway info is shown via the existing `getGatewayInfo` path so the user can see which gateways the daemon picked
- State hydrates from daemon config at startup and on `update-tunnel-config` so it stays in sync if the user changes algorithm via CLI/config
- Local TS enum replaced by generated `GatewaySelectionAlgorithm` from lib-types (3 variants: `explicit`, `autoEntryExplicitExit`, `auto`)
- Fallback to `explicit` if config field is absent (older daemons)
- Generated `bindings.ts` checked in alongside `tauri.ts`
- `tsgen` script now formats both files
- English translations added; other locales TODO

## Open questions

1. Quick connect requires `Auto` which is tied to WireGuard 2-hop. Currently toggle is hidden in Anonymous mode and state auto-resets. Alternatives: (a) disabled with tooltip, (b) auto-switch mode on enable. Picked (a) - happy to switch.
2. Toggle placement is between mode selector and Entry card. Settings could also work - preferred?
3. `enable_geo_location` is left at backend default (true). Worth surfacing in v1, or follow-up?

## Tested

- All checks pass: `cargo check` on src-tauri, `npm run tscheck`, `lint`, `fmt:check`, `build`
- E2E: ran `npm run dev:app` against installed nightly 1.30.0-beta daemon (commit `de7d981`). Toggling Quick connect on switches the daemon’s `gateway_selection_algorithm_config.gateway_selection_algorithm` from `Explicit` to `Auto` (verified in daemon log). Subsequent Connect picked Auto-selected entry/exit gateways and reached **Connected** status:

Quick connect Connected with Auto-selected gateways shown in cards

The screenshot also illustrates the HopSelect behavior described above: instead of a flat "Auto" placeholder, the cards show the gateways the daemon actually picked (entry: t-lab in Islamabad PK; exit: NodesHub in Delhi IN), so users always know where their traffic is going. Tester is located in Kazakhstan, so PK and IN are reasonable Auto picks for that region.

## Future work proposals

While the Auto algorithm worked correctly in our test, two directions could improve user-facing resilience and operator-side network balance. Happy to explore in separate PRs:

### 1. Top-N randomization within the best tier

Auto in #5112 is deterministic: among gateways passing the same-jurisdiction filter, pick the closest by Haversine distance. If the user hits a transient issue with that one gateway (firewall, ISP route, brief outage, gateway under unrelated load), Connect deterministically retries the same gateway and keeps timing out. We saw this in early local testing before the test setup was fully proper.

Proposal: pick uniformly at random from the top-N closest non-jurisdiction gateways (e.g. N=5 within a small distance window). Failed attempts then get a different candidate next try. Tiny code change, no new data needed.

### 2. Richer scoring formula

Current Auto: filter (same jurisdiction) + sort (closest by Haversine).

Additional factors we have been experimenting with in a separate gateway-scoring project (used today by node operators for deployment recommendations):
- **saturation penalty** (penalize gateways in over-concentrated countries / ASNs to spread the network)
- **exit quality signals** (SMTP egress clean, IPv6 default, residential vs hosting tier)
- **jurisdiction tier** (avoid "extreme operator-risk" countries for entry/exit even if geographically optimal)
- **per-client recent-latency cache** (fast-path pre-selection based on actually-measured RTTs to candidates)
- **entry-near-user / exit-far-from-entry tradeoff** as a tunable rather than fixed

Each factor is independent and opt-in. Shipping any of these in core would require either new fields in `nym-api` (gateway metadata) or an auxiliary dataset shipped with the daemon. Happy to open a scoping issue if the team wants to evaluate.

## Note

This PR intentionally does not change the Auto formula in core. It only exposes the existing #5112 capability in the desktop UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5185)


neacsu:
Hello,

Thank you for your PR and taking an active part in Nym's mission.

Currently we are actively working on integrating the UI with the core PR [#5112](https://github.com/nymtech/nym-vpn-client/pull/5112). Because of different developers working on this feature, we are first implementing it on the core side and afterwards integrate it into UI for each platform. nym-vpnc is an exception, as it's mostly used for development purposes.

The implemented solution will make use of all three possible types (explicit, autoEntryExplicitExit, auto) in a robust manner. We would love to receive your feedback on it, once it gets pushed or after it gets released. We are currently aiming it for the upcoming Schilthorn release.

As such, in order to not cause merge conflicts with the current work we're doing, I will close this PR.

Thank you for promptly taking interest in the gap issue between CLI and GUI and for the future work proposals. Stay tuned for the incoming work.

Bogdan

PR #5210, opened 2026-05-02
diagnostic-tool: remove boringtun to unblock fdroid build

mmsinclair:
Note: the wireguard test should be added back once time allows to troubleshoot how the fdroid build fails with `boringtun` included

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5210)


pronebird:
One of issues we found is that our and fdroid builds have very different byte size. So I'll investigate that first before going surgical. Thanks.

pronebird:
Thanks for the effort. However boringtun seems to be perfectly fine. The issue was due to shipping binaries built with wrong build profile.

PR #5271, opened 2026-05-12
[Android] F-Droid anonymous account creation + web checkout flow

rachyandco:
Adds an end-to-end anonymous-account flow for F-Droid builds (no Google
  Play Billing). The user can create an account whose mnemonic stays
  local-only, and registration with the backend is deferred until the
  user signals checkout intent.

  Rust core:
  - nym-vpn-api-client: new register_anonymous_account that POSTs to /api/public/v1/account (no purchase token, deserializes NymVpnAccountResponse since the generic endpoint returns the account object, not an account_token)
  - nym-vpn-account-controller: new RegisterAnonymousAccount command + handler, dispatched in all relevant state machine states
  - nym-vpn-lib: new VpnServiceCommand::RegisterAnonymousAccount + handler
  - nym-vpn-lib-uniffi: NymVpnServiceCommandSender.register_anonymous_account exposed for mobile bindings

  Android:
  - VpnServiceApi(Impl) + BackendManager(Impl): registerAnonymousAccount() surface
  - AppViewModel.fetchAutologin: on F-Droid, call registerAnonymousAccount before getAutologinDeeplink so the website can sign the user in; failures are swallowed to handle the already-registered case
  - GeneratingViewModel: keeps createAccount() purely local — no backend call after mnemonic generation
  - CreateAccountScreen: F-Droid branch removed; always navigates to Generating, branching on isBillingAvailable happens downstream
  - GeneratingScreen: post-create routes Google flavor to SelectPlan, F-Droid to Main
  - ConnectionButton + MainScreen: when user is logged in but has no active subscription, render "Get started" instead of "Connect" without requiring a failed-connect detour
  - MainActivity: account/response deeplink branch shows the Generating loader while not-logged-in, mirroring the privy login pattern

## Ticket

JIRA-NYM-3746

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5271)

## Summary by CodeRabbit

* **New Features**
  * Added support for anonymous account registration in F-Droid installations.
  * Enhanced account initialization with billing availability checks.

* **Improvements**
  * Streamlined account creation and signup flows.
  * Updated navigation logic based on subscription status.
  * Improved subscription state handling in connection controls.


rokas-ambrazevicius (approved):

nym-vpn-core/crates/nym-vpn-lib/src/service/vpn_service.rs
```diff
@@ -1047,6 +1052,10 @@ impl NymVpnService {
             VpnServiceCommand::RegisterAccount(tx, request) => {
                 let _ = tx.send(self.handle_register_account(request).await);
             }
+            #[cfg(any(target_os = "android", target_os = "ios"))]
+            VpnServiceCommand::RegisterAnonymousAccount(tx, ()) => {
```
pronebird, line 1056: We already have "register account" call. We can make payment token optional.

pronebird (changes requested):

nym-vpn-core/crates/nym-vpn-lib/src/service/vpn_service.rs
rachyandco: here the point is to register an account that was created outside of any apps we provide, using a valid bip39 mnemonic. if we can use the same endpoint, sure.

pronebird: Sure but the endpoint that's used in this PR is meant to be used by iOS only. I think we should bring this up with Fouad to make sure that the right calls are being made to the backend systems.

PR #5300, opened 2026-05-13
HTTP reachability in diagnostic tool testing

jmwample:
## Ticket

JIRA-NYM-1328

## Description

Expand the functionality of the diagnostic tools HTTP o a short test where each of the avaiable configurations for HTTP requests is used to make a simple request.

For returned results if a stealth API is defined then the Policy was set such that it was used for the trial.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5300)

## Summary by CodeRabbit

* **New Features**
  * HTTP diagnostics now include per-endpoint results (status and endpoint URL) instead of only aggregated summaries.
  * Diagnostic checks have been expanded to probe both Nym APIs and VPN APIs, returning detailed per-endpoint outcomes for each tested URL.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym-vpn-client/pull/5300)

### DIFF
--- a/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
+++ b/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
@@ -49,7 +49,7 @@ impl HttpDiagnostic {
     }
 
     async fn test_nym_apis(network: &Network, parallel: bool) -> Result<()> {
-        tracing::info!("nym apis");
+        tracing::info!("Running Nym api diagnostics");
 
         let api_clients = build_nym_api_clients(network).await?;
 
@@ -73,7 +73,7 @@ impl HttpDiagnostic {
     }
 
     async fn test_vpn_apis(network: &Network, parallel: bool) -> Result<()> {
-        tracing::info!("vpn apis");
+        tracing::info!("Running VPN api diagnostics");
         let api_clients = build_vpn_api_clients(network).await?;
 
         if parallel {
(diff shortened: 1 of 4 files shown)


simonwicky (approved):

nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
```diff
@@ -34,10 +44,184 @@ impl HttpDiagnostic {
...
+
+        clients.push(plain_client);
+
+        if let Some(fronts) = url.fronts() {
+            for front in fronts {
+                let fronted_url = Url::new(url.inner_url().clone(), Some(vec![front.clone()]))
+                    .map_err(|_e| Error::MissingApiUrl)?;
+
+                let fronted_client = ClientBuilder::new(fronted_url)
+                    .map_err(|_| Error::MissingApiUrl)?
```
simonwicky, line 182: nit : how about using a custom user-agent?

pronebird (approved):

nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
pronebird: it would be good to reuse the same user-agent we use in diagnostic tool

jmwample: The previous http tests were just using the default `new_user_agent!()`. I made it so that all of the http trials use that now. If there is a reason we can make a custom user-agent, but for now there isn't one and they all match.

PR #5365, opened 2026-05-19
Reduce number of Account Controller retries

jmwample:
## Ticket

JIRA-NYM-XXXX

## Description

The account controller is retrying way to many times way to quickly. 

~~This PR reduces the number of retries down to 2  -- which is actually 9 HTTP request attempts as the  HTTP client is configured with 3 retries per attempt, and inital + 2 sync retry attempts.~~

~~Also the wait between attempts now starts at a 2s delay.~~

This PR removes the retry attempts from the account controller state machine. If the account controller encounters a failed step it transitions to the error state directly.

Also adds a check for stored sufficient zk_nyms before transitioning to the RequestingZkNyms state so we skip an API access if it is not necessary. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5365)

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed repeated retry/backoff during account sync so failures progress faster to pending, error, or ready states.
  * Improved handling of zk-nym and sync failures to reduce stuck states and speed recovery.
  * Reduced verbosity of account debug output for cleaner logs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym-vpn-client/pull/5365?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/controller.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/controller.rs
@@ -117,7 +117,7 @@ where
         let (current_state_handler, initial_state) = if is_offline {
             OfflineState::enter()
         } else {
-            SyncingState::enter(&shared_state, 0)
+            SyncingState::enter(&shared_state)
         };
 
         let public_initial_state = AccountControllerState::from(initial_state);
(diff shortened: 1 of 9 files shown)


pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5365)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on jmwample, neacsu, simonwicky, and trojanfoe).

___
*[a discussion](https://reviewable.io/reviews/nymtech/nym-vpn-client/5365#-OtDxpFG6tiPmTMWj4aB:-OtDxpFG6tiPmTMWj4aC:b-sejmgt) (no related file):*
As noted on Zulip. I am not entirely sure this will yield better results. It would be better to persist account controller state and reduce number of requests for account summary.

PR #5384, opened 2026-05-21
feat(tray): toggle window on left-click, menu on right-click

VictorVow:
## Summary

Improved tray UX: **left-click toggles the main window**, **right-click opens the context menu**. Works on Windows, macOS, and Linux (KDE Plasma / GNOME w/ AppIndicator extension / any SNI host).

## Why

Today, left-clicking the tray icon only *shows* the window — never hides it. The Show/Hide menu item is the only way to hide it again. This is inconsistent with every other tray-using app, and on Linux the tray was effectively unusable for toggling because libayatana-appindicator turns every click into a menu open.

## Changes

**Windows/macOS** — small fix on the existing Tauri tray:
- `TrayIconBuilder::show_menu_on_left_click(false)` so left-click doesn't auto-open the menu.
- Left-click handler now calls `show_window(app, true)` (toggling) instead of `false` (show-only).

**Linux** — replace the Tauri ``tray-icon`` backend with [``ksni``](https://crates.io/crates/ksni):
- Tauri's GTK backend (``tray-icon-0.23.1/src/platform_impl/gtk/mod.rs``) only calls ``indicator.set_menu()``. libayatana-appindicator never emits a click event, so there is no way to bind toggle-window to a left click while keeping the menu accessible.
- ``ksni`` speaks the StatusNotifierItem protocol directly and exposes ``activate`` / ``secondary_activate`` callbacks, which we wire to ``show_window(..., true)``. The menu is rebuilt from state on each ``Handle::update``.
- Public ``TrayManager`` API is unchanged — non-Linux code paths are untouched aside from the two-line fix above.

## Implementation notes

- ``ksni`` built with ``default-features = false, features = ["async-io"]``. The default ``tokio`` feature would enable ``zbus/tokio`` transitively, and ``tauri-plugin-single-instance`` uses ``zbus::blocking`` during plugin init. The tokio-flavored blocking shim calls ``Runtime::block_on()`` inside Tauri's own runtime, which panics.
- ``ksni::TrayMethods::spawn()`` is driven on a dedicated init thread via ``futures::executor::block_on``. Tauri's setup hook is sync but runs inside Tauri's tokio runtime, so we cannot ``block_on`` there; with the async-io feature ``ksni`` owns its own executor for the service task, and the returned ``Handle`` is runtime-agnostic.
- ``quit_app`` uses ``tauri::async_runtime::spawn`` instead of ``tokio::spawn`` because on Linux it can be invoked from ``ksni``'s async-io executor thread, which has no tokio context.

## Test plan

- [x] Builds cleanly (``cargo build`` and ``npm run dev:app``).
- [x] Linux (KDE Plasma 6, Wayland): left-click toggles, right-click opens menu, Show/Hide menu item works, Quit menu item disconnects + exits cleanly. Verified via ``tray activate`` / ``quit menu clicked`` trace logs.
- [x] Windows: left-click toggle + right-click menu. *(I do not have a Windows box; please verify.)*
- [x] macOS: same. *(Same, please verify.)*
- [x] GNOME with AppIndicator extension: ``ksni`` advertises as ``KStatusNotifierItem`` which the extension also implements. Should "just work" but worth a sanity check.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5384)

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored left-click behavior for the Linux tray icon.
  * App reliably toggles/focuses the main window from tray actions.
  * Safer shutdown: quitting from the tray disconnects the VPN in relevant states.

* **Chores**
  * Unified tray handling across platforms for consistent icons and tooltips.
  * Reduced redundant icon update frequency for improved performance.

### DIFF
--- a/nym-vpn-app/src-tauri/Cargo.lock
+++ b/nym-vpn-app/src-tauri/Cargo.lock
@@ -3179,6 +3179,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "ksni"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7ca513d0be42df5edb485af9f44a12b2cb85af773d91c27dc796d1c58b78edc"
+dependencies = [
+ "async-executor",
+ "async-io",
+ "async-lock",
+ "futures-channel",
+ "futures-lite",
+ "futures-util",
+ "once_cell",
+ "pastey",
+ "serde",
+ "zbus",
+]
+
+[[package]]
 name = "language-tags"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3744,6 +3762,7 @@ dependencies = [
  "futures",
  "image",
  "itertools",
+ "ksni",
  "notify",
  "nym-platform-metadata",
  "nym-vpn-lib-types",
@@ -4307,6 +4326,12 @@ dependencies = [
  "smallvec",
  "windows-link 0.2.1",
 ]
+
+[[package]]
+name = "pastey"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2ee67f1008b1ba2321834326597b8e186293b049a023cdef258527550b9935b4"
 
 [[package]]
 name = "pathdiff"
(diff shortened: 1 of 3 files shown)


pronebird:
Thanks for submitting this PR. @agentpietrucha will take a look at it.

agentpietrucha (changes requested):
@VictorVow great catch with the tray icon! 🙌🏻 Thanks for the PR.

I see CI is failing because of `cargo fmt --check`. Could you please run `cargo fmt --all` and push your changes?

VictorVow:
@agentpietrucha done, thanks for the review :)

agentpietrucha (approved):
@VictorVow Great job, it's going in! 🎉

PR #5429, opened 2026-05-27
Blacklist specific entry

neacsu:
## Ticket

JIRA-NYM-1403

## Description

Split filtering for entry into base and optional, as performance filtering is not mandatory for explicit gateway selection, but blacklisting should be mandatory, to avoid retrying indefinitely a bad gateway.

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5429)

## Summary by CodeRabbit

* **New Features**
  * Gateway selection now detects and avoids non-functioning gateways, even when explicitly selected.
  * Added optional gateway filtering fallback mechanism for improved gateway selection.
  * Improved error reporting when a directly selected gateway exists but doesn't satisfy filters.

* **Bug Fixes**
  * Fixed missing awaits in gateway blacklist handling so blacklist updates are applied reliably during tunnel transitions.

* **Tests**
  * Added/updated tests covering direct selection with blacklisting and fallback selection behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym-vpn-client/pull/5429?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/mod.rs
```diff
@@ -41,6 +41,7 @@ type SelectedGatewaysStream =
 #[derive(Clone)]
 pub struct GatewayProvider<C: GatewayCache> {
     gateway_cache: C,
+    latest_tunnel_settings: TunnelSettings,
     tunnel_settings_tx: mpsc::Sender<SelectAndSend>,
```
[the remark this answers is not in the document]
tommyv1987: I guess what the rabbit says here - we should refresh latest_tunnel_settings on every set_tunnel_settings call, or make it shared/updatable so blacklist-triggered reselection uses current settings.

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/selector.rs
```diff
@@ -205,8 +205,21 @@ fn select_entry(
         }
     };
 
-    find_best_entry_gateway(&entry_gateways, entry_ordering_criteria, &entry_filters)
-        .map_err(GatewayProviderError::EntryGatewayUnavailable)
+    let gateway = find_best_entry_gateway(&entry_gateways, entry_ordering_criteria, &entry_filters)
```
tommyv1987: With this we should restrict ExplicitEntryGatewayNotWorking to explicit entry selection only.

Maybe we should add a unit test around this logic too?

explicit entry + blacklisted gateway => new error etc

neacsu: I've moved the erroring directly into the explicit entry selection, so that there's no error confusion. That actually makes it possible to reuse existing error types, so it's cleaner. Also added a unit test. Thanks for reviewing

trojanfoe (approved):

PR #5504, opened 2026-06-05
fix(core): gate ConnectingState behind account-readiness preflight

vvmmaann:
## Problem

When the user clicks **Connect**, `DisconnectedState` transitions directly into `ConnectingState::enter()`, which immediately applies a restrictive `FirewallPolicy::Connecting` (killswitch up). Only **after** the firewall is up does `TunnelMonitor` begin checking account readiness.

If the account controller is in `Syncing` / `RequestingZkNyms` / `PendingSubscription`, or has reached a terminal non-connectable state (`Offline`, `LoggedOut`, `Error(_)` including `InactiveSubscription`), the user is left with the killswitch on and an indefinitely-pending tunnel handshake. Empirically this surfaces as the UI being stuck at *"Setting up your account (2/6)"* while the user has lost normal internet access.

Reproduced live against `nym-vpn-lib 1.30.0`: pre-Connect, the account controller already had `InactiveSubscription` responses for an inactive account. On `Connect`, the daemon raised the firewall (`Allowing endpoints: none` for ~1s, then API-only endpoints) and entered `awaiting account readiness, try #0`. The user had to manually `Disconnect` to recover normal internet.

The same direct `ConnectingState::enter()` bypass exists at 4 additional callsites:

- `ErrorState::Connect`
- `OfflineState` auto-reconnect when connectivity returns
- `ConnectedState` tunnel-down recovery (`Reconnect`)
- `DisconnectingState` `PrivateActionAfterDisconnect::Reconnect`

## Fix

Introduce an internal `AccountPreflightState` that gates every transition into `ConnectingState`:

- `ReadyToConnect | Decentralised | UpgradeMode` -> enter `ConnectingState` immediately; behavior unchanged.
- `Syncing | RequestingZkNyms | PendingSubscription` -> reset to unrestricted networking and wait for account state changes. UI still reports `Connecting::AwaitingAccountReadiness` while waiting, but the firewall is **not** raised; the user keeps normal internet.
- `Offline | LoggedOut | Error(_)` -> fall back to `DisconnectedState`. The existing account-state UI already surfaces the specific account message (no plan / logged out / etc.) and routes the call-to-action to pricing or login, so no new public state or binding is required here.

Covered callsites (all switched from `ConnectingState::enter` to `AccountPreflightState::enter`):

- `DisconnectedState` user `Connect` command
- `ErrorState` user `Connect` command
- `OfflineState` auto-reconnect when connectivity returns
- `ConnectedState` tunnel-down recovery reconnect
- `DisconnectingState` `Reconnect` action-after-disconnect

Light refactor in `DisconnectedState`: extract `reset_to_unrestricted_networking()` and `apply_tunnel_settings()` so the preflight state can reuse them without duplicating logic.

The retry/wait logic inside `TunnelMonitor::await_account_readiness_with_retry` is left as-is; it remains useful as a fallback for races during the actual connect and for IAP propagation cases.

## Race-condition handling

`AccountPreflightState::enter` reclassifies the account state **after** subscribing to the `watch::Receiver` to close the window between the initial check and the subscription. Inside the state, `Disconnect` returns to `DisconnectedState`, `Block(reason)` defers to `ErrorState`, connectivity loss defers to `OfflineState`, and channel-drop falls back to `DisconnectedState`.

## Known tradeoff

For the `ConnectedState` tunnel-down -> reconnect path, if the account state has unexpectedly become non-Ready during the session (rare race), the preflight wait briefly leaves networking unrestricted instead of holding the killswitch as `ConnectingState::enter()` would have. This is a minor relaxation of the always-on killswitch guarantee during reconnect for users who explicitly rely on it. We can address this in a follow-up either by keeping the firewall up specifically for the `ConnectedState` reconnect path, or by adding an opt-in "strict killswitch during reconnect" setting.

## UI note

`Connecting::AwaitingAccountReadiness` is reused for the preflight wait, so users may briefly see *"Setting up your account (2/6)"* even when no firewall is applied. A follow-up that introduces an explicit account-preflight tunnel state with dedicated UI copy ("Checking subscription...") would be a small bindings-level change but is intentionally out of scope here.

## Test plan

- [x] `cargo check -p nym-vpn-lib` (clean)
- [x] `cargo fmt` (clean; pre-existing `imports_granularity` nightly-only warning is unrelated)
- [ ] Manual repro: inactive-subscription account, click Connect -> expect immediate fallback to disconnected/no-plan UI with internet intact
- [ ] Manual repro: active-plan account during `Syncing` -> expect UI to show "Setting up your account" briefly without internet drop, then transition to ConnectingState when account becomes Ready
- [ ] Manual repro: reconnect after network drop on an active-plan connection -> expect no behavior change

## Files changed

- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/account_preflight_state.rs` *(new)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/mod.rs` *(register new module)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/disconnected_state.rs` *(extract helpers, route Connect through preflight)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/error_state.rs` *(route Connect through preflight)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/offline_state.rs` *(route auto-reconnect through preflight)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs` *(route tunnel-down reconnect through preflight)*
- `nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/disconnecting_state.rs` *(route Reconnect action through preflight)*

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5504)

## Summary by CodeRabbit

* **New Features**
  * Added account preflight validation state to the tunnel connection flow, integrated across all state transitions to ensure account readiness before establishing connections.

* **Refactor**
  * Refactored networking reset and tunnel settings application logic into dedicated helper methods for improved code organization.

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -1,0 +1,57 @@
+# fix/disconnected-state-account-preflight
+
+## Summary
+
+This branch adds an internal account preflight state before entering `ConnectingState`.
+
+The key invariant is: the state machine must not call `ConnectingState::enter()` until the
+account controller is ready to connect (`ReadyToConnect`, `Decentralised`, or `UpgradeMode`).
+`ConnectingState::enter()` raises the VPN API firewall immediately, so inactive/no-plan
+accounts must be rejected or waited on before that point.
+
+Covered callsites:
+
+- `DisconnectedState` user connect command
+- `ErrorState` user connect command
+- `OfflineState` auto-reconnect when connectivity returns
+- reconnect paths through `ConnectedState` tunnel-down recovery and `DisconnectingState`
+
+The preflight state exposes `Connecting/AwaitingAccountReadiness` to the UI while it waits,
+but it first resets to unrestricted networking and does not apply the tunnel firewall. If
+the account becomes ready, it transitions to the normal `ConnectingState`. If the account
+reaches a terminal non-connectable state (`LoggedOut`, `Offline`, or `Error(_)`), it returns
+to `DisconnectedState`; the existing account-state UI then shows the specific account
+message and, for `no-subscription`, routes the main button to pricing.
+
+## Tests run
+
+```text
+cargo fmt --manifest-path nym-vpn-core\Cargo.toml --package nym-vpn-lib
+cargo check --manifest-path nym-vpn-core\Cargo.toml -p nym-vpn-lib
+```
+
+Result: `cargo check` passed.
+
+`cargo fmt` prints the existing rustfmt warning:
+
+```text
+Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
+```
+
+## Manual repro still needed
+
+Use the machine/account state from the original report:
+
+1. Start with an account that has no active subscription.
+2. Click Connect while the account controller is still `Syncing`/settling.
+3. Expected:
+   - UI may show "Setting up your account (2/6)" briefly.
+   - Internet must remain available; no "Allowing endpoints: none" blackout.
+   - Once account state resolves to no subscription, app returns to disconnected/no-plan UI.
+4. Also test reconnect after network offline/online while account is not ready.
+
+## Open questions
+
+- If product wants no visible "2/6" state for no-plan accounts, add a public non-firewall
+  account-preflight state and UI copy. I avoided that broader binding/UI change here.
+- GitHub issue/PR creation is pending repository access from this environment.
--- a/NOTES.md
+++ b/NOTES.md
@@ -26,8 +26,8 @@ message and, for `no-subscription`, routes the main button to pricing.
 ## Tests run
 
 ```text
-cargo fmt --manifest-path nym-vpn-core\Cargo.toml --package nym-vpn-lib
-cargo check --manifest-path nym-vpn-core\Cargo.toml -p nym-vpn-lib
+cargo fmt --manifest-path nym-vpn-core/Cargo.toml --package nym-vpn-lib
+cargo check --manifest-path nym-vpn-core/Cargo.toml -p nym-vpn-lib
 ```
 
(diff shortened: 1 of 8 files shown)


vvmmaann:
Thanks @coderabbitai. Addressed in `b428c22`:

- **Stale gateways on settings change during preflight (Major).** Real bug, fixed. `DisconnectedState::apply_tunnel_settings` now returns `Option<TunnelSettingsDiff>` instead of `bool`, and `AccountPreflightState` drops its cached `selected_gateways` when `entry_point`/`exit_point`/`quic` changed, mirroring `OfflineState`. `DisconnectedState`'s own `SetTunnelSettings` handler holds no cached gateways, so it intentionally ignores the returned diff.
- **NOTES.md backslash manifest paths (Minor).** Switched to forward slashes for POSIX-shell portability.

On the docstring-coverage pre-merge warning: added a doc comment on `apply_tunnel_settings` explaining the diff-return contract for caching callers.

pronebird:
Hi,

Thanks for that PR. However we're not looking to expand our tunnel state machine to more states. Account controller should be pre-fetching zknyms and account data in the background while disconnected. So when connecting it's expected for firewall to be engaged preventing leaks. We plan on introducing a few shortcuts to make the pre-flight checks much faster which hopefully should resolve those issues related to being stuck in the 2/6 state for too long.

PR #5505, opened 2026-06-05
fix(app): confirm before mode switch interrupts the VPN tunnel

vvmmaann:
## Problem

In `ModeToggle.tsx`, clicking the **Fast** or **Mixnet** pill while a VPN tunnel is active (or mid-connect) immediately fires `invoke('set_vpn_mode', ...)` and `invoke('set_gateway_selection_algorithm', ...)`. The daemon then disconnects the running tunnel and re-establishes it with the new mode. The user is given no chance to back out.

Reproduced live against `nym-vpn-app 1.30.0` while connected through `wg` (Fast) mode: a single tap on the **Mixnet** pill triggers an immediate `Disconnecting to reconnect` in the daemon log, drops the WireGuard tunnel, and starts a Mixnet handshake. Especially painful for users on slow/constrained networks where any reconnect is expensive (visible in logs as multiple WireGuard retry attempts before falling back to the bridge / port 4443).

## Fix

Pure UI change inside `ModeToggle.tsx`. Before applying a mode change, check the current tunnel state:

- If `state` is in `INTERRUPTIVE_STATES = {connected, connecting, offline-auto-reconnect}`, open the existing `ConfirmationDialog` with a mode-aware title and description.
- Otherwise (disconnected / disconnecting / error / offline / unknown), apply the change immediately - matching the previous behavior.

The dialog reuses the existing `ConfirmationDialog` component (Headless UI Dialog under the hood, with focus trapping and backdrop-click dismissal). `confirmInFlight` guards against double-clicks. Cancel keeps the previous mode and leaves the tunnel untouched; Confirm runs the existing `applyAlgorithm` -> `applyVpnMode` flow with the same rollback semantics as before.

i18n strings are added under `home.mode-toggle.switch-confirm.*` with separate descriptions for `connected` (will disconnect your VPN) and `connecting` (will cancel the current connection).

## Test plan

- [x] `cargo check` (workspace, clean)
- [x] No new TypeScript surface for the daemon - frontend-only change
- [ ] Connected in Fast mode, tap Mixnet -> confirmation dialog appears with "Switch to Mixnet mode?"; Cancel keeps tunnel + mode; Confirm disconnects and reconnects on Mixnet
- [ ] Mid-connect, tap the other mode -> confirmation dialog appears with "cancel the current connection" copy; Cancel preserves in-progress connect
- [ ] In `disconnected` state, tap the other mode -> no dialog, mode switches immediately (matches previous behavior)
- [ ] Backdrop click on the dialog -> treated as Cancel
- [ ] Rapid double-tap on Confirm -> only one invocation runs (`confirmInFlight` guard)

## Files changed

- `nym-vpn-app/src/screens/home/ModeToggle.tsx`
- `nym-vpn-app/src/i18n/en/home.json`
- `nym-vpn-app/CHANGELOG.md`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5505)

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation dialog when switching VPN modes (Fast ↔ Mixnet) if the VPN is connected, currently connecting, or about to auto-reconnect.

* **Localization**
  * Added English strings for the confirmation flow: separate titles/descriptions for connected vs connecting, plus confirm and cancel labels.

* **Documentation**
  * Updated changelog entry describing the new confirmation behavior.

### DIFF
--- a/nym-vpn-app/CHANGELOG.md
+++ b/nym-vpn-app/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Allow adding custom apps (user selected) to split tunneling for Windows & Linux
 
+### Changed
+
+- Mode switch (Fast/Mixnet) now asks for confirmation when the VPN is connected, connecting, or about to auto-reconnect. Previously, simply clicking the other mode tab would trigger an immediate disconnect/reconnect, surprising users who only wanted to view the other mode.
+
 ## [1.30.0] - 2026-05-29
 
 ### Fixed
--- a/nym-vpn-app/src/i18n/en/home.json
+++ b/nym-vpn-app/src/i18n/en/home.json
@@ -20,7 +20,15 @@
   "random-server": "Random server",
   "mode-toggle": {
     "fast": "Fast",
-    "mixnet": "Mixnet"
+    "mixnet": "Mixnet",
+    "switch-confirm": {
+      "title-connected": "Switch to {{mode}} mode?",
+      "title-connecting": "Switch to {{mode}} mode?",
+      "description-connected": "Switching modes will disconnect your VPN. You'll need to reconnect afterwards.",
+      "description-connecting": "Switching modes will cancel the current connection.",
+      "confirm": "Switch and disconnect",
+      "cancel": "Cancel"
+    }
   },
   "snackbar-disabled-message": {
     "connected": "Disabled while connected to VPN",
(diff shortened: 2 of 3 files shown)


vvmmaann:
Thanks @coderabbitai. Addressed in `2bad453`:

- **Double-confirm re-entry (Major).** Real bug, fixed. Added a synchronous `useRef` lock (`confirmLockRef`) checked and set at the top of `handleConfirmSwitch` and cleared in the `finally` block. `confirmInFlight` React state is retained purely to drive the Confirm button's loading UI and the cancel guard, so the two are intentionally separate, not redundant.

pronebird:
Hi, 

Thanks for submitting this PR. The app has multiple different settings that trigger automatic reconnect. Therefore I think a more systemic approach is necessary for consistency. Our UX team is notified and looking into it.

pronebird:
Hi,

Thank you for your PR. However the team decided against merging this feature. We do not see it necessary to confirm the switch between fast and anonymous modes. Many other settings reconnect the tunnel too. Some settings however apply at runtime with no reconnect.

PR #5561, opened 2026-06-11
Recover from bandwidth exceeded when VPN disconnects after daily reset

tommyv1987:
- When the account controller is in BandwidthExceeded error state and the VPN disconnects (API firewall down), it now triggers an immediate mandatory account sync.
- Fixes accounts stuck showing yesterday's depleted quota after UTC midnight reset while the VPN was still connected.
- Only applies to BandwidthExceeded - other error reasons are unchanged (flag cleared, no sync).

## Ticket

JIRA-NYM-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5561)

## Summary by CodeRabbit

* **Bug Fixes**
  * Firewall recovery now acknowledges a firewall-down event immediately and triggers an immediate network resync when appropriate, ensuring faster recovery from connectivity interruptions even while the controller is in an error state.

* **Tests**
  * Added integration tests validating firewall up/down behavior, confirming immediate resync behavior for both bandwidth-related and other error scenarios.

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/error_state.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/error_state.rs
@@ -127,9 +127,7 @@ impl<C: ConnectivityMonitor> AccountControllerStateHandler<C> for ErrorState {
                         return_sender.send(Ok(()));
                         if shared_state.firewall_active {
                             shared_state.firewall_active = false;
-                            if matches!(self.reason, AccountControllerErrorStateReason::BandwidthExceeded { .. }) {
-                                return NextAccountControllerState::NewState(SyncingNetworkState::enter(shared_state, SyncMode::Optimistic));
-                            }
+                            return NextAccountControllerState::NewState(SyncingNetworkState::enter(shared_state, SyncMode::Optimistic));
                         }
                     },
                     AccountCommand::VpnApiFirewallUp(return_sender) => {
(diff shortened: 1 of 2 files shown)


nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/error_state.rs
```diff
@@ -124,8 +124,13 @@ impl<C: ConnectivityMonitor> AccountControllerStateHandler<C> for ErrorState {
                     },
 
                     AccountCommand::VpnApiFirewallDown(return_sender) =>  {
-                        shared_state.firewall_active = false;
                         return_sender.send(Ok(()));
+                        if shared_state.firewall_active {
+                            shared_state.firewall_active = false;
+                            if matches!(self.reason, AccountControllerErrorStateReason::BandwidthExceeded { .. }) {
```
simonwicky: Don't match the error reason and trigger an Optimistic sync instead

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpn-account-controller/src/state_machine/error_state.rs
tommyv1987: Done 👍

simonwicky (approved):

PR #5569, opened 2026-06-12
fix: skip catch-all NAT masquerade when split tunneling is active on macOS

### LINKED ISSUE #5245: [macOS] Fix: catch-all NAT masquerade breaks split-tunnel-excluded apps on macOS 14.6–15.0
Below is a proposal for an improvement to split tunneling as this doesn't appear to work properly with my NetBird setup on macOS 14.8.5.

I've verified this vibecoded approach to the best of my knowledge, and it has been working for ~10 days now. Relevant patch and research as well as workaround script are included in the attached zip.

---

### Summary

When NymVPN (WireGuard mode) runs with split tunneling active and an excluded app operates
as a privileged root daemon — most notably **NetBird** — the app appears fully connected
but **no data flows between peers**. NetBird reports all peers as `Connected` with active
WireGuard handshakes, yet every ping times out and no TCP connections complete.

The split tunnel UI toggle for the excluded app appears to work but has **no effect**.

### Root cause

`get_nat_rules()` in `nym-vpn-core/crates/nym-firewall/src/macos.rs` unconditionally
adds catch-all NAT masquerade rules (`nat inet all` and `nat inet6 all`) on macOS
14.6–15.0, even when split tunneling is active.

The function previously destructured `FirewallPolicy` using `..` and discarded
`redirect_interface` — the field that indicates split tunneling is active. As a result,
every outbound TCP connection from excluded apps had its source IP rewritten to Nym's VPN
tunnel address (both IPv4 and IPv6). The relay server established its session against that
rewritten IP, so return traffic was routed back into Nym's tunnel rather than to the app's
real interface. **The connection was broken at the NAT layer, silently.**

**Note:** `route_everything_to()` is called in both modes — directly when split
tunneling is inactive, and via `get_split_tunnel_rules()` when split tunneling is active,
routing to the split tunnel utun for per-process classification. However, **pf NAT
processing occurs before filter rules**, so the catch-all NAT masquerade rewrites excluded
apps' source IPs before the split tunnel classifier can route them directly to en0.
The NAT layer is the sole cause.

Verified by:
- `pfctl -a nym -F rules` (flush filter only, keep NAT) + NetBird restart → fails ✗
- `pfctl -a nym -F all` (flush NAT + filter) + NetBird restart → works ✓
- Repeated consistently ✓

### Fix

Capture `redirect_interface` in `get_nat_rules()` and return early — after the specific
`no nat` exemptions — when split tunneling is active:

```rust
if redirect_interface.is_some() {
    tracing::debug!(
        "Split tunneling is active; skipping catch-all NAT masquerade to avoid \
         rewriting source IPs of excluded apps"
    );
    return Ok(rules);
}
```

The early return occurs before the loop that generates both `nat inet all` (IPv4) and
`nat inet6 all` (IPv6) catch-all rules — **both protocols are fixed by this single
change**.

When split tunneling is active, `get_split_tunnel_rules()` classifies traffic per-process
and routes excluded apps directly to the physical interface. Apple services are classified
as `Included` and continue traversing the VPN tunnel. The catch-all NAT masquerade is
therefore **redundant and harmful** in this configuration.

### IPv6

NetBird connects to relay servers via both IPv4 (`85.9.x.x`, `94.237.x.x`) and IPv6
(`2a04:3542:...`). The `nat inet6 all -> <tun_ipv6>` rule rewrites IPv6 relay connections
to Nym's IPv6 tunnel address (`fc01::...`) with the same effect as the IPv4 case.

Note: NetBird's management/signal servers (`api.netbird.io`, `signal.netbird.io`) share
the same `2a04:3542::/` IPv6 subnet as relay servers, making per-subnet static route
workarounds impractical — another reason the code fix (skipping NAT entirely in split
tunnel mode) is the right approach.

### Files changed

- `nym-vpn-core/crates/nym-firewall/src/macos.rs` — `get_nat_rules()`: capture
  `redirect_interface`; early-return when `Some`, skipping both IPv4 and IPv6 catch-all NAT
- `nym-vpn-core/CHANGELOG.md` — entry under `[Unreleased] > Fixed`

**2 files changed, 45 insertions(+)**

### Testing

Please verify on macOS 14.6–15.0 with split tunneling active:

- [ ] Excluded root daemon traffic (e.g. NetBird) flows without source IP rewriting (IPv4)
- [ ] Excluded root daemon traffic flows without source IP rewriting (IPv6)
- [ ] Apple services (iMessage, FaceTime) continue to route through the VPN tunnel
      without NAT masquerade — the split tunnel process classification should handle this,
      but warrants explicit confirmation on 14.6–15.0
- [ ] Behaviour unchanged on macOS < 14.6 and >= 15.1 (`NAT_WORKAROUND = false`)
- [ ] Kill switch remains effective when split tunneling is active

### User workaround (until fix ships)

Flush Nym's entire pf anchor after connecting. The WireGuard tunnel and kernel default
route via utun8 remain intact; only the strict kill switch is lost:

```bash
# Run after NymVPN connects:
sudo pfctl -a nym -F all
sudo netbird service stop && sleep 3 && sudo netbird service start
```

> ⚠️ Without Nym's `block drop quick all` rule, traffic will not be blocked if Nym
> disconnects unexpectedly. Users requiring strict kill switch behaviour should wait for
> the upstream fix.

### Environment

- macOS 14.8.5 (Sonoma) — MacBookPro12,1 (x86_64)
- NymVPN macOS app, WireGuard mode, split tunnel enabled for NetBird
- NetBird 0.68.3, userspace WireGuard, relayed connections via Upcloud Frankfurt/London (IPv4+IPv6)

### Related

Secondary issue (not fixed here): the split tunnel UI toggle for NetBird shows as enabled
but generates no effective pf exemption rules, because NetBird's daemon is a privileged
root process with no `.app` bundle that the process monitor can match. Users enabling the
toggle receive no feedback that it is silently ineffective.

---

[nym-split-bug-files.zip](https://github.com/user-attachments/files/27567020/nym-split-bug-files.zip)

mvanhorn:
## Ticket

Fixes #5245

## Description

WireGuard mode with split tunneling on macOS no longer NAT-masquerades excluded apps into the tunnel. `get_nat_rules()` now checks `policy.redirect_interface()` and skips the catch-all masquerade rules when split tunneling is active, keeping the no-nat exemptions (localhost, LAN nets, peer IPs, allowed endpoints, tun interfaces) in place in both branches.

Excluded apps running as privileged daemons (the reporter's NetBird case in #5245) die silently because pf NAT processing happens before filter rules, so `nat from 0.0.0.0/0 -> [tun ip]` rewrites their source IPs before the split tunnel classifier can route them to the physical interface, and return traffic flows back into the tunnel. The reporter verified with pfctl that flushing NAT plus filter rules fixes it while flushing filter rules alone does not. The masquerade exists only as a workaround for the macOS 14.6-15.1 source-IP bug (per the comment block above the function) and is incompatible with per-process classification, so the skip is logged at debug level. `cargo check` passes on the `nym-firewall` crate; behavior without split tunneling is unchanged.

## Checklist:

- [x] Changelog

Not UI related.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5569)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue on macOS where split tunneling mode was not properly preventing certain traffic from being routed through the VPN.


pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5569)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on mvanhorn).

___
*[`nym-vpn-core/crates/nym-firewall/src/macos.rs` line 390 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/5569#-OvGM-MI2NnLTnJ5NhvF:-OvGM-MI2NnLTnJ5NhvG:b-unqi7q) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/564840b8783f924ce055f22b60b9bc31153fd3a3/nym-vpn-core/crates/nym-firewall/src/macos.rs#L390)):*
> ```XML
>         }
> 
>         if policy.redirect_interface().is_some() {
> ```

Thanks for your contribution!

I think it makes sense. I glanced at this function and it seems that it sets a couple of exceptions ("no NAT") rules and then adds a single NAT rule to rewrite the rest of packets to vpn tunnel.

I wonder if we even need to set up any "no nat" rules considering that `nat_primary_to_tun` is not being set anymore? Perhaps returning very early on if `redirect_interface.is_some()` would be simpler?

pronebird (commented):
> @pronebird reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5569)* status: all files reviewed, 1 unresolved discussion (waiting on mvanhorn).

pronebird (approved):
Thanks for your contribution!

mvanhorn:
Thanks @pronebird for reviewing and merging the split-tunnel masquerade fix.

mvanhorn:
Thanks @pronebird! Skipping the catch-all NAT masquerade under split tunneling closes a sneaky leak path.

PR #5570, opened 2026-06-12
[Android] NYM-1500: Handle another vpn detection

mrkiwi:
## Ticket

JIRA-NYM-1500

## Description

Changes for `core` and Android side vpn lifecycle to handle another vpn detection. 

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5570)

## Summary by CodeRabbit

* **New Features**
  * Added detection of competing VPN apps with automatic disconnection handling.

* **Bug Fixes**
  * Improved error handling for VPN tunnel establishment failures.
  * Enhanced handling of lost VPN permissions during tunnel configuration.


agentpietrucha (approved):

agentpietrucha (approved):

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
```diff
@@ -254,26 +254,33 @@ impl ConnectedState {
...
+                tracing::info!("Tunnel provider failed. Disconnecting.");
+                NextTunnelState::NewState(
+                    DisconnectingState::enter(
+                        PrivateActionAfterDisconnect::Nothing,
+                        Some(self.tunnel_monitor_handle),
+                        shared_state,
+                    )
+                    .await,
+                )
+            }
```
neacsu: I think `ErrorStateReason::TunnelProvider` is a broader error, it doesn't necessarily mean there is another VPN running. Changing the behaviour for the tunnel state machine to go into disconnected when all other apps go into error state seems wrong. _Maybe_ it could be done if the error was more specific, pointing to something that exactly signifies there is a different vpn running.

neacsu (changes requested):
Is the running of a different VPN detected specifically or just guessed from the library error?
If it is precisely detected, it must be on the kotlin side. And if so, no need to include library changes, the connection should be prevented from triggering right from the app.

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
mrkiwi: Revert all core changes. Let's keep all the changes on Android side for now.

neacsu (approved):
We should make sure this gets planned for having a consistent approach, across different platforms

PR #5674, opened 2026-06-23
Swap internal resolution to Custom Resolver while in Connected state

jmwample:
## Ticket

JIRA-NYM-1374

## Description

In general once into the connected state the internal resolver should NOT be going to network for resolutions as it will provide responses from the static-pre-resolve that match the resolved (and firewall excepted) addresses for critical domains. 

However in the connected state if the internal `HickoryDnsResolver` is used for DNS lookups it will: 

* be slower than necessary because it uses DoT / DoH

* might get rate limited by the nameservers because all users per gateway will look like they come from the gateway IP. 

* On iOS the `HickoryDnsResolver` could send outside of the tunnel

#### Solution

On entering the Connected state swap the DNS resolver to system_resolver instead of the configured resolver. 

 On exiting the Connected state revert the internal resolver to use the custom cofigured DNS resolver.

## Checklist:

- [X] Changelog
- [X] Add swap to system resolver for internal DNS

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5674)

## Release Notes

* **Bug Fixes**
  * Improved DNS resolver handling during the tunnel connection lifecycle: the app uses system DNS while the tunnel is connected, and automatically reverts to the configured custom nameservers when disconnecting.

* **Documentation**
  * Updated the Unreleased changelog entry to clarify the new DNS resolver behavior during the Connected state and teardown.

### DIFF
--- a/nym-vpn-core/CHANGELOG.md
+++ b/nym-vpn-core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- While in Connected state swap internal resolver to use custom DNS (via system resolver). (https://github.com/nymtech/nym-vpn-client/pull/5674_
+
 ## [2026.11.0] - TBD
 
 ### Added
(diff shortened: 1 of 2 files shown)


nym-vpn-core/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- While in Connected state swap internal resolver to use custom DNS (via system resolver). (https://github.com/nymtech/nym-vpn-client/pull/5674_
```
neacsu, line 12: This should be under a section, maybe `Changed`?

neacsu (approved):

PR #5707, opened 2026-06-26
Exit bandwidth metadata uses netstack proxy on single-tun wireguard

tommyv1987:
- Single-tun wireguard (iOS and Android netstack) routes exit bandwidth metadata through the same local TCP proxy as entry, instead of binding HTTP to the utun interface name.
- Removes the immediate broken exit metadata dispatch that caused bandwidth check failures, tunnel shutdown, and roughly 13-second auto-reconnect loops after connect.

## Ticket

JIRA-NYM-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5707)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5139,7 +5139,7 @@ dependencies = [
 
 [[package]]
 name = "nym-apple-network"
-version = "2026.11.0-beta.3"
+version = "2026.11.0-beta.5"
 dependencies = [
  "block2",
  "dispatch2",
@@ -5248,7 +5248,7 @@ dependencies = [
 
 [[package]]
 name = "nym-cgroup"
-version = "2026.11.0-beta.3"
+version = "2026.11.0-beta.5"
 dependencies = [
  "anyhow",
  "log",
@@ -5396,7 +5396,7 @@ dependencies = [
 
 [[package]]
 name = "nym-common"
-version = "2026.11.0-beta.3"
+version = "2026.11.0-beta.5"
 dependencies = [
  "tracing",
  "tracing-test",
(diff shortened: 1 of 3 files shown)


neacsu (changes requested):
I don't think this change is correct, the endpoints were queried correctly before.

trojanfoe:
I hit a connection loop on my Linux test VM and this is what Claude thought about the logs I fed it:

```

  Root cause: The exit gateway bandwidth metadata check consistently fails with a TCP connect error, and the bandwidth controller tears down the tunnel after 2 consecutive failures, reconnects to the same
  gateway pair, and loops infinitely.

  Why the exit check fails (Linux dual-tun)

  In the dual-tun setup, both gateway bandwidth clients try to reach http://10.1.0.1:51830 (the same IP/port), but via different interfaces:

  - Entry gateway: bind_ip = 10.1.138.97 (client IP on tun0), send via tun0
  - Exit gateway: bind_ip = 10.1.18.84 (client IP on tun1), intended to go via tun1

  The route table has an explicit host route 10.1.0.1/32 → tun0. If SO_BINDTODEVICE("tun1") isn't being correctly applied by ReqwestClientBuilder::interface() in nym-http-api-client, the kernel follows the
  routing table and tries to send the exit check packet through tun0 with source IP 10.1.18.84. Since that IP isn't assigned to tun0, the TCP connect fails immediately (EADDRNOTAVAIL → wrapped as client
  error (Connect)).

  The entry check succeeds for the same reason in reverse: its bind_ip of 10.1.138.97 matches tun0 which is exactly what the routing table already uses, so no interface binding is needed.

  What the fix branch does (and doesn't fix)

  fix/core-exit-bandwidth-metadata-ulrichshorn introduces MetadataTcpProxy — a local loopback TCP listener that explicitly creates a new socket with SO_BINDTODEVICE and proxies connections — replacing the
  reqwest builder interface binding. This is more robust. However, looking at tunnel_monitor.rs in that branch, the fix only touches the TunnelInterface::One (single-tun mobile) path:

  TunnelInterface::Two { entry, exit } => {
      // unchanged from release branch — still sends TunnelMetadata, no proxy
      let (entry_event, exit_event) = two_tunnel_bandwidth_metadata_events(entry, exit);

  The Linux dual-tun (TunnelInterface::Two) case is not fixed in that branch. The same MetadataTcpProxy approach needs to be applied to the exit side of TunnelInterface::Two on Linux.

PR #5740, opened 2026-06-30
Bound QUIC bridge connect with timeout

tommyv1987:
## Description

The current behavior when a quic transport connection fails is to wait for the full timeout or allow the tunnel conection attempt to timeout which can be a long time (~60s).  

This PR reduces the time taken by a quic transport connection establishment failure. Returns `TransportError::TimedOut` so the tunnel monitor can retry with a different gateway within a more reasonable amount of time (10s).

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5740)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a timeout for QUIC connection attempts to prevent them from hanging indefinitely.
  * If the connection takes too long, the app now stops the attempt, closes the connection cleanly, and reports a timeout error with the duration included.


jmwample (approved):
Seems good to add a bounded timeout here other than the  overall tunnel connection timeout. 👍

PR #5751, opened 2026-07-01
NYM-4894: Re-use the firewall API endpoints if we already have them.

trojanfoe:
## Ticket

JIRA-NYM-4894

## Description

When connecting, re-use the firewall API endpoints if we already have them.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5751)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection setup by populating the connecting phase with resolved API socket addresses, ensuring the correct API endpoints are targeted during VPN establishment.
  * Improved resilience when resolved API endpoints are unavailable by using available address resolution results instead of starting empty.
  * Gateway caching now automatically pauses and resumes when networking is restricted or unrestricted, improving consistency and endpoint freshness.

### DIFF
--- a/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
+++ b/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
@@ -297,11 +297,6 @@ impl GatewayCache {
                         Command::Pause(paused) => {
                             tracing::info!("Gateway caching is {}", if paused { "Paused" } else { "Resumed" });
                             self.paused = paused;
-                            if !paused
-                                && self.connectivity_handle.connectivity().await.is_online()
-                            {
-                                self.perform_initial_fetch_once().await;
-                            }
                         }
                     }
                 }
--- a/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
+++ b/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
@@ -118,6 +118,16 @@ impl GatewayCacheHandle {
             .map_err(|_| Error::Cancelled)
     }
 
+    /// Pause or resume the background gateway cache refresh.
+    ///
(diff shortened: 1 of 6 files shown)


neacsu (approved):

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/gateway_cache.rs
```diff
@@ -23,6 +24,10 @@ impl GatewayCache for GatewayCacheHandle {
     async fn refresh_all(&self) -> Result<(), Error> {
         self.refresh_all().await
     }
+
+    fn set_paused(&self, paused: bool) {
```
tommyv1987: This silently swallows the errors - let's capture the errors here.

nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
```diff
@@ -279,10 +294,19 @@ impl GatewayCache {
                         Command::ClearCache => {
                             self.clear_cache();
                         }
+                        Command::Pause(paused) => {
```
tommyv1987, line 297: Maybe fire the initial fetch as a background task so the command channel stays responsive?

trojanfoe: I think that will lead to much more complication.  It's expected that the gateway cache can be busy sometimes, actually fetching gateways, so its loop will be unresponsive, however I don't think this new pause/resume logic will add any more delays.

Perhaps it's best to just set the pause flag and let the updating happen naturally as it was before?

tommyv1987: Agreed, that's cleaner, lets do that 👍

PR #5762, opened 2026-07-02
Fix tunnel health dual signal and mobile tcp

tommyv1987:
Fix iOS false reconnects when ICMP on utun fails but in-tunnel metadata/ bandwidth is healthy. Use TCP probes on mobile and defer teardown when both-leg metadata recently succeeded. 

Disclaimer here as well... We should do a follow  for zombie-connected edge case and ICMP-fallback + real-outage reconnect behavior.

On iOS we saw the tunnel auto-reconnect to the same gateways after ICMP probe failures on utun, while WireGuard and in-tunnel bandwidth/metadata checks were still succeeding seconds earlier. Users get an unnecessary VPN drop (no nym firewall on mobile).

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5762)

### DIFF
--- a/nym-vpn-core/Cargo.lock
+++ b/nym-vpn-core/Cargo.lock
@@ -5118,7 +5118,7 @@ dependencies = [
 
 [[package]]
 name = "nym-apple-network"
-version = "2026.11.0-beta.7"
+version = "2026.11.0-beta.8"
 dependencies = [
  "block2",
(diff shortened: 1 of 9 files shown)


trojanfoe (approved):
LGTM

PR #5792, opened 2026-07-07
Fix/nym 1561 reset dns on linux

trojanfoe:
## Ticket

JIRA-NYM-1561

## Description

Clean-up DNS on Linux when the system was rebooted while the VPN was up.

## Checklist:

- [ ] Changelog

This does add a nasty message to the log though:

image

Update: Fixed by removing logging from the Linux DNS implementations and instead returning more detailed error enum variants.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5792)


tommyv1987 (approved):

pronebird (changes requested):
It adds nasty message because tunnel_index = 0 which is invalid device id

See `nym-dns/src/linux/systemd_resolved.rs`

trojanfoe:
This change is nonsense as `tunnel_index` will never be valid in `reset()`.  Closing.

PR #5840, opened 2026-07-10
Add android to the same cfg that already binds on linux/ios

tommyv1987:
- Android InterfaceName metadata clients now call reqwest .interface(&interface) before bind IP

## Ticket

JIRA-NYM-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5840)


tommyv1987:
On a fresh Android install, two-hop WireGuard connected fine at first. After I turned ad blocking on, the app kept reconnecting: the tunnel handshake still succeeded, but connect then died about ten seconds later with "Metadata endpoints not reachable" the exit got blacklisted, and it looped. 

Looking at the logs, the entry bandwidth/metadata check over the local netstack proxy (127.0.0.1) usually worked, while the exit check straight to 10.1.0.1 often never came back - so both legs never passed the new wait for metadata before Connected gate.

pronebird (approved):

PR #5870, opened 2026-07-15
fix: allow forgetting account in non-connected tunnel states

### LINKED ISSUE #5668: minor bug: cannot forget account in some corner cases
```
nym-vpnc account forget
Failed to forget account: internal error: Unable to forget account while connected
Error: internal error: Unable to forget account while connected

nym-vpnc status
State: Disconnected
```
need `nym-vpnc disconnect`

`nym-vpnc 2026.11.0-beta.2`
git branch: release/2026.11-ulrichshorn

mvanhorn:
## Summary

`nym-vpnc account forget` now succeeds when the tunnel is in a non-connected state (`Error` or `Offline`), instead of failing with "Unable to forget account while connected".

## Why this matters

Reported in #5668: `account forget` returned `internal error: Unable to forget account while connected` even though `status` reported `State: Disconnected`, and the user only succeeded after an explicit `disconnect`. The guard in `handle_forget_account` rejected the operation whenever the state was `!= TunnelState::Disconnected`. But `TunnelState` has two other states documented as disconnected: `Error(_)` ("Tunnel is disconnected due to failure") and `Offline` ("Tunnel is disconnected, network connectivity is unavailable"). When the tunnel is stuck in `Error` or `Offline`, no tunnel holds the persistent storage the wipe needs to clear, so the strict equality check blocked forget unnecessarily. @pronebird confirmed on the issue that the block is only meant to apply while genuinely connected ("the daemon needs to disconnect in order for internal components to stop using persistent storage").

## Changes

- Add `TunnelState::is_tunnel_active()`, which returns `true` only for the states where a tunnel is actually up or transitioning (`Connecting`, `Connected`, `Disconnecting`) and `false` for the genuinely-disconnected states (`Disconnected`, `Error`, `Offline`). The match is exhaustive on purpose so a future variant forces the classification to be revisited.
- Use it in the forget-account guard (`vpn_service.rs`) in place of `!= TunnelState::Disconnected`, preserving the same error and message for the truly-active case. The change is scoped to the forget path only.

## Testing

Added unit tests for `is_tunnel_active()` covering the disconnected states (`Disconnected`, `Error`, `Offline { .. }`) returning `false` and a live state (`Disconnecting`) returning `true`.

Closes #5668

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5870)

## Summary by CodeRabbit

* **New Features**
  * Account removal now works while offline.
  * Account and device data is cleared even if VPN disconnection times out.
  * VPN disconnection is tracked and completed gracefully before data removal when needed.

* **Bug Fixes**
  * Fixed offline account removal being incorrectly rejected.
  * Improved handling and reporting when device removal, disconnection, or local data cleanup fails.
  * Account removal now transitions cleanly to the logged-out state after successful cleanup.

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/command_sender.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/command_sender.rs
@@ -125,6 +125,28 @@ impl AccountCommandSender {
         rx.await.map_err(AccountCommandError::internal)?
     }
 
+    /// Try to unregister the device from the API. This is best-effort and returns Ok
+    /// even if the unregister fails, allowing the caller to proceed with local cleanup.
+    #[instrument(skip(self))]
+    pub async fn unregister_device(&self) -> Result<(), AccountCommandError> {
+        let (tx, rx) = ReturnSender::new();
+        self.command_tx
+            .send(AccountCommand::UnregisterDevice(tx))
+            .map_err(AccountCommandError::internal)?;
+        rx.await.map_err(AccountCommandError::internal)?
+    }
+
+    /// Wipe local account data without attempting to unregister from the API.
+    /// Use this when the unregister has already been attempted separately.
+    #[instrument(skip(self))]
+    pub async fn wipe_local_account_data(&self) -> Result<(), AccountCommandError> {
+        let (tx, rx) = ReturnSender::new();
+        self.command_tx
+            .send(AccountCommand::WipeLocalAccountData(tx))
+            .map_err(AccountCommandError::internal)?;
+        rx.await.map_err(AccountCommandError::internal)?
+    }
+
     #[instrument(skip(self))]
     pub async fn rotate_keys(&self) -> Result<(), AccountCommandError> {
         let (tx, rx) = ReturnSender::new();
--- a/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/dispatch.rs
+++ b/nym-vpn-core/crates/nym-vpn-account-controller/src/commands/dispatch.rs
@@ -34,6 +34,12 @@ pub enum AccountCommand {
     /// Delete the stored account and every associated data
     ForgetAccount(ReturnSender<(), AccountCommandError>),
 
+    /// Unregister the device from the API (best-effort, silently continues on failure)
+    UnregisterDevice(ReturnSender<(), AccountCommandError>),
+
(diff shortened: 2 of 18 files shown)


pronebird (changes requested):
Hi,

I think instead we should disconnect the tunnel and then wipe all data. Some part of it may require sending API calls. Ideally those calls should happen inside of the tunnel if we have one.

We should not error in offline and instead wipe our local storage in which case device will remain registered online, so be it. 

We never meant to be using tunnel state to block any action.

mvanhorn:
You're right, thanks — reworked to your design in 395a19b. The tunnel-state gate is gone entirely.

Forget now requests `TargetState::Unsecured`, waits for the tunnel to settle, then wipes. `Disconnected`, `Offline` and `Error` skip the wait, so offline no longer errors — local storage is wiped and the device stays registered online, as you said.

Two details worth flagging:

The wait is bounded by the existing teardown limit rather than a new number. It was private to `nym-vpn-lib-uniffi` as `DISCONNECT_TIMEOUT`; I exported it from `nym-vpn-lib::service` as `VPN_DISCONNECT_TIMEOUT` and pointed the uniffi shutdown path at the same constant. On timeout the wipe still happens and the caller gets an explicit teardown-timed-out error — a wedged tunnel must not be able to prevent forgetting the account, since that was the original complaint in #5668.

My first cut waited by reading the tunnel event receiver inside the command handler. That preserved event order but let a pending forget delay every other main-loop branch. It now waits on a dedicated watch subscription, so the run loop stays the only reader.

Tests: disconnected wipes immediately, offline wipes without erroring, connected disconnects then wipes once settled, and a wedged `Disconnecting` hits an injected short timeout and wipes rather than hanging. I mutation-checked each one.

Not verified: no real device or tunnel here, so teardown behaviour is covered by tests rather than observed. The account-controller integration tests could not run in my sandbox either — WireMock can't bind a localhost port — so those assertions are unexercised on my side and rely on CI.

mvanhorn:
The flow is now unregister (while the tunnel can still route) → disconnect → wipe local.

Offline or a failed unregister no longer errors. We still wipe local, and the device may stay registered on the server. Tunnel state never blocks the action.

`UnregisterDevice` and `WipeLocalAccountData` are separate commands so a pending forget doesn't pin the command loop the way the old single task did.

PR #5900, opened 2026-07-20
Fix anonymous Mixnet MTU and retransmissions on desktop

justinhub0:
## Ticket

N/A (external reproduction)

## Description

Fix anonymous Mixnet website stalls on desktop by aligning the client TUN MTU
with the existing mobile behavior and restoring the SDK's normal retransmission
policy for client-to-IPR data messages.

Desktop clients currently default the Mixnet TUN to 1500 bytes, while iOS and
Android use 1280. The IPR Linux TUN defaults to 1420, and its connect response
does not negotiate an effective MTU. In a controlled IPR test, a 1500-byte
client completed TCP and TLS but then stalled receiving HTTP data; the exit
kernel emitted ICMP Fragmentation Needed and incremented `IpFragFails`.

The client-to-IPR v8 path also explicitly sets `max_retransmissions` to zero.
That makes the first ACK timeout discard the Mixnet message instead of applying
the SDK's normal retry policy, which can leave the encapsulated TCP flow stalled.

This PR:

- uses a 1280-byte default for anonymous Mixnet TUN devices on desktop, matching
  the existing iOS/Android value;
- caps Linux/Windows entry-route MTU detection at 1280 while preserving any
  lower detected value;
- continues to honor an explicitly configured Mixnet MTU;
- removes the request-side `with_max_retransmissions(0)` override so normal SDK
  retransmission behavior applies; and
- adds a regression test proving the retransmission fix does not change the
  public IPR payload, recipient, anonymous-message mode, reply SURB count, or
  transmission lane.

The change is limited to anonymous Mixnet IP routing. It does not modify the IPR
wire protocol, IPR server implementation, or WireGuard/two-hop behavior.

A locally built macOS client containing these two changes successfully loaded
sites that had previously stalled after the TLS ClientHello or during HTTP body
transfer.

## Validation

```text
cargo test -p nym-vpn-lib data_request_keeps_public_ipr_wire_format_with_normal_retransmissions --lib
cargo fmt -p nym-vpn-lib --check
git diff --check
```

The focused Rust test passes (`1 passed, 0 failed`). The macOS app and modern
installer package also built successfully with the embedded daemon.

## Checklist:

- [ ] Changelog

Not applicable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5900)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/processor.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/mixnet/processor.rs
@@ -356,10 +356,13 @@ impl MixnetMessageSinkTranslator for ToIprDataRequest {
         // sphinx packets that carry the actual data, since we try to keep the payload for IP
         // traffic contained within a single sphinx packet.
         let surbs = 0;
-        Ok(
-            InputMessage::new_anonymous(self.recipient, packet, surbs, lane, packet_type)
-                .with_max_retransmissions(0),
-        )
+        Ok(InputMessage::new_anonymous(
+            self.recipient,
+            packet,
+            surbs,
+            lane,
+            packet_type,
+        ))
     }
 }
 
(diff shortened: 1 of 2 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/mixnet/processor.rs
```diff
@@ -356,10 +356,13 @@ impl MixnetMessageSinkTranslator for ToIprDataRequest {
         // sphinx packets that carry the actual data, since we try to keep the payload for IP
         // traffic contained within a single sphinx packet.
         let surbs = 0;
-        Ok(
-            InputMessage::new_anonymous(self.recipient, packet, surbs, lane, packet_type)
-                .with_max_retransmissions(0),
```
mfahampshire, line 361: I don't think we should remove the `.with_max_retransmissions` - from a branch on a parallel problem, its an issue with the IPR that I'm fixing over there that has to do with MTU

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
```diff
@@ -99,11 +99,7 @@ use crate::{
 };
 
 /// Default MTU for mixnet tun device.
-const DEFAULT_TUN_MTU: u16 = if cfg!(any(target_os = "ios", target_os = "android")) {
```
mfahampshire, line 102: The actual MTU expected by the IPR is 1420 - hardcoding 1280 (mobile) is wasteful for desktop apps, so I think we should change this. 

I am modifying the IPRs to (a) accept 1500 byte MTUs, as per the expection of much of the non-mobile internet, and (b) advertise to clients what their max MTU size is on connection, which can then be used by clients to set their expectations accordingly. 

We could patch for now and just change line 105 from `1500` -> `1420`, and in the future add the logic to check what the connection response in the new wire version contains?

mfahampshire (changes requested):

nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel_monitor.rs
pronebird: I concur. Keep mobile as is, for desktop branch change 1500 -> 1420 as per @mfahampshire suggestion.

Clamping at DEFAULT_TUN_MTU also looks good

pronebird (changes requested):

PR #5905, opened 2026-07-22
Close secondary 8 MiB frame-cap gap that #5275 left asymmetric

tommyv1987:
## Description

Forwarder codec now lifts the 8 MiB default cap to usize::MAX, matching MultiplexCodec in test-rpc (PR #5275 .max_frame_length(usize::MAX)). Removes the asymmetry where a >8 MiB gRPC frame would error this codec, drop the forwarder, and time out every subsequent RPC.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5905)

### DIFF
--- a/nym-vpn-core/crates/test/test-manager/src/nym_daemon.rs
+++ b/nym-vpn-core/crates/test/test-manager/src/nym_daemon.rs
@@ -11,7 +11,7 @@ use hyper_util::rt::TokioIo;
 use nym_vpn_proto::rpc_client::RpcClient as NymProxyClient;
 use test_rpc::transport::{ConnectionHandle, GrpcForwarder};
 use tokio::io::{AsyncReadExt, AsyncWriteExt, DuplexStream};
-use tokio_util::codec::{Decoder, LengthDelimitedCodec};
+use tokio_util::codec::LengthDelimitedCodec;
 use tower::Service;
 
 const GRPC_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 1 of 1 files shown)


pronebird (approved):

nym-vpn-core/crates/test/test-manager/src/nym_daemon.rs
```diff
@@ -71,7 +71,11 @@ pub fn new_rpc_client(
     connection_handle: ConnectionHandle,
     nym_daemon_transport: GrpcForwarder,
 ) -> RpcClientProvider {
-    let mut framed_transport = LengthDelimitedCodec::new().framed(nym_daemon_transport);
+    // Match the daemon-side/test-rpc MultiplexCodec, which lifts the frame cap to
+    // usize::MAX.
+    let mut framed_transport = LengthDelimitedCodec::builder()
```
pronebird, line 76: MultiplexCodec in `test/test-rpc/src/transport.rs` is created as:

```rs
const MULTIPLEX_LEN_DELIMITED_HEADER_SIZE: usize = 4;

let mut codec_builder = LengthDelimitedCodec::builder();

codec_builder
            .length_field_length(MULTIPLEX_LEN_DELIMITED_HEADER_SIZE)
            .max_frame_length(usize::MAX);
```

However `length_field_length` is not present in `test/test-manager/src/nym_daemon.rs`. Does it need to be aligned too?

pronebird: Actually, seems like default is `4` so maybe not, however for `usize::MAX`, I'd perhaps make sure that `usize` can really fit, i.e:

```rs
 // This does: std::mem::size_of::<usize>() which is `8`!
codec_builder..length_field_type::<usize>()
```

Make sure to align transport and nym_daemon implementations

PR #5939, opened 2026-07-28
Fixing the test harness

tommyv1987:
## Description

Phwoar this was a battle....

- Fixes the flaky Debian12 E2E suite, which was failing across most tests for reasons unrelated to the VPN client itself.

- The serial link between the test host and the guest VM now has a real handshake, so a recovered session can no longer be mistaken for a closed one and silently kill the next test.

- Tunnel and account state are observed from inside the guest over short polls instead of a long-lived event stream, which removes the main source of wedged daemon sessions.

- The harness recovers from a broken daemon connection instead of poisoning every following test, and the guest console no longer corrupts the serial line.

- DNS checks inside the VM retry for a shared budget after connect, because the tunnel reports Connected before the exit gateway resolver starts answering.

- Blocklist tests that deliberately block the public resolvers now verify connectivity by IP rather than trying to resolve hostnames through the resolvers they just blocked.

- The DNS leak test understands systemd-resolved, checking the nameservers on the tunnel link rather than a resolv.conf the daemon never rewrites, and proves the pre-VPN resolver is unreachable while connected.

- Failures now capture per-test daemon logs, resolver state, and upstream DNS reachability, so a red run says why rather than just that it timed out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5939)

## Summary by CodeRabbit

* **New Features**
  * Added device cleanup before end-to-end test runs to ensure a fresh account state.
  * Added daemon log and guest diagnostics collection when tests fail.
  * Added observed tunnel and account state monitoring with timeout-aware waiting.
  * Added automatic recovery for interrupted daemon communication and test sessions.

* **Bug Fixes**
  * Improved IPv6 traffic blocking and restoration behavior.
  * Added SSH operation timeouts and clearer command failure reporting.
  * Improved transport resynchronization and session handling for more reliable tests.

* **Tests**
  * Expanded coverage for DNS leaks, connectivity blocking, retries, recovery, timeouts, and transport synchronization.


jalolk (approved):
Amazing work!

trojanfoe (approved):

PR #5973, opened 2026-07-31
VPN-4908: Avoid DNS on macOS from setting DNS on invalid interfaces.

trojanfoe:
## Ticket

JIRA-VPN-4908

## Description

Some network services can have a leftover `.../DNS` key in the dynamic store with no corresponding `.../Interface` entry (e.g. disabled or orphaned services). Writing DNS to those never sticks, and the write itself re-triggers our own change notification, so treating them as manageable causes a self-sustaining save/notify loop. Skip them entirely.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/5973)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS state handling on macOS by ignoring entries that are not associated with a valid network interface.
  * Prevents invalid or incomplete network configuration data from affecting DNS behavior.

### DIFF
--- a/nym-vpn-core/crates/nym-dns/src/macos.rs
+++ b/nym-vpn-core/crates/nym-dns/src/macos.rs
@@ -514,6 +514,9 @@ fn read_all_dns(store: &SCDynamicStore) -> HashMap<ServicePath, Option<DnsSettin
         for state_path in paths.iter() {
             let state_path_str = state_path.to_string();
             let setup_path_str = state_to_setup_path(&state_path_str).unwrap();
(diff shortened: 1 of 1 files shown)


neacsu (approved):

pronebird (changes requested):
I don't understand the issue.

trojanfoe:
We cannot reproduce the issue, even when adding a network interface and disabling it, so we are closing this so we don't introduce unverified changes.

PR #6016, opened 2026-08-05
Keep an Android VPN interface up while switching servers so it cannot leak

tommyv1987:
## Description

- On Android server switch, a blocking placeholder VPN interface is installed before the old tunnel is released, so other apps cannot fall back to the ISP path during reconnect.
- The VPN app itself is excluded from that placeholder so control-plane registration can still reach gateways; other apps stay covered.
- Cold connect does not use the placeholder, avoiding a regression where entry registration timed out.
- Error state uses the same cover instead of leaving Android traffic uncovered.
- Intentional disconnect still tears everything down so normal internet returns.
- Shared blocking settings live in one Rust module with unit tests for settings and install-before-release ordering.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6016)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
@@ -56,6 +56,10 @@ impl ConnectedState {
         tunnel_monitor_event_receiver: TunnelMonitorEventReceiver,
         shared_state: &mut SharedState,
     ) -> (Box<dyn TunnelStateHandler>, PrivateTunnelState) {
+        // Real tunnel replaced the blocking placeholder via configure_tunnel; drop the stale FD.
+        #[cfg(target_os = "android")]
+        shared_state.clear_android_blocking_tun();
+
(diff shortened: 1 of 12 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/connected_state.rs
```diff
@@ -56,6 +56,10 @@ impl ConnectedState {
         tunnel_monitor_event_receiver: TunnelMonitorEventReceiver,
         shared_state: &mut SharedState,
     ) -> (Box<dyn TunnelStateHandler>, PrivateTunnelState) {
+        // Real tunnel replaced the blocking placeholder via configure_tunnel; drop the stale FD.
+        #[cfg(target_os = "android")]
+        shared_state.clear_android_blocking_tun();
```
pronebird, line 64: I think that what's missing in this PR is the following line in `ConnectingState::enter()`:

```rs
#[cfg(target_os = "android")]
ensure_android_blocking_tun();
```

pronebird (changes requested):

PR #6055, opened 2026-08-10
Fix Split Tunnel IPv6 edge-case (#6052)

trojanfoe:
If the default internet doesn't have an IPv6 address but the tunnel does have an IPv6 address, Split Tunnelling can fail if the app uses IPv6.

This change also means the split tunnelled app loses IPv6 access, which should perhaps be surfaced to the user via the UI?

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6055)

### DIFF
--- a/nym-vpn-core/CHANGELOG.md
+++ b/nym-vpn-core/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+
+
 ## [2026.12.0] - TBD
 
 ### Added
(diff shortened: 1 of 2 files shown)


nym-vpn-core/CHANGELOG.md
```diff
@@ -5,14 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+
+## [2026.12.0] - TBD
```
pronebird: Leave it as unreleased for release workflows to function properly

pronebird (changes requested):

nym-vpn-core/CHANGELOG.md
trojanfoe: Done.

pronebird (approved):

PR #6068, opened 2026-08-11
Make resolver more resilient.

trojanfoe:
- `ResolverOverrides::from_urls` now tracks each candidate domain as either a primary endpoint or a front (decoy fallback), instead of flattening them into one undifferentiated set.
- A front that fails to resolve is dropped and logged as a warning; it just won't be available as a fallback route.
- The whole call only fails (`HostnamesResolutionError`) if no primary URL at all resolved, meaning there's truly no usable route.
- If some but not all primaries resolve, it proceeds with what worked and logs a warning about the ones that didn't.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6068)

### DIFF
--- a/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/gateway.rs
+++ b/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/gateway.rs
@@ -137,7 +137,7 @@ impl GatewayDiagnostic {
             None,
         )?;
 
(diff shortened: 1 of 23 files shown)


pronebird (approved):

trojanfoe:
This is unlikely to make the release, so I have moved it over to `develop` branch: https://github.com/nymtech/nym-vpn-client/pull/6077

PR #6077, opened 2026-08-12
VPN-4909: Make resolver more resilient.

trojanfoe:
## Ticket

JIRA-VPN-4909

## Description

Don't error if one or more of the domain fronting hosts fails to resolve. Remove failed fronting hosts from the HTTP clients.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6077)

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved asynchronous initialization for VPN, gateway, and network connections.
  * Enhanced route resolution to remove unavailable fronting routes while retaining usable alternatives.
  * Improved handling when primary connection routes cannot be resolved.
  * Preserved cached network configuration when refreshed connection setup fails.
* **Bug Fixes**
  * Prevented connection failures caused by unresolved optional fronting domains.
  * Preserved fallback routes when route pruning would otherwise remove all available options.

### DIFF
--- a/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/gateway.rs
+++ b/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/gateway.rs
@@ -137,7 +137,7 @@ impl GatewayDiagnostic {
             None,
         )?;
 
-        GatewayClient::new(config, new_user_agent!())
+        GatewayClient::new(config, new_user_agent!()).await
     }
 
     async fn lookup_gateway(
--- a/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
+++ b/nym-vpn-core/crates/nym-diagnostic/src/diagnostic/http.rs
@@ -11,7 +11,7 @@ use std::fmt;
 use nym_http_api_client::{Client, ClientBuilder, FrontPolicy, Url};
 use nym_platform_metadata::new_user_agent;
 use nym_validator_client::nym_api::NymApiClientExt;
-use nym_vpn_api_client::{VpnApiClient, api_urls_to_urls};
+use nym_vpn_api_client::{ResolverOverrides, VpnApiClient, api_urls_to_urls};
 use nym_vpn_lib_types::{
     ApiTimeSkew, ApiUrl, DiagnosticEndpointResponse, DiagnosticResult, HttpReport,
 };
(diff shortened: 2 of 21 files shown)


pronebird:
*[`nym-vpn-core/crates/nym-vpn-api-client/src/types/resolver_overrides.rs` line 190 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/6077#-OzpdMxw8Yd0j_lkGMJS:-OzpdMxw8Yd0j_lkGMJT:b3mmnuq) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/d6f22c6113563100e7273773920609395b8d30a8/nym-vpn-core/crates/nym-vpn-api-client/src/types/resolver_overrides.rs#L190)):*
> ```XML
>                 }
> 
>                 Url::new(
> ```

Re-creation of `Url` would reset internal index that points to the currently used front.  See `new` implementation for `Url`:

```rs
impl Url {
    /// Create a new `Url` instance with the given something that can be parsed as a  URL and
    /// optional tunneling domains
    pub fn new<U: reqwest::IntoUrl>(
        url: U,
        fronts: Option<Vec<U>>,
    ) -> Result<Self, reqwest::Error> {
        let mut url = Self {
            url: url.into_url()?,
            fronts: None,
            current_front: Arc::new(AtomicUsize::new(0)), // This
        };
```

Perhaps it would be better to avoid re-creating `Url` when possible to retain the current front?

@jmwample this behavior is very opaque on the outside and perhaps `current_front` should be removed from `Url`. 

Also we should rename that type to something else to avoid confusion with `url::Url`. Perhaps `FrontedUrl`?

trojanfoe (commented):
> @trojanfoe made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6077)* status: 0 of 21 files reviewed, all discussions resolved (waiting on jmwample, neacsu, and pronebird).

___
*[`nym-vpn-core/crates/nym-vpn-api-client/src/types/resolver_overrides.rs` line 190 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/6077#-OzpdMxw8Yd0j_lkGMJS:-Ozpo9Ja3BL4D3pHxk4H:b-896fix) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/d6f22c6113563100e7273773920609395b8d30a8/nym-vpn-core/crates/nym-vpn-api-client/src/types/resolver_overrides.rs#L190)):*

Re-creation of `Url` would reset internal index that points to the currently used front.  See `new` implementation for `Url`:

```rs
impl Url {
    /// Create a new `Url` instance with the given something that can be parsed as a  URL and
    /// optional tunneling domains
    pub fn new<U: reqwest::IntoUrl>(
        url: U,
        fronts: Option<Vec<U>>,
    ) -> Result<Self, reqwest::Error> {
        let mut url = Self {
            url: url.into_url()?,
            fronts: None,
            current_front: Arc::new(AtomicUsize::new(0)), // This
        };
```

Perhaps it would be better to avoid re-creating `Url` when possible to retain the current front?

@jmwample this behavior is very opaque on the outside and perhaps `current_front` should be removed from `Url`. 

Also we should rename that type to something else to avoid confusion with `url::Url`. Perhaps `FrontedUrl`?
</blockquote></details>

Done.

jmwample:
Artificially created a situation where this occurs (nameservers blocked and domains removed from fallback dns) and the vpn was able to make progress with this change. Debug logs attached.

[vpnd.log](https://github.com/user-attachments/files/31281216/vpnd.log)

jmwample:
One important note on the way this functions. Doing `resolve_and_prune` as an async operations makes any caller wait for the function to complete. This means that in the cases where pruning is relevant (i.e. nameservers are blocked and resolution fails) you have to wait the async timeout (10s) everywhere that you add this in.

PR #6087, opened 2026-08-13
Prevent refresh storm

pronebird:
## Description

`GatewayCache::lookup_gateways` already performs network request when gateways become stale. Since call to `GatewayCache::refresh_all()` happens too frequently, it could double number of requests in case of failure to connect to our API. Instead, let's rely on `GatewayCache::lookup_gateways` alone.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087)

## Summary by CodeRabbit

* **Changes**
  * Gateway information is no longer refreshed automatically when the gateway client is replaced.
  * Gateway refreshes continue to occur during initial setup and after connectivity changes.
  * Removed the manual “refresh all gateways” operation from the available interface.

### DIFF
--- a/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
+++ b/nym-vpn-core/crates/nym-gateway-directory/src/gateway_cache.rs
@@ -33,14 +33,6 @@ impl GatewayCacheHandle {
         Self { tx }
     }
 
-    /// Refresh all gateways and countries without blocking until the operation is complete.
-    pub async fn refresh_all(&self) -> Result<()> {
-        self.tx.send(Command::RefreshAll).map_err(|_| {
-            tracing::error!("Gateway cache command channel closed (RefreshAll)");
-            Error::Cancelled
-        })
-    }
-
     /// Lookup gateways waiting for any pending fetch request or initiating one if needed.
     pub async fn lookup_gateways(&self, gw_type: GatewayType) -> Result<GatewayList> {
         let (tx, rx) = tokio::sync::oneshot::channel();
@@ -190,7 +182,6 @@ impl GatewayCacheHandle {
 }
 
 enum Command {
-    RefreshAll,
     LookupGateways(
         GatewayType,
         tokio::sync::oneshot::Sender<Result<GatewayList>>,
@@ -281,9 +272,6 @@ impl GatewayCache {
             tokio::select! {
                 Some(cmd) = self.command_rx.recv() => {
                     match cmd {
-                        Command::RefreshAll => {
-                            self.refresh_all().await;
-                        }
                         Command::LookupGateways(gw_type, tx) => {
                             tx.send(self.lookup_gateways(gw_type).await).ok();
                         }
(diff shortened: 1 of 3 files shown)


neacsu (changes requested):
> @neacsu made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on pronebird and trojanfoe).

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/mod.rs` line 280 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087#-OzuVuM9E61dygvhJkUW:-OzuVuM9E61dygvhJkUX:bwygus0) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/48a12dc1cb462599fc5a874a28087b1991ec57fa/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/mod.rs#L280)):*
> ```XML
>             .replace_gateway_client(gateway_client)
>             .ok();
>         self.gateway_cache.refresh_all().await.ok();
> ```

But cache could be invalidated by the `replace_gateway_client` call, so a refresh would have to be called. The `refresh_all` call doesn't do a query per se, it only does it if the cache was indeed invalidated.
Perhaps the
```
if new_config.min_gateway_performance() != old_config.min_gateway_performance() {
            self.cached_gateways.clear();
            self.cached_nymnodes = None;
        }
```
could be modified to be leaner on clearing the cache completely and it could just update the cache in some situations (when the gateway performance change moves in a direction where we can just trim the existing cache, no network call needed)

neacsu (commented):
> @neacsu made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087)* status: 0 of 3 files reviewed, all discussions resolved (waiting on trojanfoe).

___
*[`nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/mod.rs` line 280 at r1](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087#-OzuVuM9E61dygvhJkUW:-OzuYPXG9ONgzn3WoNHh:b-y45sg3) ([raw file](https://github.com/nymtech/nym-vpn-client/blob/48a12dc1cb462599fc5a874a28087b1991ec57fa/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/mod.rs#L280)):*

But cache could be invalidated by the `replace_gateway_client` call, so a refresh would have to be called. The `refresh_all` call doesn't do a query per se, it only does it if the cache was indeed invalidated.
Perhaps the
```
if new_config.min_gateway_performance() != old_config.min_gateway_performance() {
            self.cached_gateways.clear();
            self.cached_nymnodes = None;
        }
```
could be modified to be leaner on clearing the cache completely and it could just update the cache in some situations (when the gateway performance change moves in a direction where we can just trim the existing cache, no network call needed)
</blockquote></details>

On a second thought, the cache should be repopulated every time it is needed, so we might gain some savings here, considering the query per type we use now

neacsu (approved):
> @neacsu reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6087)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on trojanfoe).

PR #6124, opened 2026-08-18
fix(android): trim v2026.12.0 Play changelog under 500-char limit

sallysundries:
Google Play caps release notes at 500 chars; en-US changelogs/20261200.txt was 662, causing 'metadata character limit exceeded' in publish-nym-vpn-android. Trims it to 493 (all six items kept) and adds a VERSION_CODE-scoped pre-flight check that runs before fastlane supply.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6124)

### DIFF
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -30,3 +30,15 @@ jobs:
     steps:
       - if: github.event.pull_request.milestone == null && contains( env.LABELS, 'no-milestone' ) == false
         run: exit 1
+
+  check-changelog-length:
+    name: Play Store changelog length
+    runs-on: arc-linux-latest
+    # Only run on real code changes, not label/milestone toggles.
+    if: contains( fromJSON('["opened","reopened","synchronize"]'), github.event.action )
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 1
+      - name: Check release-notes length (<= 500 chars)
+        run: bash .github/scripts/release/check-changelog-length.sh
--- a/.github/scripts/release/check-changelog-length.sh
+++ b/.github/scripts/release/check-changelog-length.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+# Fail fast if the release notes about to be published exceed Google's limit.
+# fastlane supply uploads the changelog whose filename matches the AAB's
+# version code, so this guard checks exactly changelogs/<VERSION_CODE>.txt in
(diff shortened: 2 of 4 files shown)


pronebird:
`publish-nym-vpn-android` runs during the last phase of release workflow. How about introducing a new workflow that can catch that issue much sooner at PR level?

I am thinking of a workflow that tracks changes to `fastlane/metadata/android/en-US/changelogs` and checks the length of new changelog files used by Google Play.

pronebird (changes requested):
See my comment above

sallysundries:
> `publish-nym-vpn-android` runs during the last phase of release workflow. How about introducing a new workflow that can catch that issue much sooner at PR level?
> 
> I am thinking of a workflow that tracks changes to `fastlane/metadata/android/en-US/changelogs` and checks the length of new changelog files used by Google Play.

Patched to wire the gate into the PR checks. a new check-changelog-length job in pr-validation.yaml that runs on every PR to develop and release/**. So an over-limit release note now fails a fast PR check — minutes in, on the "Play Store changelog length" job — instead of surviving all the way to the publish step. The check reuses the same VERSION_CODE-scoped script, so it only flags the changelog actually being shipped and won't trip on historical notes.

pronebird (approved):

.github/workflows/pr-validation.yaml
```diff
@@ -30,3 +30,15 @@ jobs:
     steps:
       - if: github.event.pull_request.milestone == null && contains( env.LABELS, 'no-milestone' ) == false
         run: exit 1
+
+  check-changelog-length:
```
pronebird, line 34: Nit: it would be better to add a separate workflow that is only triggered on changes to changelog. In contrast, pr-validation PR runs very frequently and changelog check is not needed if none of files have changed.

For example:

```yml
name: check-android-changelog-length
on:
  pull_request:
    types: [opened, synchronize, reopened]
    paths:
      - ".github/workflows/check-android-changelog-length.yml"
      - "fastlane/metadata/android/**" # NOTE how this would only trigger on changes to metadata
jobs:
  check-changelog-length:
    runs-on: arc-linux-latest
    steps:
      - uses: actions/checkout@v7
        with:
          fetch-depth: 1
      - name: Check release-notes length (<= 500 chars)
        run: bash .github/scripts/release/check-changelog-length.sh
```

pronebird:
This must be fixed by now

PR #6150, opened 2026-08-21
Refresh clock-skew cache when the system clock changes.

trojanfoe:
Refresh clock-skew cache when the system clock jumps or a gateway rejects our timestamp

  Registration was failing with "Gateway rejected registration: invalid timestamp"
  after the system clock was corrected mid-run, because SkewManager cached the
  device/API clock-skew for up to 4h and kept applying the stale offset to an
  already-correct clock. Only a service restart cleared it.

  - SkewManager now detects clock jumps opportunistically inside cached_skew(),
    comparing elapsed wall-clock time against elapsed monotonic time since the
    last call, and evicts the cache if they've diverged - no extra background
    task needed, since it piggybacks on however often callers already read it.
  - tunnel_monitor now force-resyncs the skew immediately when LP registration
    is rejected specifically for an invalid timestamp, instead of waiting for
    the next gateway attempt to reuse the same stale value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6150)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-api-client/src/skew_manager.rs
+++ b/nym-vpn-core/crates/nym-vpn-api-client/src/skew_manager.rs
@@ -170,15 +170,13 @@ impl SkewManager {
             let monotonic_elapsed =
                 TimeDuration::try_from(monotonic - last_monotonic).unwrap_or(TimeDuration::ZERO);
 
-            if is_clock_jump(wall_elapsed, monotonic_elapsed) {
+            if is_clock_jump(wall_elapsed, monotonic_elapsed)
+                && let Ok(mut state) = self.inner.skew_state.try_write()
+            {
+                state.take();
                 tracing::info!(
-                    wall_elapsed = %wall_elapsed,
-                    monotonic_elapsed = %monotonic_elapsed,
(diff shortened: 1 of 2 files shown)


nym-vpn-core/crates/nym-vpn-api-client/src/skew_manager.rs
```diff
@@ -124,6 +134,57 @@ impl SkewManager {
...
+        if let Some((last_wall, last_monotonic)) = *watch {
+            let wall_elapsed = wall - last_wall;
+            let monotonic_elapsed =
+                TimeDuration::try_from(monotonic - last_monotonic).unwrap_or(TimeDuration::ZERO);
+
+            if is_clock_jump(wall_elapsed, monotonic_elapsed) {
+                tracing::info!(
+                    wall_elapsed = %wall_elapsed,
+                    monotonic_elapsed = %monotonic_elapsed,
+                    "Detected a system clock change; invalidating cached VPN API time skew"
```
neacsu: This log should be printed inside the `if` block bellow, since that's where the cache invalidation actually happens, no?

neacsu (approved):

PR #6173, opened 2026-08-22
(libwg) Fix MTE crash on dVPN connect by replacing C.GoString

rachyandco:
Go's C.GoString locates the NUL terminator via runtime.findnull, whose arm64 IndexByteString does a 32-byte-aligned vector load that reads up to 31 bytes outside the string's allocation. With ARM MTE enabled (e.g. GrapheneOS on Pixel 8/9), the uapi-config CStrings passed from Rust live in 16-byte-tagged Scudo granules, so the out-of-bounds read hits a differently-tagged granule and kills the process with SEGV_MTESERR (crash in libwg.so at internal/bytealg.IndexByteString, right after "Created tun device"). Only dVPN mode loads the Go library, which is why mixnet was unaffected. Triggering depends on allocation alignment (~50% per string), not on app version.

Replace all C.GoString call sites with goStringFixed(), which finds the length byte-by-byte and copies via the strictly-in-bounds C.GoStringN — same approach as mullvad/mullvadvpn-app#6727.

Verified on Pixel 9 Pro / GrapheneOS with MTE force-enabled: unpatched build crashes on first dVPN connect at pc 0x150c08; patched build survives 6 consecutive tunnel establishments.

## Ticket

JIRA-NYM-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6173)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of configuration, endpoint, address, and DNS values across supported platforms.
  * Prevented potential memory overreads when processing tunnel setup and network configuration data.
  * Preserved existing validation, error handling, and tunnel behavior while improving stability.

### DIFF
--- a/wireguard/libwg/libwg.go
+++ b/wireguard/libwg/libwg.go
@@ -30,6 +30,25 @@ type TunnelContext struct {
 	Logger *device.Logger
 }
 
+// goStringFixed converts a NUL-terminated C string to a Go string without
+// reading outside the string's allocation. The standard C.GoString locates
+// the terminator with vectorized scans that read up to 31 bytes beyond it and
+// before the start of the string; under ARM MTE (16-byte tag granules) those
+// out-of-bounds reads hit differently-tagged granules and kill the process
+// with SEGV_MTESERR. See https://github.com/mullvad/mullvadvpn-app/pull/6727.
+func goStringFixed(cString *C.char) string {
+	if cString == nil {
+		return ""
+	}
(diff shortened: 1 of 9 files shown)


pronebird:
I think that this issue must have been fixed in go1.27 (https://go-review.googlesource.com/c/go/+/749062)

We may avoid this patch altogether by simply updating golang

PR #6174, opened 2026-08-23
Don't buffer offline errors in gateway selection

nil0ft:
## Ticket

n/a

## Description

First connect after boot takes ~100s on linux. Restarting vpnd fixes it.

What I found:
- its not slow connecting, it fails 10x then works on the 11th
- every failure is instant, "failed to lookup gateways / no connectivity"
- but 1ms later the cache logs "returning 569 cached gateways" so its not offline
- the 100s is just backoff piling up

continuous_select buffers selections ahead and errors get buffered too, so
offline errors from a blip at boot get eaten by connect attempts later when the
network is fine again.

Patch skips buffering those and retries instead, with a small backoff so it
doesnt spin when actually offline.

been running it for a couple days now, connects are instant

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6174)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway selection recovery after temporary offline errors.
  * Prevented transient failures from being buffered, reducing delays before the first connection after startup.
  * Added controlled retry timing that increases briefly between attempts and resets after successful recovery.
  * Improved connection responsiveness when connectivity returns, helping avoid unnecessary delays during startup and temporary network interruptions.

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/gateway_cache.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/gateway_cache.rs
@@ -27,7 +27,13 @@ impl GatewayCache for GatewayCacheHandle {
 
 #[cfg(test)]
 pub mod tests {
-    use std::{sync::Arc, time::Duration};
+    use std::{
+        sync::{
+            Arc,
+            atomic::{AtomicBool, AtomicUsize, Ordering},
+        },
+        time::Duration,
+    };
 
     use nym_gateway_directory::Gateway;
     use tokio::sync::RwLock;
@@ -40,6 +46,8 @@ pub mod tests {
         /// Artificial delay applied to every `lookup_gateways` call, used to
         /// model a fresh selection that hasn't landed a value in the stream yet.
         lookup_delay: Duration,
+        offline: Arc<AtomicBool>,
+        lookups: Arc<AtomicUsize>,
     }
 
     impl MockGatewayCache {
@@ -47,6 +55,8 @@ pub mod tests {
             Self {
                 gateways,
                 lookup_delay: Duration::ZERO,
+                offline: Arc::new(AtomicBool::new(false)),
+                lookups: Arc::new(AtomicUsize::new(0)),
             }
         }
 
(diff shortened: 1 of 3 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/tunnel/gateway_provider/gateway_cache.rs
```diff
@@ -40,13 +46,17 @@ pub mod tests {
         /// Artificial delay applied to every `lookup_gateways` call, used to
         /// model a fresh selection that hasn't landed a value in the stream yet.
         lookup_delay: Duration,
+        offline: Arc<AtomicBool>,
```
pronebird, line 49: This seems wrong to me considering that we have:
A: Tokio runtime where we can tokio::select many futures
B: we have connectivity handle somewhere which we can use to poll for updates

nil0ft: Thank you for the feedback. I am not familiar with the codebase, so I will look into it by tomorrow

neacsu: Thank you for looking into this issue @nil0ft . I'm currently reviewing your changes and adding some modifications so that this problem is fixed. Instead of using timers, I think we can re-use the connectivity handle we already have to figure out better moments for retry.
I will come back shortly with a commit.

neacsu:
This was indeed caused by caching offline errors that eventually needed to be popped out one by one, 10 in total, over the course of the 10 retries of the state machine.
However, the offline errors were produced in a corner case where the daemon was started with no internet connection and `Connect` was issued, and only afterwards the internet would become available, thus giving time for the queue to be filled with the errors from an empty gateway cache.
Since https://github.com/nymtech/nym-vpn-client/pull/6015 was merged to `develop`, that corner case no longer exists because the gateway cache is filled right from the start with a snapshot, even if no internet is available. So it no longer produces any offline errors that could be queued. This is not part of the Vanil Noir release, but will become available in the next Wildstrubel release.

TL;DR https://github.com/nymtech/nym-vpn-client/pull/6015 fixes the issue, in a different part (no more gateway caching errors)

PR #6180, opened 2026-08-23
fix(android): restore offline "No internet" status and one-press disconnect

rachyandco:
The v3.5.0 main-screen rewrite collapsed ConnectionState.Offline and WaitingForConnection into the disconnected "not protected" branch, so losing connectivity while connected no longer surfaced an offline state. The backend still detects offline and emits Tunnel.State.Offline correctly; this was purely a UI regression.

- ConnectionStatus: render Offline/WaitingForConnection as "No internet" with the red status color, and restore the pulsing red indicator dot (recreate the Pulse animation removed in the rewrite).
- ActionButton: show "Disconnect" for WaitingForConnection instead of "Connect"; the resolver only produces this state while a tunnel session is active, so disconnect is always the correct action.
- offline_state: a user-initiated Disconnect while offline now tears the tunnel down (DisconnectedState) instead of only disarming auto-reconnect and re-emitting Offline. The stale Offline event raced the client's optimistic Down and forced a second press to disconnect. Gated to Android to preserve desktop kill-switch teardown behavior.

## Ticket

JIRA-NYM-XXXX

## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6180)

### DIFF
--- a/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/service/tile/VpnQuickTile.kt
+++ b/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/service/tile/VpnQuickTile.kt
@@ -90,7 +90,7 @@ class VpnQuickTile :
 				setActive()
 			}
 
-			Tunnel.State.Offline -> {
+			is Tunnel.State.Offline -> {
 				setTileDescription(this@VpnQuickTile.getString(R.string.offline))
 				setActive()
 			}
--- a/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/ui/common/animations/Pulse.kt
+++ b/nym-vpn-android/app/src/main/java/net/nymtech/nymvpn/ui/common/animations/Pulse.kt
@@ -1,0 +1,96 @@
+package net.nymtech.nymvpn.ui.common.animations
+
+import androidx.compose.animation.core.InfiniteRepeatableSpec
+import androidx.compose.animation.core.RepeatMode.Restart
+import androidx.compose.animation.core.StartOffset
+import androidx.compose.animation.core.animateFloat
+import androidx.compose.animation.core.rememberInfiniteTransition
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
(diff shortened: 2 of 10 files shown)


agentpietrucha:
## 🟠 "Disconnect" button shows up when nothing is connected

With this change, the app can show a **Disconnect** button as the main action even though no tunnel session exists — for example when you simply open the app in airplane mode. The user never connected, yet the primary button invites them to disconnect. Pressing it is harmless (the state self-corrects to "No internet" + Connect), but the main CTA is wrong in a fairly common situation, and before this PR these scenarios showed Connect/Get started.

The root cause: the PR assumes `WaitingForConnection` only occurs while a tunnel session is active, but the backend also reports `Offline` when the device has no internet while fully disconnected — and the UI can't tell the two apart.

### How to reproduce

**Scenario A — open the app while offline:**
1. Make sure the VPN is disconnected.
2. Enable airplane mode (or disable Wi-Fi + mobile data).
3. Kill and reopen the app.
4. ❌ The main button reads **Disconnect** (expected: Connect / Get started). The status correctly says "No internet".

**Scenario B — lose internet while disconnected:**
1. Open the app, connect, then disconnect (so the core service is initialized and stays alive).
2. With the app still open and the VPN disconnected, enable airplane mode.
3. ❌ The main button flips to **Disconnect**.

In both cases, pressing Disconnect makes the backend transition `Offline → Disconnected`, after which the UI recovers and shows Connect.

### Technical details

- The Rust state machine runs whenever the service is bound, not only during a session: `VpnService.onBind` → `ensureReadyForManagementBestEffort` → `NymVpnService.newService` spawns it.
- It enters `Offline { reconnect: false }` **with no session** in two places:
  - at startup when the device is offline (`tunnel_state_machine/mod.rs:1244`),
  - from `DisconnectedState` when connectivity drops while the VPN is down (`states/disconnected_state.rs:124`).
- On the Kotlin side, the `reconnect` flag — the only thing distinguishing "session armed, waiting for network" from "nothing connected" — is discarded: `asTunnelState()` maps every `TunnelState.Offline` to the parameterless `Tunnel.State.Offline` (`LibExtensions.kt:21`).
- `resolveConnectionState` (`MainViewModel.kt`) then maps any non-Down/non-Error tunnel state with `NetworkStatus.Disconnected` to `WaitingForConnection`, and `ConnectionState.from` (`ConnectionState.kt:34`) maps `Tunnel.State.Offline` to `WaitingForConnection` unconditionally — so `Offline(reconnect=false)` always renders the new Disconnect branch of `ActionButton`.

**Suggested fix:** `TunnelState::Offline { reconnect }` already crosses the uniffi boundary. Propagate it (`Tunnel.State.Offline(reconnect: Boolean)`) and only produce `WaitingForConnection` (Disconnect button) when `reconnect == true`; map `reconnect == false` to the existing `ConnectionState.Offline` (Connect button, keeping the "No internet" label + pulse).

agentpietrucha (approved):
change looks good now. Let's test on a device and then good to go

rachyandco:
tested successfully on nokia stock android 14

PR #6186, opened 2026-08-24
LAN bypass was dropping the exit metadata address.

tommyv1987:
## Description
Alow_lan=true removes 10.0.0.0/8 from VpnService. Exit metadata is http://10.1.0.1:51830, so that packet leaves on Wi-Fi. Handshake still succeeds; HTTP never returns; we log Metadata endpoints not reachable and never reach Connected.

Samsung dump (LAN on, QUIC off): handshake -500ms, then starting new connection Some("10.1.0.1"), then 10s later Metadata endpoints not reachable. Four retries, zero Connected. Same phone with this keep-route reached Connected at 11:33:46 after debug apk was deployed to device in question.

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6186)

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib-uniffi/src/tunnel_provider/tunnel_settings.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib-uniffi/src/tunnel_provider/tunnel_settings.rs
@@ -313,8 +313,8 @@ impl TunnelNetworkSettings {
             for network in nym_firewall_config::keep_on_tunnel_after_lan_bypass(interface_addrs) {
                 match network {
                     IpNetwork::V4(address) => {
-                        if let Ok(net) = Ipv4Net::new(address.ip(), address.prefix())
-                            .inspect_err(|e| {
+                        if let Ok(net) =
+                            Ipv4Net::new(address.ip(), address.prefix()).inspect_err(|e| {
                                 tracing::error!(
                                     "Failed to re-add IPv4 keep-on-tunnel route for {}: {}",
                                     address,
@@ -326,8 +326,8 @@ impl TunnelNetworkSettings {
                         }
                     }
                     IpNetwork::V6(address) => {
-                        if let Ok(net) = Ipv6Net::new(address.ip(), address.prefix())
-                            .inspect_err(|e| {
+                        if let Ok(net) =
+                            Ipv6Net::new(address.ip(), address.prefix()).inspect_err(|e| {
                                 tracing::error!(
                                     "Failed to re-add IPv6 keep-on-tunnel route for {}: {}",
                                     address,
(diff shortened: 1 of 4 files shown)


nym-vpn-core/crates/nym-vpn-lib-uniffi/src/tunnel_provider/tunnel_settings.rs
```diff
@@ -263,27 +263,83 @@ impl TunnelNetworkSettings {
                 .chain(ALLOWED_LAN_MULTICAST_NETS.iter())
             {
                 match network {
-                    IpNetwork::V4(address) => match Ipv4Net::new(address.ip(), address.prefix()) {
```
neacsu, line 266: All this refactoring work is not really useful and just makes reviewing more difficult. Any aesthetic changes should be in different PRs, if really needed, and not pointed to release branches.

nym-vpn-core/crates/nym-vpn-lib-uniffi/src/tunnel_provider/tunnel_settings.rs
```diff
@@ -263,27 +263,83 @@ impl TunnelNetworkSettings {
...
+                        if let Ok(net) =
+                            Ipv4Net::new(address.ip(), address.prefix()).inspect_err(|e| {
+                                tracing::error!(
+                                    "Failed to re-add IPv4 keep-on-tunnel route for {}: {}",
+                                    address,
+                                    e
+                                )
+                            })
+                        {
+                            tunnel_ipv4.add(net);
```
neacsu, line 293: This seems like a lot of code for basically adding two hard-coded, single IP networks to these two lists. The other part of the `flat_map`ped list seems to be the network interface, which is not used anywhere afaik

nym-vpn-core/crates/nym-vpn-lib-uniffi/src/tunnel_provider/tunnel_settings.rs
```diff
@@ -263,27 +263,83 @@ impl TunnelNetworkSettings {
...
             }
 
             tunnel_ipv4 = tunnel_ipv4.exclude(&exclude_ipv4_lan);
             tunnel_ipv6 = tunnel_ipv6.exclude(&exclude_ipv6_lan);
+
+            // WG metadata (10.1.0.1) lives inside the RFC1918 LAN carve-out.
+            let interface_addrs = self
+                .ipv4_settings
+                .as_ref()
+                .into_iter()
```
neacsu: This is so strange, calling an iterator over an `Option` and than `flat_map`ping. I kind of see what it does in the end, but it seems like over-engineering it and it's very hard to read.

neacsu (changes requested):
I'm having a hard time understanding why this change fixes something which seemed to be working previously. Also, what is the difference between stock and degoogled here that makes any difference?
Empirical evidence should only be providing hints towards understanding the solution, but we can't have a PR code solution without understanding why exactly something is failing.

neacsu (approved):

PR #6233, opened 2026-08-28
fix(core): survive system sleep during reconnect and auto-recover from error state

rachyandco:
Repeated macOS sleep/wake cycles during a reconnect burned the entire retry budget, blacklisted healthy gateways, and left the daemon in Error(ConnectionAttemptsExceeded) with the blocking firewall engaged until manual intervention.

- Detect sleep via wall-vs-monotonic clock divergence (SleepMarker)
- Don't count sleep-interrupted attempts towards the reconnect limit
- Don't blacklist gateways when sleep caused the failure
- Auto-recover from ConnectionAttemptsExceeded after 60s

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6233)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VPN reconnection handling after device sleep or wake events.
  * Sleep-interrupted connection attempts no longer consume retry limits or unnecessarily blacklist gateways.
  * Gateway selection errors now distinguish invalid pinned identities from temporarily unavailable gateways.

* **Improvements**
  * Connection attempts continue retrying instead of stopping after a fixed number of failures.
  * Gateway blacklist entries now include the reason for the failure.
  * Selected gateways are retained when a connection attempt is interrupted by system sleep.

### DIFF
--- a/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/mod.rs
+++ b/nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/mod.rs
@@ -11,6 +11,7 @@ mod ipv6_availability;
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 mod route_handler;
 mod states;
+mod system_sleep;
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 mod tun_ipv6;
 #[cfg(any(target_os = "ios", target_os = "android"))]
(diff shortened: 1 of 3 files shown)


nym-vpn-core/crates/nym-vpn-lib/src/tunnel_state_machine/states/error_state.rs
```diff
@@ -49,7 +54,18 @@ const BLOCKING_INTERFACE_ADDRS: [IpAddr; 2] = [
     )),
 ];
 
+/// Delay before automatically attempting to recover from a transient error.
+const AUTO_RECOVERY_DELAY: Duration = Duration::from_secs(60);
+
+/// Returns true for error reasons caused by transient network conditions, from which
+/// the state machine should automatically attempt to recover instead of staying blocked
+/// until user interaction.
+fn is_auto_recoverable(reason: &ErrorStateReason) -> bool {
```
pronebird: It was designed in a way that recoverable errors kept state machine in connecting state. Irrecoverable errors took it into error state. 

- If the problem is in delay, then I suppose tweaking the delay in connecting state is a simpler way to go about it. 
- If the problem is in UI presentation, then extending connecting state metadata to feed more useful information into UI would be simpler too

PR #6247, opened 2026-08-31
Split disconnected home CTA labels

tommyv1987:
## Ticket

Home connect label should match account state instead of always saying Connect or Get started.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6247)

## Summary by CodeRabbit

- **New Features**
  - Connect buttons now reflect account status, including missing accounts, unavailable accounts, and account-summary checks.
  - Added clearer actions for getting started, choosing a plan, connecting, and retrying when an account cannot be reached.
  - Account-summary refreshes now provide progress feedback and a retry notification when unsuccessful.
  - Added support for opening the Welcome drawer through coordinator actions.
- **Bug Fixes**
  - Menu bar connection status now updates when account, credential, or tunnel information changes.
  - Updated labels and localized messaging for account-related connection states.

### DIFF
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
@@ -319,6 +319,8 @@ private extension OneClickView {
             "purchasePlan.chooseMyPlan".localizedString
         case .accountUnreachable:
             "home.accountUnreachable".localizedString
+        case .checkingAccount:
+            "requestingZkNyms".localizedString
         }
     }
 
@@ -329,7 +331,7 @@ private extension OneClickView {
         switch viewModel.connectState {
         case .disconnected, .noAccount, .noSubscription, .accountUnreachable:
             return .primary
-        case .connecting, .disconnecting, .noInternet:
+        case .connecting, .disconnecting, .noInternet, .checkingAccount:
             return .connecting
         case .stop:
             return .destructive
@@ -343,7 +345,7 @@ private extension OneClickView {
             return true
         }
         switch viewModel.connectState {
-        case .connecting, .disconnecting, .noInternet:
+        case .connecting, .disconnecting, .noInternet, .checkingAccount:
             return true
         case .disconnected, .stop, .connected, .noAccount, .noSubscription, .accountUnreachable:
             return false
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
@@ -313,14 +313,18 @@ private extension OneClickView {
             "disconnecting".localizedString
         case .noInternet:
             "offline".localizedString
-        case .noSubscription:
+        case .noAccount:
             "home.getStarted".localizedString
+        case .noSubscription:
+            "purchasePlan.chooseMyPlan".localizedString
+        case .accountUnreachable:
+            "home.accountUnreachable".localizedString
         }
     }
 
     var connectButtonStyle: NymButton.Style {
         switch viewModel.connectState {
-        case .disconnected, .noSubscription:
+        case .disconnected, .noAccount, .noSubscription, .accountUnreachable:
             .primary
         case .connecting, .disconnecting, .noInternet:
             .connecting
@@ -335,7 +339,7 @@ private extension OneClickView {
         switch viewModel.connectState {
         case .connecting, .disconnecting, .noInternet:
             true
-        case .disconnected, .stop, .connected, .noSubscription:
+        case .disconnected, .stop, .connected, .noAccount, .noSubscription, .accountUnreachable:
             false
         }
     }
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
@@ -323,6 +323,9 @@ private extension OneClickView {
     }
 
     var connectButtonStyle: NymButton.Style {
+        if viewModel.isRefreshingAccountSummary {
+            return .connecting
+        }
         switch viewModel.connectState {
         case .disconnected, .noAccount, .noSubscription, .accountUnreachable:
             .primary
@@ -336,6 +339,9 @@ private extension OneClickView {
     }
 
     var connectButtonDisabled: Bool {
+        if viewModel.isRefreshingAccountSummary {
+            return true
+        }
         switch viewModel.connectState {
         case .connecting, .disconnecting, .noInternet:
             true
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
@@ -328,13 +328,13 @@ private extension OneClickView {
         }
         switch viewModel.connectState {
         case .disconnected, .noAccount, .noSubscription, .accountUnreachable:
-            .primary
+            return .primary
         case .connecting, .disconnecting, .noInternet:
-            .connecting
+            return .connecting
         case .stop:
-            .destructive
+            return .destructive
         case .connected:
-            .connected
+            return .connected
         }
     }
 
@@ -344,9 +344,9 @@ private extension OneClickView {
         }
         switch viewModel.connectState {
         case .connecting, .disconnecting, .noInternet:
-            true
+            return true
         case .disconnected, .stop, .connected, .noAccount, .noSubscription, .accountUnreachable:
-            false
+            return false
         }
     }
 
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
@@ -13,6 +13,7 @@ import ImpactGenerator
 import NetworkMonitor
 import TunnelStatus
 import UIComponents
+import Theme
 #if os(macOS)
 import GRPCManager
 #endif
@@ -51,8 +52,10 @@ public final class OneClickViewModel {
 
     @ObservationIgnored var connectDisconnectTask: Task<Void, Never>?
     @ObservationIgnored var resolveTask: Task<Void, Never>?
+    @ObservationIgnored var accountSummaryRefreshTask: Task<Void, Never>?
     @ObservationIgnored var cancellables = Set<AnyCancellable>()
     @ObservationIgnored var isConnectDisconnectInFlight = false
+    var isRefreshingAccountSummary = false
 
 #if os(iOS)
     public init(
@@ -111,7 +114,7 @@ public final class OneClickViewModel {
 #endif
 
     func connectButtonTapped() {
-        guard !isConnectDisconnectInFlight else { return }
+        guard !isConnectDisconnectInFlight, !isRefreshingAccountSummary else { return }
         guard connectionManager.currentTunnelStatus != .disconnecting else { return }
 
         impactGenerator.impact()
@@ -138,13 +141,42 @@ public final class OneClickViewModel {
             sessionCoordinator?.handle(.requestInactiveSubscriptionPurchase)
             return true
         case .accountUnreachable:
-            Task { @MainActor [weak self] in
-                await self?.credentialsManager.updateAccountSummary(force: true)
-            }
+            startAccountUnreachableRefresh()
             return true
         case .disconnected, .connecting, .stop, .connected, .disconnecting, .noInternet:
             return false
         }
+    }
+
+    func startAccountUnreachableRefresh() {
+        guard !isRefreshingAccountSummary else { return }
+        isRefreshingAccountSummary = true
+        accountSummaryRefreshTask = Task { @MainActor [weak self] in
+            guard let self else { return }
+            defer {
+                isRefreshingAccountSummary = false
+                accountSummaryRefreshTask = nil
+            }
+            await credentialsManager.updateAccountSummary(force: true)
+            if credentialsManager.accountSummaryLastFetchFailed {
+                presentAccountUnreachableRetryFailed()
+            }
+        }
+    }
+
+    func presentAccountUnreachableRetryFailed() {
+        snackbarManager.enqueue(
+            SnackbarItem(
+                style: .critical,
+                title: "home.accountUnreachable".localizedString,
+                message: "error.unexpected".localizedString,
+                actionTitle: "retry".localizedString,
+                onAction: { [weak self] in
+                    self?.snackbarManager.clear()
+                    _ = self?.handleDisconnectedHomeCTATap()
+                }
+            )
+        )
     }
 
     func disconnectFromError() {
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
@@ -117,11 +117,33 @@ public final class OneClickViewModel {
         impactGenerator.impact()
         snackbarManager.clear()
 
+        if handleDisconnectedHomeCTATap() {
+            return
+        }
+
         let isConnectingTap = connectionManager.currentTunnelStatus != .connected
 
         connectDisconnectTask?.cancel()
         connectDisconnectTask = Task { @MainActor [weak self] in
             await self?.performConnectDisconnect(isConnectingTap: isConnectingTap)
+        }
+    }
+
+    func handleDisconnectedHomeCTATap() -> Bool {
+        switch connectState {
+        case .noAccount:
+            sessionCoordinator?.handle(.requestWelcome)
+            return true
+        case .noSubscription:
+            sessionCoordinator?.handle(.requestInactiveSubscriptionPurchase)
+            return true
+        case .accountUnreachable:
+            Task { @MainActor [weak self] in
+                await self?.credentialsManager.updateAccountSummary(force: true)
+            }
+            return true
+        case .disconnected, .connecting, .stop, .connected, .disconnecting, .noInternet:
+            return false
         }
     }
 
@@ -208,7 +230,6 @@ public final class OneClickViewModel {
             }
         }
     }
-
 }
 
 extension OneClickSpeedMode {
--- a/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
+++ b/nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
@@ -143,6 +143,8 @@ public final class OneClickViewModel {
         case .accountUnreachable:
             startAccountUnreachableRefresh()
             return true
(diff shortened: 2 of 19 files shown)


nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
```diff
@@ -313,30 +313,40 @@ private extension OneClickView {
...
+            "purchasePlan.chooseMyPlan".localizedString
+        case .accountUnreachable:
+            "home.accountUnreachable".localizedString
         }
     }
 
     var connectButtonStyle: NymButton.Style {
+        if viewModel.isRefreshingAccountSummary {
+            return .connecting
+        }
```
agentpietrucha, line 330: should we display "Connecting" when the account summary is being refreshed? I don't think this is correct. 
Also, does the `viewModel.isRefreshingAccountSummary` evaluate to `true` when the account is being refreshed in the background? Resulting in "Connecting" button message even when the tunnel is connected?

agentpietrucha: nevermind, I found the `isRefreshingAccountSummary` when I read through the rest of the diff

nym-vpn-apple/Services/Sources/AccountPrefetchGates/DisconnectedHomeCTA.swift
```diff
@@ -0,0 +1,27 @@
...
+        if !isCredentialImported {
+            return .getStarted
+        }
+        if accountSummaryLastFetchFailed, !isAccountActive {
+            return .accountUnreachable
+        }
+        if !isAccountActive, hasAccountSummary {
+            return .choosePlan
+        }
+        return .connect
```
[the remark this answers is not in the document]
agentpietrucha: @tommyv1987 this is nice comment, let's implement this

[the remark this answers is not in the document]
tommyv1987: Sure thing, I will do this shortly.

nym-vpn-apple/Services/Sources/AccountPrefetchGates/DisconnectedHomeCTA.swift
```diff
@@ -0,0 +1,28 @@
...
+        hasAccountSummary: Bool
+    ) -> Self {
+        if !isCredentialImported {
+            return .getStarted
+        }
+        if accountSummaryLastFetchFailed, !isAccountActive {
+            return .accountUnreachable
+        }
+        if !isAccountActive {
+            return hasAccountSummary ? .choosePlan : .checking
```
agentpietrucha, line 24: **`.checking` is a terminal state with no way out. This is a regression.**

`.checking` is returned whenever a credential is imported, the account is not active, and **no fetch failure was recorded**. Once the UI is in the resulting `.checkingAccount`:

- the button is disabled - `OneClickView.connectButtonDisabled`
- the tap is swallowed - `handleDisconnectedHomeCTATap` returns `true` and does nothing
- the macOS menu bar renders it as non-actionable `Text` - `ConnectButtonState.menuBarItemIsAction`

and nothing in this PR ever triggers the account-summary fetch the state is waiting for. `.accountUnreachable` got a tap-to-retry; `.checking` got no path out. Recovery depends on an unrelated trigger - `handleSceneBecameActive` (a foreground transition) or the post-disconnect timer - neither of which fires while a macOS user sits in an already-foregrounded window.

Before this PR the same inputs produced `.noSubscription` with an **enabled** CTA, and the tap ran `passesConnectPreflight()` (`OneClickViewModel+State.swift`, line 199), which itself called `updateAccountSummary()` and then either offered purchase or connected. That self-healing escape hatch is gone.

**Reachable when `accountSummary == nil` and `isCredentialImported` and `!accountSummaryLastFetchFailed`:**

- `CredentialsManager.swift` line 139 - `runAccountRegistration` nils the summary *before* `login()`, so any failure after the credential lands leaves it nil with the flag clear.
- `CredentialsManager.swift` lines 819-840 - `handleEnvironmentDidChange` nils the summary **and clears the failure flag**, then calls `try await ensureAccountRegisteredForCurrentEnvironment()`. On throw it falls into `catch`, logs, and never reaches `updateAccountSummary(force: true)`.

**Repro (deterministic, iOS Santa build):**

1. iOS TestFlight or debug build with `SANTA` defined and Santa's menu enabled (`isSantaClaus`).
2. Sign in, let the account summary land, confirm home shows Connect.
3. Enable airplane mode (or point at an unreachable environment).
4. Santa's menu, switch VPN environment. `handleEnvironmentDidChange` nils the summary, clears the flag, and `ensureAccountRegisteredForCurrentEnvironment()` throws.
5. Home CTA is now "Requesting ZkNyms...", greyed out, unresponsive. Restore the network - nothing changes. Tapping does nothing. Only backgrounding and refocusing the app recovers it.

**Suggested fix:** trigger the refresh when this state is entered, rather than only offering it for `.accountUnreachable`. Either kick `updateAccountSummary(force: true)` off from `recomputeConnectState()` when the resolved CTA is `.checking`, or at minimum make the tap do it (see my comment on `OneClickViewModel.swift`) and remove `.checkingAccount` from `connectButtonDisabled` so the tap can reach the view model at all.

nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel.swift
```diff
@@ -125,6 +132,55 @@ public final class OneClickViewModel {
...
+        case .noAccount:
+            sessionCoordinator?.handle(.requestWelcome)
+            return true
+        case .noSubscription:
+            sessionCoordinator?.handle(.requestInactiveSubscriptionPurchase)
+            return true
+        case .accountUnreachable:
+            startAccountUnreachableRefresh()
+            return true
+        case .checkingAccount:
```
agentpietrucha, line 146: This is the swallow that makes `.checkingAccount` a dead end (see my comment on `DisconnectedHomeCTA.swift`). Returning `true` reports "handled" to `connectButtonTapped`, which then returns early - so the tap does nothing at all, and there is no other affordance to retry.

Suggested fix - give it the same self-service `.accountUnreachable` has:

```swift
case .checkingAccount:
    startAccountSummaryRefresh()   // renamed from startAccountUnreachableRefresh()
    return true
```

The existing method already guards re-entry via `isRefreshingAccountSummary`, sets the in-flight flag, and surfaces the retry snackbar on failure, so it needs no changes beyond the rename. Pair this with dropping `.checkingAccount` from `connectButtonDisabled`, otherwise the button is (intended to be) untappable and the new branch is unreachable.

nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
```diff
@@ -313,30 +313,42 @@ private extension OneClickView {
...
         case .noInternet:
             "offline".localizedString
-        case .noSubscription:
+        case .noAccount:
             "home.getStarted".localizedString
+        case .noSubscription:
+            "purchasePlan.chooseMyPlan".localizedString
+        case .accountUnreachable:
+            "home.accountUnreachable".localizedString
+        case .checkingAccount:
```
agentpietrucha, line 322: **Wrong copy on the primary CTA.**

`requestingZkNyms` resolves to "Requesting ZkNyms...". Its only other call site is `SettingsViewModel.swift` line 328, for the actual zk-nym prefetch - a different backend operation. So the primary home CTA reads "Requesting ZkNyms..." while the app is only checking account entitlement.

For a ticket whose entire scope is "home connect label should match account state" (NYM-1764), this label describes the wrong state, in internal jargon users will not recognise.

Suggested fix: add a dedicated key, for example `home.checkingAccount` -> "Checking account...", and use it here and at `ConnectButtonState.swift` line 82.

nym-vpn-apple/Home/Sources/26/OneClick/OneClickView.swift
```diff
@@ -313,30 +313,42 @@ private extension OneClickView {
...
+            "requestingZkNyms".localizedString
         }
     }
 
     var connectButtonStyle: NymButton.Style {
+        if viewModel.isRefreshingAccountSummary {
```
agentpietrucha, line 328: **Refresh-flag check sits ahead of the state switch, so it can mask connected states.**

Following up on the thread you and @agentpietrucha had here. The withdrawal is correct about the case raised: `isRefreshingAccountSummary` is only ever set by `startAccountUnreachableRefresh()`, never by the scene-foreground or post-disconnect refreshes, so a background refresh cannot cause this.

But the ordering concern survives in a narrower form. Because the flag is checked *before* the `connectState` switch here and in `connectButtonDisabled`, if the tunnel comes up during a user-initiated retry the button reads "Connected" while being disabled and styled `.connecting` - so **disconnect is unavailable**. iOS always-on / on-demand can start the tunnel autonomously (`TunnelsManager+Extensions.swift` line 25), and the window is not short: `AccountSummaryRefreshPolicy.pollDelays` sums to 26 seconds on iOS, and the macOS retry loop to about 22 seconds. There is no cancellation path.

Suggested fix: scope the check to the states it is meant to cover, rather than short-circuiting the whole switch.

```swift
var connectButtonStyle: NymButton.Style {
    switch viewModel.connectState {
    case .accountUnreachable, .checkingAccount:
        return viewModel.isRefreshingAccountSummary ? .connecting : .primary
    // ...
    }
}
```

Separately: while refreshing, the label stays "Can't reach account" and only the colour changes. Swapping the label to the checking copy for the duration would make the "progress feedback" in the PR description actually legible.

nym-vpn-apple/NymVPNDaemon/NymVPNDaemonApp.swift
```diff
@@ -220,13 +220,29 @@ private extension NymVPNDaemonApp {
...
-            )
+            refreshMenuBarConnectButtonState()
+        }
+        .onChange(of: appSettings.isCredentialImportedPublisher) { _, _ in
+            refreshMenuBarConnectButtonState()
+        }
+        .onChange(of: credentialsManager.accountSummaryLastFetchFailed) { _, _ in
+            refreshMenuBarConnectButtonState()
         }
+        .onChange(of: credentialsManager.accountSummary?.isActive) { _, _ in
```
agentpietrucha, line 233: **This trigger misses changes that the consumer actually cares about.**

`refreshMenuBarConnectButtonState()` passes `credentialsManager.isAccountActive()`, which treats a **future `validUntilDate` as active even when `isActive == false`** (`CredentialsManager.swift` lines 495-505). But this observes `accountSummary?.isActive`.

So a refresh that moves `validUntilDate` across "now" while leaving `isActive` unchanged fires **none** of the four `onChange` handlers, and the menu bar goes stale: still offering "Connect" for an account that just lapsed, or still showing "Choose my NymVPN plan" for one that just got extended.

The home screen does not have this gap - it observes `$accountSummary` directly, and since `AccountSummary` is `Codable` but not `Equatable`, every assignment republishes. I assume that non-`Equatable`ness is exactly why `?.isActive` was reached for here.

Suggested fix - observe the derived value you actually consume:

```swift
.onChange(of: credentialsManager.isAccountActive()) { _, _ in
    refreshMenuBarConnectButtonState()
}
```

It is a `Bool`, so `onChange` is satisfied, and it collapses the `isActive`, `validUntil` and no-summary cases into a single trigger.

nym-vpn-apple/UIComponents/Sources/UIComponents/26/Buttons/ConnectButton/ConnectButtonState.swift
```diff
@@ -48,16 +73,22 @@ public enum ConnectButtonState {
...
         case .installingDaemon:
             "home.installDaemonButton".localizedString
-        case .noAccount, .noSubscription:
+        case .noAccount:
             "home.getStarted".localizedString
+        case .noSubscription:
+            "purchasePlan.chooseMyPlan".localizedString
+        case .accountUnreachable:
+            "home.accountUnreachable".localizedString
+        case .checkingAccount:
```
agentpietrucha, line 82: Second call site for the "Requesting ZkNyms..." copy problem - see my comment on `OneClickView.swift` line 322. This one surfaces it in the macOS menu bar.

Separate, lower-priority note on the same block: the three new or newly-used keys here are English-only. `home.accountUnreachable`, `purchasePlan.chooseMyPlan` and `requestingZkNyms` each have only an `en` localization, whereas `home.getStarted` - which `.noSubscription` rendered before this PR - has 16. So a non-English user with an inactive subscription regresses from a translated CTA to English text. 429 of 712 keys in the catalog are English-only, so this looks like the normal English-first flow rather than an oversight - flagging it so it is a decision rather than a surprise at release.

nym-vpn-apple/UIComponents/Sources/UIComponents/26/Buttons/ConnectButton/ConnectButtonState.swift
```diff
@@ -69,9 +100,9 @@ public enum ConnectButtonState {
 extension ConnectButtonState {
     public var menuBarItemIsAction: Bool {
         switch self {
-        case .connect, .disconnect, .stop, .noInternetReconnect, .noInternet, .noAccount, .noSubscription:
+        case .connect, .disconnect, .stop, .noInternetReconnect, .noInternet:
             true
-        case .disconnecting, .installingDaemon:
+        case .disconnecting, .installingDaemon, .noAccount, .noSubscription, .accountUnreachable, .checkingAccount:
```
agentpietrucha, line 105: Moving `.noAccount` and `.noSubscription` out of `menuBarItemIsAction` is the right call - the old button could only call `connectDisconnect()`, which would fail for both states. But the result is now an un-clickable "Get started" or "Choose my NymVPN plan" label in the menu bar, with no affordance at all: a menu-bar-only user (`appMode == .menubarOnly`) sees a plan CTA they cannot act on.

Worth considering routing these two to something actionable, for example a "Open NymVPN" item that brings the window forward, rather than a dead label.

nym-vpn-apple/Home/Tests/HomeTests/ProcessingAccountViewModelTests.swift
```diff
@@ -25,6 +25,7 @@ final class FakeProcessing: AccountProcessing {
     var storeDeeplinkError: Error?
     var registerError: Error?
     var accountActive = true
+    var becomesActiveAfterSync = false
```
agentpietrucha, line 28: `becomesActiveAfterSync` is never set to `true` by any test - it is only declared here and read at line 61, so the branch it guards is dead. Along with the `#expect(coordinator.actions == ...)` addition further down, this whole file's changes look unrelated to the CTA work and are probably rebase residue. Suggest dropping them from this PR unless they are load-bearing for something I have missed.

nym-vpn-apple/Home/Tests/HomeTests/OneClickViewModelCTATapTests.swift
```diff
@@ -0,0 +1,135 @@
...
+        snackbarManager = SnackbarManager()
+    }
+
+    override func tearDown() {
+        snackbarManager.clear()
+        snackbarManager = nil
+        super.tearDown()
+    }
+
+    private func makeViewModel() -> OneClickViewModel {
```
agentpietrucha, line 36: These view models are built on the real `.shared` singletons, and `testAccountUnreachableTapStartsRefreshAndIgnoresReentry` then awaits the real `credentialsManager.updateAccountSummary(force: true)`.

It only passes quickly because `isValidCredentialImported` reads `false` in the test host, so `updateAccountSummary` returns at its guard. If a previously-run test, a leftover `credenitalExists` value in the host's `UserDefaults`, or a `-MOCK_MODE` launch argument makes that `true`, the test performs real account-store or gRPC work and picks up the poll-delay sleeps - 26 seconds on iOS, about 22 on macOS - or hangs.

Injecting a `CredentialsManager` double (or at least asserting the guard precondition in `setUp`) would make these deterministic. The other four tests in this file are pure state assertions and are fine as they are.

agentpietrucha (changes requested):
Reviewed the full diff (18 files, +548/-32) against the merge base. The shape of the change is good: `DisconnectedHomeCTA` as a pure policy next to `ConnectPlanPurchaseGatePolicy` is the right factoring, and I verified the branch ordering preserves the existing rule that a failed summary fetch must **not** offer plan purchase. Both CodeRabbit items are genuinely implemented (`stagePreauthDrawer(.welcome)` plus the pending/commit test, and the no-summary to checking mapping).

Requesting changes for one functional regression and two correctness gaps; details are inline.

**Summary of inline comments**

1. `.checkingAccount` is a terminal state - disabled button, swallowed tap, and nothing ever triggers the fetch it waits for. This is a regression against the previous behaviour and it comes from the no-summary fix. (`DisconnectedHomeCTA.swift`, `OneClickViewModel.swift`)
2. `.checkingAccount` renders "Requesting ZkNyms..." - wrong state, internal jargon, on the primary CTA. (`OneClickView.swift`, `ConnectButtonState.swift`)
3. macOS menu bar goes stale when only `validUntil` changes, because the trigger watches `accountSummary?.isActive` but the consumer reads `isAccountActive()`. (`NymVPNDaemonApp.swift`)

Plus four smaller items inline (refresh-flag ordering, menu-bar dead labels, unused test-double flag, tests on real singletons).

**One item I could not anchor inline** (the file is not in this diff, so no diff hunk to attach to):

`connectButtonDisabled` does not actually disable anything. `NymButton` seeds `isDisabled` into `@State` from its initialiser (`UIComponents/Sources/UIComponents/26/Buttons/NymButton.swift`, lines 79 and 98), so SwiftUI ignores every later value for a stable view identity - and `connectButton` sits at a fixed position in `baseSection`'s `VStack`, so its identity never changes. This is pre-existing and mostly masked here: the `.connecting` style renders identically whether enabled or disabled, and the view-model guards make the tap a no-op. What does leak through is that VoiceOver announces an enabled button and macOS shows a hover highlight on a control you intended to be disabled. Worth knowing because `isRefreshingAccountSummary` was added specifically to disable this button - and because it is the only reason comment 1 is not a hard brick today.

**Verified clean**, for the record:

- No security surface touched: no credential, keychain, token, networking or logging changes.
- Devex is fine: the SPM additions are `.product` deps on packages already declared at package level (I checked every product name against the `Services`, `ServicesMutual` and `Theme` manifests), and there is no `UIComponents -> Services -> UIComponents` cycle. No new env vars, ports, scripts or tooling.
- No feature-flag leaks: no gates touched, `Home/Sources` contains only `26`, and `ConnectButton` has zero call sites, so `ConnectButtonState`'s only live consumer is the macOS menu bar.
- `ConnectButtonState.init`'s new default arguments preserve `.connect` for any two-argument caller.
- `initial: true` on the tunnel-status `onChange` is a real fix - the menu bar icon and CTA were previously stuck at their `@State` defaults until the first status change.
- No leftover conflict markers in any changed file (checked, given the rebase).
- Every localization key referenced by the diff exists in the catalog.

PR #6253, opened 2026-09-01
Unregistered cta honesty

tommyv1987:
## Description

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6253)

## Summary by CodeRabbit

* **New Features**
  * Inactive accounts now proceed directly to plan selection or purchase flows when appropriate, including during login and connection attempts.
  * Account setup can skip the introductory carousel when an account is inactive.
  * Renewal actions now display account-specific titles, such as “Choose Plan” or “Renew Now.”
  * Settings provide clearer status messaging for unavailable accounts, pending registration, and missing plans.

* **Bug Fixes**
  * Improved account summary refresh behavior and retry handling.
  * Corrected connection and menu bar states for inactive accounts without subscription details.


tommyv1987:
@copilot resolve the merge conflicts in this pull request

agentpietrucha (changes requested):
Requesting changes. The root problem is that `accountSummary == nil` does not mean "unregistered", so the two changes keyed off it regress states this PR doesn't need to touch.

Verified in the core: `NymVpnAccountStorage::get_account_summary` (`nym-vpn-core/crates/nym-vpn-lib-uniffi/src/vpn_account_storage.rs:313-355`) returns `Ok(None)` **only when no account is stored locally**. A registered account with no plan returns `Ok(Some(summary))` with `subscription: None` — non-nil summary, `isActive == false`. A network failure with no cache returns `Err`. So the unregistered case this PR targets already resolved to `.choosePlan` under the old `hasAccountSummary` guard; that guard only gated "summary not fetched yet".

### Blocking

**1. Active subscribers get the paywall CTA — `Services/Sources/AccountPrefetchGates/DisconnectedHomeCTA.swift:21`**

Dropping `hasAccountSummary` makes `.choosePlan` fire for `summary == nil`, which on macOS is reachable and is not an error. `ServicesMacOS/.../GRPCManager+Credentials.swift:26` optional-chains the daemon client:

```swift
guard let summary = try await self?.rpcClient?.getAccountSummary() else { return nil }
```

`rpcClient` is nil until `RpcClient()` connects and is reset to nil on disconnect (`GRPCManager.swift:122`). A refresh in that window returns nil **without throwing** → `fetchAccountSummary` assigns nil → the `accountSummary` `didSet` wipes the persisted cache (`AppSettings.swift:166-172`) → `accountSummaryLastFetchFailed` stays `false` → `.choosePlan`. Home button and menu bar (`NymVPNDaemonApp.swift:242`) then read "Choose my plan" for a paying user, and the tap fires `.requestInactiveSubscriptionPurchase`. Triggered by the foreground refresh (`AppFeatureViewModel.swift:276`) or post-disconnect refresh (`:298`) while the daemon is restarting or stopped.

Fix: keep the `hasAccountSummary` guard. It costs nothing, because the unregistered case has a summary.

**2. The retry that covered that window was removed — `CredentialsManager.swift:608-615`**

Old behaviour was `if accountSummary != nil { break }`, so a nil summary retried across `[.zero, 2, 4, 6, 10]s` — which is what let the macOS daemon come up. `shouldFinishSummaryPoll(untilActive: false, …)` returns `hasAccountSummary || !lastFetchFailed`, and a nil-but-successful gRPC read leaves `lastFetchFailed == false`, so the poll now exits after one attempt.

Separately, the doc comment "Failed fetches still retry" isn't achievable in either caller: `CredentialsManager+iOSAccountController.swift:240-256` hardcodes `lastFetchFailed: false` in both branches, and a real iOS failure throws out of the loop rather than iterating. The parameter is dead in every call site.

### Should fix before merge

**3. `passesConnectPreflight()` always returns `false` — `OneClickViewModel+State.swift:208-212`**

The tail runs whenever `isAccountValid()` is false. In the `isAccountRegistrationInFlight` and `accountSummaryLastFetchFailed` branches `shouldOfferPlanPurchaseOnConnect` is false, so nothing happens at all: no purchase sheet, no alert, and `performConnectDisconnect` returns at `guard canProceed else { return }` before `handleInactiveSubscriptionErrorIfNeeded()` and `clearLastErrorIfNeeded()`.

This is reachable on Stop/Cancel taps: `handleDisconnectedHomeCTATap()` returns `false` for `.stop` (tunnelStatus `.error`) and `.connecting`, while `isConnectingTap = status != .connected` is `true`. A user whose tunnel is in `.error` while the summary can't be fetched can no longer press Stop — the button is inert and the error state never clears. The old `return true` fell through to `connectDisconnect()` + `clearLastErrorIfNeeded()`.

Fix: gate the preflight on `currentTunnelStatus == .disconnected`, or return `false` only in the `shouldOfferPurchase` branch.

**4. New "Account unreachable" Settings copy can't be reached live — `SettingsViewModel.swift:276-284, 346-366`**

`nilSummaryAccountCopy` reads `accountSummaryLastFetchFailed` and `isAccountRegistrationInFlight`, but `setupCredentialManagerObservers()` (`:243-262`) only subscribes to `$accountIdentifier` and `$accountSummary`. With the summary staying nil nothing refreshes the section, so the row keeps what `reloadSections()` rendered — normally "No active plan / Choose my plan" — and never flips to "Account unreachable". `isAccountRegistrationInFlight` isn't `@Published` at all (`CredentialsManager.swift:47`), so it can never drive an update.

Fix: merge `$accountSummaryLastFetchFailed` into the observers and publish `isAccountRegistrationInFlight`.

**5. The button-drift guard was weakened instead of fixed — `AccountReportTests.swift:187`**

`AccountAndDevicesView+AccountStatus.swift:173` now renders `renewButtonTitle` (→ `purchasePlan.chooseMyPlan` when inactive), but `ServicesMutual/.../AccountButtons.swift:53` still emits `titleKey: "settings.account.renewNow"` unconditionally. `descriptorMatchesViewButtonSet` exists to catch exactly that divergence; adding `source.contains("renewButtonTitle")` to the scanner makes the sets match again while the descriptor and the generated QA report stay wrong for inactive accounts.

Fix: derive the title in `accountButtons` from `summary.isActive` rather than teaching the scanner a new alias.

### Note

CodeRabbit skipped this PR ("auto reviews disabled on base/target branches other than the default branch"), so it has had no bot coverage — worth triggering `@coderabbitai review` manually. SonarCloud reports 0.0% coverage on new code.

agentpietrucha:
@coderabbitai review

agentpietrucha (changes requested):
Re-reviewed at `21810a8a8`. Points **1**, **2** and **4** from my earlier review are addressed, and CodeRabbit's three inline comments are in — thanks for the quick turnaround.

Requesting changes on one thing, which came out of the fix for my point **3**.

## Blocking — macOS: Connect becomes a silent no-op when the daemon isn't serving

**`nym-vpn-apple/Home/Sources/26/OneClick/OneClickViewModel+State.swift:166`**

```swift
if isConnectingTap, connectionManager.currentTunnelStatus == .disconnected {
```

I suggested gating on `.disconnected`, so the imprecision is on me — but `derivedConnectState()` two functions up folds `.unknown` in with `.disconnected`:

```swift
// :147
case .disconnected, .unknown:
    return .disconnected(DisconnectedHomeCTA.resolve(...))
```

So `.unknown` renders a normal **Connect** CTA whose tap now skips `passesConnectPreflight()` entirely.

`.unknown` is a persistent macOS state, not a startup blip:

- `GRPCManager.tunnelStatus` is initialised to `.unknown` and reset to `.unknown` in `onDisconnect()`, which also nils `rpcClient` (`ServicesMacOS/.../GRPCManager.swift:26`, `:119-122`). It stays there until `onConnect` runs.
- `ConnectionManager+macOS.swift:54-56` mirrors it into `currentTunnelStatus`.
- Nothing refreshes the account summary out of that state, so the cached active summary survives and the CTA stays `.connect`:
  - `handleTunnelStatusChange` has `guard newStatus == .disconnected else { return }` (`AppFeatureViewModel.swift:292`) — `.unknown` never schedules the post-disconnect refresh.
  - `handleSceneBecameActive` computes `shouldBypassThrottle = isPurchaseFlowActive || !isAccountActive`, which is `false` for an active account, so the foreground refresh is throttled (`:264-272`).

The tap then does **nothing at all** — not even an error. `connectDisconnect()` → `connect()` → `startTunnel()` → `grpcManager.connect()`, which is `try await self?.rpcClient?.connectTunnel()` (`GRPCManager+Connection.swift:13-17`): optional-chained through a nil `rpcClient`, so it returns without throwing. It also increments `appSettings.statisticsConnectionCount` for a connection that never happened (`ConnectionManager+macOS.swift:16-17`).

And the user loses the only way out: `passesConnectPreflight():191` is the sole production trigger for `onRequestDaemonEnable` → `SettingLink.daemonEnable` (`AppFeatureView.swift:307`). `ConnectButtonState.installingDaemon` is never constructed anywhere, so `DaemonInstallView` has no other entry point from Home. Before this PR the preflight ran in `.unknown` and routed there.

iOS is unaffected — `TunnelStatus(from: NEVPNStatus)` never produces `.unknown` (`.invalid` maps to `.disconnected`).

### Repro (macOS)

1. Sign in with an account that has an active plan, and connect/disconnect once so the summary is cached.
2. Stop the daemon: `sudo launchctl bootout system/net.nymtech.vpn.daemon` (or kill it so the gRPC client disconnects).
3. Tap **Connect** on Home.
4. **Expected:** navigate to Settings → daemon install/enable. **Actual:** nothing happens, no error, and the button stays live.

### Suggested fix

Make the gate mirror `derivedConnectState()`'s disconnected case instead of matching a single status:

```swift
if isConnectingTap, ConnectPreflightPolicy.runsPreflight(for: connectionManager.currentTunnelStatus) {
```

```swift
// AccountPrefetchGates
public enum ConnectPreflightPolicy {
    /// Mirrors OneClickViewModel.derivedConnectState()'s disconnected case: a fresh
    /// connect attempt. Stop/cancel taps (.error, .connecting, .offline) skip preflight.
    public static func runsPreflight(for status: TunnelStatus) -> Bool {
        status == .disconnected || status == .unknown
    }
}
```

A policy type keeps the two switches from drifting apart again and is directly unit-testable. My other original suggestion — returning `false` only in the `shouldOfferPurchase` branch and `true` otherwise — also works and avoids the coupling entirely. Either is fine.

I have a few smaller non-blocking notes (the weakened `descriptorMatchesViewButtonSet` guard, the actionless pending-summary state in Account & Devices, and a couple of unreachable branches) — I'll leave those out of the change request and post them separately so they don't hold up the fix.

PR #6254, opened 2026-09-01
macOS login checkout session

tommyv1987:
## Ticket

## Checklist:

- [ ] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6254)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plan-purchase transitions so the purchase screen appears only after the drawer has fully closed.
  * Kept the purchase overlay visible throughout checkout transitions for a smoother experience.
  * Login now completes successfully when credentials already exist, without displaying an unnecessary error notification.
  * Improved account-status handling during sign-in and setup, including faster recognition of inactive accounts.
  * Updated account summary polling to retry appropriately and stop when the account reaches a final inactive state.


nym-vpn-apple/ServicesMacOS/Sources/GRPCManager/GRPCManager+Credentials.swift
```diff
@@ -2,12 +2,20 @@ import NymVPNLib
...
 extension GRPCManager {
     public func storeAccount(with request: StoreAccountRequest) async throws {
-        try await Task.detached { [weak self] in
-            try await self?.rpcClient?.storeAccount(request: request)
-        }.value
+        do {
+            try await Task.detached { [weak self] in
+                try await self?.rpcClient?.storeAccount(request: request)
+            }.value
+        } catch let error as VpnError {
```
agentpietrucha: **Blocking — this catch never matches, so the whole feature is inert on macOS.**

`rpcClient.storeAccount` does not throw `VpnError`. In `nym-vpn-core/crates/nym-vpn-lib-uniffi/src/rpc.rs:265-273` it returns `Result<(), RpcError>` (`pub type Result<T, E = RpcError>` at :573), and the generated binding confirms it: `errorHandler: FfiConverterTypeRpcError__as_error_lift`. The codebase already knows this elsewhere — `GRPCManager.swift:54` does `(error as? RpcError)?.displayChain()`.

Failure path: daemon already holds a mnemonic → `AccountCommandError::ExistingAccount` → lifted as `RpcError` → this catch is skipped → `ErrorReason.existingAccount` never thrown → `isExistingAccountStoreError` fails both branches (`as? ErrorReason` is nil; `localizedDescription` is `String(reflecting:)` = `"NymVPNLib.RpcError"`) → `PassphraseSignInViewModel` rethrows and shows a critical snackbar reading `NymVPNLib.RpcError`.

```swift
} catch let error as RpcError {
    if case .existingAccount = error.accountError() {
        throw ErrorReason.existingAccount
    }
    throw error
}
```

`RpcError.accountError() -> AccountCommandError?` is exported for exactly this (`rpc.rs:533-540`).

Separately, while you're in this block: `try await self?.rpcClient?.storeAccount(...)` optional-chains, so a nil `rpcClient` returns success with nothing stored and the caller reports a successful login. Pre-existing, but this function is now the login gate.

nym-vpn-apple/Services/Sources/Services/CredentialsManager/CredentialsManager.swift
```diff
@@ -600,20 +606,36 @@ extension CredentialsManager {
...
-                isSubscriptionActive: accountSummary?.isActive == true,
-                hasAccountSummary: accountSummary != nil,
-                lastFetchFailed: accountSummaryLastFetchFailed
-            ) {
-                break
+        var knownInactive = false
+        if !untilActive {
+            knownInactive = await grpcManager.isAccountKnownInactiveForLogin()
+        }
+        if !knownInactive {
```
agentpietrucha: **Blocking — this skips every `fetchAccountSummary()`, freezing a stale entitlement.**

`accountSummary` is persisted (`AppSettings.swift:149-177`, `@AppStorage` JSON), so it survives relaunches.

Scenario: subscription active at T0 → summary cached with `isActive: true`, `validUntil: T1`. At T1 the sub lapses and the daemon enters `ErrorState(InactiveSubscription)`. Every `updateAccountSummary(force: true, untilActive: false)` now returns without fetching. `isAccountActive()` (:496-510) checks `accountSummary.isActive` **first** and returns `true`, so the expired user reads as subscribed indefinitely, no paywall — and nothing ever re-fetches, because `isAccountValid()` only forces a refresh when `isAccountActive()` is already false.

This is not login-scoped: `untilActive: false` is the default on every caller — `SettingsView.onAppear`, `AccountAndDevicesView` (×3), `OneClickViewModel`, `AppFeatureViewModel` foreground refresh, `isAccountValid()`. Pre-PR all of them fetched on attempt 0.

Suggest dropping the `if !knownInactive` wrapper (and the equivalent iOS early return) and letting the loop run. The in-loop `isAccountKnownInactive:` argument to `shouldFinishSummaryPoll` already terminates the poll after a single attempt — it just does the fetch first, which is what you want.

nym-vpn-apple/Services/Sources/Services/CredentialsManager/CredentialsManager+iOSAccountController.swift
```diff
@@ -233,7 +233,21 @@ extension CredentialsManager {
...
 
-                for delay in AccountSummaryRefreshPolicy.pollDelays(untilActive: untilActive) {
+                var knownInactive = false
+                if !untilActive {
+                    let phase = Self.accountPreparationPhase(
+                        from: await controller.getAccountState()
+                    )
+                    knownInactive = OnboardingAccountPreparationPolicy
+                        .isTerminalInactiveForLogin(phase)
+                    if knownInactive {
```
agentpietrucha: **Blocking — same issue as the macOS path.**

This `return` exits the `withController` closure before the poll loop, so `getAccountSummary()` / `applyVpnAccountSummary()` never run and the persisted summary is left untouched. See the detail on `CredentialsManager.swift:613`.

The in-loop `isAccountKnownInactive:` check below already gives you the RPC saving you're after, without skipping the fetch.

nym-vpn-apple/Services/Sources/Services/CredentialsManager/CredentialsManager+iOSAccountController.swift
```diff
@@ -243,17 +257,34 @@ extension CredentialsManager {
...
+                            knownInactive = OnboardingAccountPreparationPolicy
+                                .isTerminalInactiveForLogin(phase)
+                        }
+                        if AccountSummaryRefreshPolicy.shouldFinishSummaryPoll(
+                            untilActive: untilActive,
+                            isSubscriptionActive: false,
+                            hasAccountSummary: false,
+                            lastFetchFailed: false,
+                            attemptIndex: attemptIndex,
+                            isAccountKnownInactive: knownInactive
```
agentpietrucha, line 280: **The iOS terminal-inactive return marks a false fetch failure and suppresses the checkout CTA.**

When the early return above fires for the exact target case (unregistered account, no summary yet), `accountSummary` stays nil and execution falls through to `if accountSummary == nil { setAccountSummaryLastFetchFailed(true) }` at :293.

That flag flows through `makeSessionEnvironment` (`AppFeatureViewModel.swift:668`) into `DrawerSessionPolicy.shouldOfferPlanPurchaseAfterProcessing`, whose first line for `.login` is `if accountSummaryLastFetchFailed { return false }` — no purchase offered.

This directly contradicts the comment you added in this PR at `OnboardingSessionPolicy.swift:119`: *"Missing summary is inactive (offer purchase) … failed fetch is the offline exception."* The flag conflates "unregistered, no summary" with "offline", so the exception swallows the rule. The `accountSummary == nil` line is pre-existing, but this PR routes a new, intentionally-successful path onto it.

nym-vpn-apple/Home/Sources/26/PassphraseSignIn/PassphraseSignInViewModel.swift
```diff
@@ -75,6 +67,28 @@ public final class PassphraseSignInViewModel {
         }
     }
 
+    private func storeCredentialTreatingExistingAccountAsSuccess(_ credential: String) async throws {
+        do {
+            try await credentialStore.storeLoginCredential(credential)
+        } catch {
+            guard OnboardingSessionPolicy.isExistingAccountStoreError(error) else { throw error }
```
agentpietrucha, line 74: **Any valid mnemonic logs the user into the *previously stored* account.**

The daemon returns `ExistingAccount` without comparing the submitted mnemonic — the payload is discarded in every state that can produce it:

- `state_machine/error_state.rs:81` — `AccountCommand::StoreAccount(return_sender, _) => ...ExistingAccount`
- `pending_subscription_state.rs:67`, `syncing_state/local_state.rs:231`, `syncing_state/network_state.rs:228`, `ready_state.rs:160`

So this reports success for *any* mnemonic typed on a device that already has one stored, and `completeLoginAfterStore()` then emits `.authCompleted(.login)` for whatever account was already there. The comment ("second app window, retry") describes one cause, but the check can't distinguish it from "user typed a different account's passphrase".

Live on iOS today — `account.rs:276-284` validates via `StorableAccount::try_from` first, so it's gated to well-formed mnemonics but not to the *correct* one. On macOS it's latent behind the `VpnError`/`RpcError` bug and goes live the moment that's fixed. A user trying to switch accounts is silently kept on the old one.

Suggest gating the swallow on evidence that the stored account is the one just submitted (e.g. compare `getAccountIdentity()`), or keep surfacing the error with a "log out first" affordance.

nym-vpn-apple/Services/Sources/AccountPrefetchGates/AccountSummaryRefreshPolicy.swift
```diff
@@ -24,17 +24,50 @@ public enum AccountSummaryRefreshPolicy {
...
             return shouldStopUntilActivePoll(isSubscriptionActive: isSubscriptionActive)
         }
-        return hasAccountSummary || !lastFetchFailed
+        if isAccountKnownInactive {
+            return true
+        }
+        if hasAccountSummary, !lastFetchFailed {
+            return true
+        }
+        if lastFetchFailed {
```
agentpietrucha, line 68: **A failed fetch with a cached summary now burns the full 26s ladder.**

Old: `return hasAccountSummary || !lastFetchFailed` → stale summary + failed fetch → finish immediately.
New: the `hasAccountSummary, !lastFetchFailed` branch misses, then this `return false` keeps polling all 7 attempts.

`pollDelays` is now used on macOS too and is `[0,1,2,3,4,6,10]` = 26s (macOS previously had a local `[0,2,4,6,10]` = 22s). Offline / daemon-unreachable users with a cached summary sit through the whole ladder on every login-path refresh. Because `updateAccountSummary` serializes on `accountSummaryUpdateTask` (`CredentialsManager.swift:558-560`), a concurrent `isAccountValid()` — the Connect gate at `OneClickViewModel+State.swift:196` — blocks behind it.

Suggest bailing on `lastFetchFailed` once a usable cached summary exists, or capping the failure ladder well below 26s.

nym-vpn-apple/Services/Sources/AccountPrefetchGates/OnboardingSessionPolicy.swift
```diff
@@ -51,6 +58,26 @@ public enum OnboardingSessionPolicy: Equatable, Sendable {
...
+        if let reason = error as? VPNErrorReason, case .existingAccount = reason {
+            return true
+        }
+#endif
+#if os(macOS)
+        if let reason = error as? ErrorReason, case .existingAccount = reason {
+            return true
+        }
+#endif
+        return error.localizedDescription == unmappedExistingAccountStoreMessage
```
agentpietrucha, line 79: **This string fallback can never fire.**

Both uniffi error types define `errorDescription` as `String(reflecting: self)`, so `localizedDescription` yields `"NymVPNLib.VpnError.ExistingAccount"` / `"NymVPNLib.RpcError"` — never the Rust `Display` string `"an account is already stored"` (`nym-vpn-lib-types/src/account/mod.rs:91-92`).

The doc comment says it's "for an unmapped leaked `VpnError`", but that's precisely the case it misses. The two tests covering it pass only because they construct synthetic `LocalizedError`s that return the constant.

It's a safety net that provides no safety, and a locale-sensitive string compare by construction. Either drop it, or match on `RpcError.displayChain()`, which does return the Rust `Display` output.

nym-vpn-apple/Home/Sources/26/AppFeature/AppFeatureView.swift
```diff
@@ -279,21 +276,35 @@ private extension AppFeatureView {
...
+        guard PurchaseTransitionPolicy.shouldPushPlanPurchaseAfterDrawerHidden(
+            drawerHidden: viewModel.drawerContent == nil,
+            checkoutNavigationPending: viewModel.isCheckoutNavigationPending
+        ) else { return }
         pushPlanPurchaseNavigation()
         viewModel.consumeNavigationIntent()
         viewModel.checkoutNavigationDidComplete()
     }
 
     func handleDrawerHiddenChange(_ drawerHidden: Bool, viewModel: AppFeatureViewModel) {
```
agentpietrucha, line 291: **This handler is now unreachable.**

It fires on `.onChange(of: drawerContent == nil)`, but `isCheckoutNavigationPending` is only set ~350ms later in `completeCheckoutDrawerTransition`, and `drawerContent == nil` doesn't change again afterwards. So `shouldPushPlanPurchaseAfterDrawerHidden` can never be satisfied here.

No user impact — the token handler covers the real path — but the fallback this guard used to provide is gone. Worth either deleting it or re-deriving a condition that can actually fire.

nym-vpn-apple/Home/Sources/26/AppFeature/AppFeatureView.swift
```diff
@@ -202,25 +202,22 @@ private extension AppFeatureView {
...
-                )
+                if !viewModel.purchaseTransitionOverlayVisible {
+                    ConnectionStatusBackdropLayer(
+                        connectionStatus: viewModel.connectionStatus,
+                        drawerContentIsNil: viewModel.drawerContent == nil,
+                        drawerHeight: drawerHeight,
+                        bottomSafeAreaInset: bottomSafeAreaInset
+                    )
+                }
+                if viewModel.purchaseTransitionOverlayVisible {
```
agentpietrucha, line 213: **Backdrop remount instead of crossfade.**

`ConnectionStatusBackdropLayer` is now conditionally *removed* rather than covered, and `.transition(.opacity)` plus the `.animation(_:value: purchaseTransitionOverlayVisible)` modifier were both dropped. Entering checkout is a hard cut, and returning re-mounts `ConnectionStatusBackdrop` fresh, replaying its entry animation — the same class of jitter this change is meant to suppress, just moved to the return path.

Also, `.ignoresSafeArea()` on the overlay is a no-op now that it sits inside the `.clipped()` ZStack.

nym-vpn-apple/Home/Sources/26/PassphraseSignIn/PassphraseSignInViewModel.swift
```diff
@@ -75,6 +67,28 @@ public final class PassphraseSignInViewModel {
...
+        let outcome = await AuthCompletionOutcomeResolver.resolveAfterLoginRegistration(
+            isAccountActive: { self.credentialStore.isAccountActive() },
+            updateAccountSummary: {
+                await self.credentialStore.updateAccountSummary(force: true, untilActive: false)
+            }
+        )
+        sessionCoordinator?.handle(
+            .session(.authCompleted(outcome: outcome, flow: .login))
+        )
+        submissionState = .idle
```
agentpietrucha, line 89: Minor: resetting to `.idle` here re-enables the login button before the screen leaves. On the `.registeredNeedsPurchase` route the sign-in screen stays up for ~500ms (150ms + 350ms drawer stage) with an idle, tappable button, so a second tap can restart the whole login while checkout navigation is in flight.

agentpietrucha (changes requested):
Deep review of the login/checkout session changes. Two blockers plus a few smaller things; details are inline.

**Blocking**

1. **The macOS already-stored path is dead code.** `catch let error as VpnError` never matches — `RpcClient.storeAccount` throws `RpcError` (`nym-vpn-lib-uniffi/src/rpc.rs:265-273`, `pub type Result<T, E = RpcError>` at :573). The headline behaviour of this PR does not ship on macOS.
2. **The `knownInactive` pre-loop early return skips the summary fetch entirely** (both platforms), so a persisted `isActive: true` summary is never refreshed once the controller reports inactive. `isAccountActive()` trusts that flag first, so an expired subscription reads as active indefinitely — and this fires on every `untilActive: false` refresh, not just login.

**Also worth addressing**

3. The daemon returns `ExistingAccount` without comparing the submitted mnemonic, so any valid passphrase logs the user into whatever account was already stored.
4. A failed fetch with a cached summary now burns the full 26s ladder where it used to return immediately.
5. On iOS the new terminal-inactive return lands on `setAccountSummaryLastFetchFailed(true)`, which suppresses the very checkout CTA this stack is trying to surface.

**Verified clean:** package wiring and product resolution (no dependency cycle, `TunnelStatusTests` target exists, `TunnelStatus` already a `GRPCManager` dep); no orphaned call sites for the changed `showsPurchaseTransitionOverlay` / `shouldFinishSummaryPoll` signatures; `requestPlanPurchase` is the only `.pushPlanPurchase` producer and always pairs it with `.stageOneClickForCheckout`, so the tightened push guard doesn't strand real navigation; `paywallTransitionDuration` is the same 0.35 value as before.

Note that CodeRabbit skipped this PR (auto-review is off for non-default base branches), so nothing here was pre-flagged.

agentpietrucha:
@coderabbitai review

nym-vpn-apple/Services/Sources/Services/CredentialsManager/CredentialsManager+iOSAccountController.swift
tommyv1987: Resolved

nym-vpn-apple/Home/Sources/26/PassphraseSignIn/PassphraseSignInViewModel.swift
tommyv1987: This I will split into a follow up ticket.

nym-vpn-apple/Services/Sources/AccountPrefetchGates/AccountSummaryRefreshPolicy.swift
tommyv1987: Intentional s2b fail-first. Changing it reopens honesty. Follow-up I guess if products wants a shorter offline cap.

nym-vpn-apple/Services/Sources/AccountPrefetchGates/OnboardingSessionPolicy.swift
tommyv1987: Typed store error is what ships. Drop the string later - not a login break.

nym-vpn-apple/Home/Sources/26/AppFeature/AppFeatureView.swift
tommyv1987: Token path still navigates. We restore an empty home drawer. Deleting the dead handler is cleanup, not a blocker.

nym-vpn-apple/Home/Sources/26/AppFeature/AppFeatureView.swift
tommyv1987: Animation only. No wrong plan or login.

nym-vpn-apple/Services/Sources/Services/CredentialsManager/CredentialsManager.swift
```diff
@@ -442,7 +442,7 @@ import PathManager
 #if os(macOS)
         try await grpcManager.handleSubscriptionPayment()
         try? await Task.sleep(for: .seconds(2))
-        await updateAccountSummary(force: true)
+        await updateAccountSummary(force: true, untilActive: true)
```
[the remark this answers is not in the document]
tommyv1987: untilActive + failed fetch keeps polling. Test: `untilActivePollRetriesWhenFetchFailedWithCachedActiveSummary`.

agentpietrucha (approved):

PR #6306, opened 2026-09-07
Remove device time out of sync error

pronebird:
## JIRA ticket

JIRA-NYM-1775

## Description

Abolish errors related to device time being out of sync

## Checklist:

- [x] Changelog

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym-vpn-client/6306)

## Bug Fixes

- Accounts with device clock differences no longer enter a device-time error state during synchronization.
- Account and tunnel connection flows proceed without blocking on remote clock synchronization.

## Removed

- Removed device-time synchronization error statuses and related account-summary indicators from user-visible state reporting.
- Removed device-time error messages from user-facing error displays and localized resources.


nym-vpn-core/crates/nym-vpn-account-controller/tests/integrations/paths.rs
```diff
@@ -141,7 +141,7 @@ async fn desynced_device_test() -> anyhow::Result<()> {
 
     test_bench
         .assert_state(AccountControllerState::Error(
-            AccountControllerErrorStateReason::DeviceTimeDesynced,
+            AccountControllerErrorStateReason::MaxDeviceReached, // todo: test what happens now!
```
simonwicky: The installed endpoint are a ready to connect account and a desynced health endpoint. 
In this case, we should get into ready to connect. 

One of two things, either rename the test and change the assertion to capture that a desynced device doesn't cause any problems anymore or delete the test altogether since it's not relevant anymore

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
```diff
@@ -974,8 +972,7 @@ message VpnAccountSummary {
   uint64 remaining_devices = 16;
   bool is_device_active = 17;
   optional google.protobuf.Timestamp last_synced_utc = 18;
-  bool time_synced = 19;
-  bool stale = 20;
+  bool stale = 19;
```
simonwicky: Isn't there a problem with changing numbers like that?

simonwicky (changes requested):

nym-vpn-core/crates/nym-vpn-proto/proto/nym_vpn_service.proto
pronebird: Technically, let's keep the old index

nym-vpn-core/crates/nym-vpn-account-controller/tests/integrations/paths.rs
pronebird: Removed that test.

mrkiwi (approved):

simonwicky (approved):

